PR #10, opened 2016-02-02
Web SDK 109102 changes:

jasonpang:
- getTags,sendTag,sendTags,deleteTag,deleteTags:
        - Return promises
        - Accept an extra parameter for a callback (so the user knows when
         the tags have finished sending)
        - Have tests for both the promises and callbacks

[Review on Reviewable](https://reviewable.io/reviews/one-signal/onesignal-website-sdk/10)

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -70,4 +70,16 @@ gulp.task("copy-js", function() {
       .pipe(rename("/OneSignalSDKWorker.js.map"))
       .pipe(gulp.dest(oneSignalSourceDir + "/public/" + targetFolder));
   }
+
+  if (IS_TEST_BUILD) {
+    gulp.src("./dist/OneSignalSDKTests.js")
+      .pipe(rename("/OneSignalSDKTests.js"))
+      .pipe(gulp.dest(oneSignalSourceDir + "/public/" + targetFolder));
+
+    if (fs.existsSync("./dist/OneSignalSDKTests.js.map")) {
+      gulp.src("./dist/OneSignalSDKTests.js.map")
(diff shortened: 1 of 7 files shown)


gdeglin:
Reviewed 7 of 7 files at r1.
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

*[src/sdk.js, line 1286 \[r1\]](https://reviewable.io:443/reviews/one-signal/onesignal-website-sdk/10#-K9VKoZ4MlhIzKsAyjfa:-K9VKoZ4MlhIzKsAyjfb:508220573) ([raw file](https://github.com/one-signal/onesignal-website-sdk/blob/6801e0ca851e6d43c616e725768f74d41be971be/src/sdk.js#L1286)):*
Maybe we should just use value = value.toString()? This will also prevent users from doing silly things like using [] as a tag value.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/one-signal/onesignal-website-sdk/10)*

jasonpang:
*[src/sdk.js, line 1286 \[r1\]](https://reviewable.io:443/reviews/one-signal/onesignal-website-sdk/10#-K9VKoZ4MlhIzKsAyjfa:-K9_4_6Y0Gv1pZ_4zm1Y:-1038266327) ([raw file](https://github.com/one-signal/onesignal-website-sdk/blob/6801e0ca851e6d43c616e725768f74d41be971be/src/sdk.js#L1286)):*
Note: We decided to just check for false in the end.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/one-signal/onesignal-website-sdk/10)*

PR #21, opened 2016-02-26
Do not auto-prompt the user to subscribe if the user has previously s…

jasonpang:
…ubscribed, and has reset permissions back to Always Ask, and autoregister is false.

[Review on Reviewable](https://reviewable.io/reviews/one-signal/onesignal-website-sdk/21)


jasonpang:
Merging because George approved this change.

PR #26, opened 2016-03-05
Fix sendTags() when called before user registration so that the callb…

jasonpang:
…ack is delayed until registration completes.

- Previously, tags would be sent after the registration completed, but the callback would not be called back.

[Review on Reviewable](https://reviewable.io/reviews/one-signal/onesignal-website-sdk/26)


gdeglin:
Reviewed 1 of 1 files at r1.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/one-signal/onesignal-website-sdk/26)*

PR #32, opened 2016-03-09
Notification click event

jasonpang:
The following are the changes to addListenerForNotificationOpened():

If `notificationClickHandlerMatch` is set to `exact` or omitted from OneSignal.init:
- One of any existing tabs exactly matching the notification's URL is focused and the event is fired on it

If `notificationClickHandlerMatch` is set to `origin`:
- As long as the URL opens to something on the same origin as the page, one of any existing tabs with the same origin is focused and the event is fired on it

In all cases, the event listener does not get destroyed after the first listen. However, in cases with multiple existing tabs, the event is always only fired on one of the tabs.

[Review on Reviewable](https://reviewable.io/reviews/one-signal/onesignal-website-sdk/32)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "onesignal-web-sdk",
   "version": "1.0.0",
-  "sdkVersion": "109156",
+  "sdkVersion": "109157",
   "description": "Web push notifications from OneSignal.",
   "main": "src/entry.js",
   "dependencies": {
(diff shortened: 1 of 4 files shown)


gdeglin:
Reviewed 4 of 4 files at r1.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/one-signal/onesignal-website-sdk/32)*

PR #35, opened 2016-03-15
New events and apis

jasonpang:
All events are now accessible via `OneSignal.on('...')`.

Events we should publicize:

```javascript
OneSignal.on('subscriptionChange', function(newSubscription) {
    // newSubscription is true or false
});
```

```javascript
OneSignal.on('initialize', function() {
    // SDK is initialized
});
```

Legacy support for the following events:

- `window.addEventListener('onesignal.prompt.native.permissionchanged')`
- `window.addEventListener('onesignal.subscription.changed')`
- `window.addEventListener('onesignal.prompt.custom.clicked')`

... since some of our customers are still using these internal listeners.

Some new APIs are available too, and we should discuss which of them should be made permanently public!

`OneSignal.getNotificationPermission`: On Chrome and Firefox, `Notification.permission` returns the notification permission, but on Safari a different call `window.safari.pushNotification.permission(safariWebId).permission` needs to be used. Furthermore for HTTP sites, messages have to be posted to our nested iFrame in order to get the real notification permission of our subdomain. I think we should make this call public.

`OneSignal.getUserId`: Returns the OneSignal user ID if one exists, or `null` otherwise. Does not return the registration ID, only returns the OneSignal user ID.

`OneSignal.optOut` and `OneSignal.isOptedOut`: Not a high priority right now, but I think `setSubscription` is misleading because some users think the subscription can be forced set true. I don't think `optOut` is the best name, maybe something like `setNotificationPreference` is better. We don't have to make these public yet (i.e. don't put them on our docs). Currently we have `setSubscription` but no corresponding `getSubscription`. A few people have asked about this.

---
This change is [Reviewable](https://reviewable.io/reviews/one-signal/onesignal-website-sdk/35)


gdeglin:
Reviewed 1 of 7 files at r2, 13 of 13 files at r3.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/one-signal/onesignal-website-sdk/35)*

PR #41, opened 2016-04-14
Get extra subscription info for new push protocol

jasonpang:
Firefox 46+ and Chrome 50+ support the new encrypted web push protocol.
We obtain and forward the two parameters 'auth' and 'p256dh', both a
member of the PushSubscription object we obtain from
PushManager.subscribe(), to OneSignal. Firefox 44 & 45 do not support
the 'auth' property so we only forward each property if it exists.
Chrome < 50 does not support either 'auth' or 'p256dh'.

---
This change is [Reviewable](https://reviewable.io/reviews/one-signal/onesignal-website-sdk/41)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "onesignal-web-sdk",
   "version": "1.2.0",
-  "sdkVersion": "109250",
+  "sdkVersion": "109300",
   "description": "Web push notifications from OneSignal.",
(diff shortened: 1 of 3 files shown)


gdeglin:
Reviewed 3 of 3 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/one-signal/onesignal-website-sdk/41)*

PR #64, opened 2016-05-17
Add notify button display predicate

jasonpang:
Add a notify button init option that allows passing a display predicate
that hides the notify button if it resolves to false. Predicate accepts
returns of both promises and non-promise values.

For example:

```javascript
notifyButton: {
            prenotify: true,
            showCredit: false,
            size: 'medium',
            position: 'bottom-right',
            enable: true,
            displayPredicate: function() {
              return OneSignal.isPushNotificationsEnabled()
                      .then(function(isPushEnabled) {
                          /* The user is subscribed, so we want to return "false" to hide the notify button */
                          return !isPushEnabled;
                      });
            },
```

---
This change is [Reviewable](https://reviewable.io/reviews/one-signal/onesignal-website-sdk/64)

### DIFF
--- a/src/OneSignal.js
+++ b/src/OneSignal.js
@@ -169,8 +169,21 @@ export default class OneSignal {
         objectAssign(OneSignal.config.bell, OneSignal.config.notifyButton);
         objectAssign(OneSignal.config.notifyButton, OneSignal.config.bell);
       }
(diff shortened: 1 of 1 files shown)


gdeglin:
Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/one-signal/onesignal-website-sdk/64)*

PR #155, opened 2016-10-30
Checking if this is instance of String in String.prototype.capitalize

niklabh:
Capitalize can be called without this set to string, which will break the original function.

---

This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/155)

### DIFF
--- a/src/string.js
+++ b/src/string.js
@@ -3,5 +3,6 @@
  * @returns {string} The string with the first letter capitalized.
  */
 String.prototype.capitalize = function() {
+  if (!(this instanceof String)) return this;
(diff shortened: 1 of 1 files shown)


niklabh:
For Example. This is breaking with jquery on:

$.param("")

jasonpang:
Hi @niklabh,

Sorry about that, we'll make sure not to override default classes! This was from a while ago and we didn't remove it.

jasonpang:
This issue has been fixed and we no longer override prototypes anywhere in our production SDK.

Issue #219, opened 2017-03-21
Support using custom service worker filenames 


jasonpang:
`isPushNotificationsEnabled()` returns `false`, because the [check for whether the service worker is active](https://github.com/OneSignal/OneSignal-Website-SDK/blob/d790181d7b8f50bf497e894b1a934a7471f85d29/src/helpers/ServiceWorkerHelper.ts#L161) assumes a file name containing `"OneSignalSDKWorker"` or `"OneSignalSDKUpdaterWorker"`.

There are other places in the source code containing the hardcoded string `"OneSignalSDKWorker"` and `"OneSignalSDKUpdaterWorker"`. These should be moved off to use `OneSignal.SERVICE_WORKER_PATH` or `OneSignal.SERVICE_WORKER_UPDATER_PATH`. If these variables are not accessible in the service worker, they should be saved to IndexedDb and loaded from there.

Original Intercom Conversation: [Link](https://app.intercom.com/a/apps/344a89aeac3f033e4dec4370781543b948aece6d/inbox/search/conversations/8474401681?q=https%3A%2F%2Fgithub.com%2FOneSignal%2FOneSignal-Website-SDK%2Fissues%2F212)

tixastronauta:
Hi @jasonpang! Is this issue still open or have you guys managed to have `OneSignal.SERVICE_WORKER_PATH` everywhere?

jasonpang:
This issue is still open; a refactored `ServiceWorkerManager` is in the works that doesn't use any hardcoded `"OneSignalSDK"` strings.

jasonpang:
Refactored `ServiceWorkerManager` now accepts custom service worker file names.

PR #236, opened 2017-06-06
support sw custom name

marcmrf:
As discussed here [https://github.com/OneSignal/OneSignal-Website-SDK/issues/219](https://github.com/OneSignal/OneSignal-Website-SDK/issues/219
), there is an option on the sdk for defining a custom file name.

It's also documented in the code, as you can see in the next snippet:

```js
/**
   * The additional path to the worker file.
   *
   * Usually just the filename (in case the file is named differently), but also supports cases where the folder
   * is different.
   *
   * However, the init options 'path' should be used to specify the folder path instead since service workers will not
   * auto-update correctly on HTTPS site load if the config init options 'path' is not set.
   */
  static SERVICE_WORKER_UPDATER_PATH = 'OneSignalSDKUpdaterWorker.js';
  static SERVICE_WORKER_PATH = 'OneSignalSDKWorker.js';
```
Most of the code is using this variables but in some places, the code is still using the literal filename, breaking some methods of the SDK..

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/236)

### DIFF
--- a/src/helpers/MainHelper.ts
+++ b/src/helpers/MainHelper.ts
@@ -291,11 +291,11 @@ export default class MainHelper {
 
   static getPrefixedServiceWorkerNameForEnv() {
     if (Environment.isDev()) {
-      OneSignal.SERVICE_WORKER_PATH = DEV_PREFIX + 'OneSignalSDKWorker.js';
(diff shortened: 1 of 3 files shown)


jasonpang:
@marcmrf This sounds great, thanks for helping out!

marcmrf:
I'll reopen it today after some fixes

PR #348, opened 2018-05-08
Fixed notification clicked event when site is open

### LINKED ISSUE #344: Getting TypeError: OneSignal.getListeners is not a function when pressing On Notification
**Description:**
My Web App used to work but since a couple of Days its not working anymore.
Problem is that when on Chrome Browser (Web and Android) I click on the Push Notification.

**Steps to Reproduce Issue:**
1. Press on the Push Notification
2. Nothing Happens
3. The following shows on the console.log
```ERROR Error: Uncaught (in promise): TypeError: OneSignal.getListeners is not a function```

**Anything else:**
I can't find the same issue here or on google so I think the problem is on my Code?
Am using Angular and the following is from my code:

```
    ngOnInit() {
      console.log('Init OneSignal');
      OneSignal.push(['init', {
        appId: '',
        // autoRegister: false,
        allowLocalhostAsSecureOrigin: true,
        notifyButton: {
          enable: false
        }
      }]);
      console.log('OneSignal Initialized');
      OneSignal.push(() => {
        console.log('Register For Push');
        OneSignal.push(['registerForPushNotifications'])
      });
      OneSignal.push(() => {
        // Occurs when the user's subscription changes to a new value.
        OneSignal.on('subscriptionChange', function (isSubscribed) {
          console.log(`The user's subscription state is now:`, isSubscribed);
          OneSignal.getUserId().then(function (userId) {
            console.log('User ID is', userId);
          });
        });
      });
      this.addListenerForNotificationOpened();
    }

    addListenerForNotificationOpened() {
      OneSignal.push(['addListenerForNotificationOpened', notificationData => {
        console.log('Received NotificationOpened:', notificationData);
        this.addListenerForNotificationOpened();

        // Actions
        const payload = notificationData.data;
        if (payload.action === 'OpenDetail') {
          this.router.navigate(['/app', 'forms', payload.item]);
        }
      }]);
    }
```

### PR DESCRIPTION
**Pull request by jkasten2**
* Issue only occurred when a tab to the site was open in the background when opening a notification
* Resolves #344

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/348)

### DIFF
--- a/src/libraries/Emitter.ts
+++ b/src/libraries/Emitter.ts
@@ -8,8 +8,8 @@ export type OnceEventHandler = {
   listener: EventHandler
 };
 
- interface ListenerMap {
-    [index: string]: (EventHandler | OnceEventHandler)[];
+interface ListenerMap {
+  [index: string]: (EventHandler | OnceEventHandler)[];
 }
 
 /**
@@ -100,7 +100,7 @@ export default class Emitter {
    * @example
    * me.listeners('ready');
    */
-  public listeners(event: string): (EventHandler | OnceEventHandler)[] {
+  public listeners(event: string): (EventHandler | OnceEventHandler)[] | undefined {
     try {
       return this._events[event];
     } catch (e) {
--- a/src/libraries/Emitter.ts
+++ b/src/libraries/Emitter.ts
@@ -25,7 +25,7 @@ export default class Emitter {
 
   constructor() {
     this._events = {};
-  };
+  }
 
   /**
    * Adds a listener to the collection for a specified event.
@@ -33,15 +33,15 @@ export default class Emitter {
   public on(event: string, listener: EventHandler): Emitter {
     this._events[event] = this._events[event] || [];
     this._events[event].push(listener);
-    return (this as any);
-  };
+    return this;
+  }
 
   /**
    * Adds a one time listener to the collection for a specified event. It will
    * execute only once.
    */
   public once(event: string, listener: EventHandler): Emitter {
-    let that = this;
+    const that = this;
 
     function fn(this: Function) {
       that.off(event, fn);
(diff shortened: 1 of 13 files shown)

### CODE REVIEW

File: src/libraries/WorkerMessenger.ts [hunk 1]
```diff
@@ -47,14 +47,13 @@ export class WorkerMessengerReplyBuffer {
...
+    if (this.findListenersForMessage(command).length > 0)
       this.replies[command.toString()].push(record);
-    } else {
+    else
       this.replies[command.toString()] = [record];
-    }
   }
 
-  public findListenersForMessage(command: WorkerMessengerCommand): any {
+  public findListenersForMessage(command: WorkerMessengerCommand): any[] {
```
TimBest: should be able to update this return type to `object[]`. not much but it is slightly more restrictive then `any[]` and matches the type definition of `replies`

jkasten2: There was a `WorkerMessengerReplyBufferRecord` type already but wasn't being used. Just switched to using this here

src/libraries/Emitter.ts
```diff
@@ -102,32 +100,48 @@ export default class Emitter {
    * @example
    * me.listeners('ready');
    */
-  public listeners(event: string): (EventHandler | OnceEventHandler)[] {
+  public listeners(event: string): (EventHandler | OnceEventHandler)[] | undefined {
```
TimBest: no need to change anything here just want to share some thoughts 
 
Convention for return types it to use `void` over `undefined`. `type void = undefined | null` making it a little more flexible for what is returned. 

Although I have heard compelling arguments to never use `null` and only use `undefined` since then you only have one null like type

src/libraries/WorkerMessenger.ts
```diff
@@ -24,37 +24,34 @@ export enum WorkerMessengerCommand {
...
-  command: WorkerMessengerCommand,
-  payload: WorkerMessengerPayload
+  command: WorkerMessengerCommand;
+  payload: WorkerMessengerPayload;
 }
 
-export interface WorkerMessengerReplyBufferRecord {
-  callback: Function,
-  onceListenerOnly: boolean
+export class WorkerMessengerReplyBufferRecord {
```
TimBest: should this be a `class`?

PR #350, opened 2018-05-09
Added tslint and compilerOptions options for IDEs

jkasten2:
* This is the first step in improving code quality with IDE only errors.
* These new settings will be ignored in builds.

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/350)

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -3,6 +3,15 @@
   // Options here only apply to IDE, build compile options defined in the build tsconfig.json above.
   "compilerOptions": {
     "strictNullChecks": true,
-    "noImplicitAny": true
+    "noImplicitAny": true,
+    "noEmitHelpers": true,
+    "noFallthroughCasesInSwitch": true,
+    "noImplicitReturns": true,
+    "noImplicitThis": true,
+    "noUnusedLocals": true,
+    "noUnusedParameters": true,
+    "strictPropertyInitialization": true,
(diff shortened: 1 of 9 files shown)


tsconfig.json
```diff
@@ -1,3 +1,8 @@
 {
-  "extends": "./build/config/tsconfig.json"
+  "extends": "./build/config/tsconfig.json",
+  // Options here only apply to IDE, build compile options defined in the build tsconfig.json above.
+  "compilerOptions": {
+    "strictNullChecks": true,
+    "noImplicitAny": true
```
TimBest: If this is IDE only we should add all these.   
```
    "noEmitHelpers": true,
    "noFallthroughCasesInSwitch": true,
    "noImplicitReturns": true,
    "noImplicitThis": true,
    "noUnusedLocals": true,
    "noUnusedParameters": true,
    "strictPropertyInitialization": true,
    "strictFunctionTypes": true,
    "forceConsistentCasingInFileNames": true,
```

TimBest: also looks like Typescript is okay with comments in JSON?

TimBest:
Also if we are editing tsconfig you should add a `BaseUrl` as well https://github.com/Microsoft/TypeScript/wiki/What's-new-in-TypeScript#base-url

then we can start cleaning up those relative imports

PR #369, opened 2018-06-12
Fix for privateShowHttpPrompt error with force

### LINKED ISSUE #363: TypeError: Cannot read property 'force' of undefined
The new version of script is using `privateShowHttpPrompt` instead of `showHttpPrompt`, but in  `privateShowHttpPrompt` this following code is missing, causing error     
```
if (!options)
      options = {};
```
causing the following error 

```
Uncaught (in promise) TypeError: Cannot read property 'force' of undefined
    at Function.<anonymous> (OneSignal.js:281)
    at EmailDeviceRecord.js:7
    at Object.next (EmailDeviceRecord.js:7)
    at s (EmailDeviceRecord.js:7)
```

jkasten2:
* Fixes "Cannot read property 'force' of undefined" introduced in 150201
* resolves #363

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/369)

### DIFF
--- a/src/OneSignal.ts
+++ b/src/OneSignal.ts
@@ -336,14 +336,14 @@ export default class OneSignal {
    * @PublicApi
    */
   static async showHttpPrompt(options?) {
-    if (!options)
-      options = {};
-
     await awaitOneSignalInitAndSupported();
     await OneSignal.privateShowHttpPrompt(options);
   }
 
   static async privateShowHttpPrompt(options?) {
+    if (!options)
+      options = {};
+
     /*
     Only show the HTTP popover if:
     - Notifications aren't already enabled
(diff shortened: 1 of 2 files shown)


itrush:
:lgtm:

---

Reviewed 2 of 2 files at r1.
Review status: all files reviewed, 1 unresolved discussion (waiting on @jkasten2)

---

*[test/unit/public-sdk-apis/showHttpPrompt.ts, line 24 at r1](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/369#-LEoqJIu9PIB47ZojuOu:-LEoqJIvCHXma9Q-CeqP:bco64a) ([raw file](https://github.com/onesignal/onesignal-website-sdk/blob/a45dc17d5bf79af5baaacd98f12addd313a26d84/test/unit/public-sdk-apis/showHttpPrompt.ts#L24)):*
> ```XML
> 
>   sinonSandbox.stub(MainHelper, "wasHttpsNativePromptDismissed").resolves(true);
>   sinonSandbox.stub(OneSignal, "privateIsPushNotificationsEnabled").resolves(false);
> ```

Just a note, not blocking. 

Even though this method is called privateIsPushNotificationsEnabled, it's visibility in OneSignal class is default, i.e. public. If it was real private member, you wouldn't be able to stub it. The naming is a bit confusing. We may add a note and refactor it and similar methods in the future.

---


*Comments from [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/369#-:-LEoqolv183s-F3byliv:bnfp4nl)*

jkasten2:
Review status: all files reviewed, 1 unresolved discussion (waiting on @jkasten2)

---

*[test/unit/public-sdk-apis/showHttpPrompt.ts, line 24 at r1](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/369#-LEoqJIu9PIB47ZojuOu:-LEp1N0L7Hqp8PPqDdrT:b-tq8wlx) ([raw file](https://github.com/onesignal/onesignal-website-sdk/blob/a45dc17d5bf79af5baaacd98f12addd313a26d84/test/unit/public-sdk-apis/showHttpPrompt.ts#L24)):*
<details><summary><i>Previously, itrush (Iryna Trush) wrote…</i></summary><blockquote>

Just a note, not blocking. 

Even though this method is called privateIsPushNotificationsEnabled, it's visibility in OneSignal class is default, i.e. public. If it was real private member, you wouldn't be able to stub it. The naming is a bit confusing. We may add a note and refactor it and similar methods in the future.
</blockquote></details>

@itrush Good catch, there are a few of these so we should clean them all up in another PR

---


*Comments from [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/369)*

PR #379, opened 2018-07-11
Fix error caused by a user blocking localStorage

jordikroon:
Fix error caused by a user blocking localStorage
This PR fixes the error:

```
Failed to read the 'localStorage' property from 'Window': Access is denied for this document.
```

We host a high traffic website. We see a small percentage of users having this cookies/storage disabled.

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/379)


jkasten2:
@jordikroon Thanks for the the PR, looks good. We will get this in the next release.

Issue #289, opened 2017-11-26
getNotificationPermission returns only "default" and "granted", not "denied"


amankapoor:
Hello,

Refer to https://documentation.onesignal.com/docs/web-push-sdk#section--getnotificationpermission-

It says `getNotificationPermission` can return "default", "granted", "denied", but while implementing it I never get a "denied" response. This also affects `notificationPermissionChange`.

Please take a look. You might want to change the docs, or code.

This is the code I am writing, you may try implementing it on your machine.

```
OneSignal.push(function() {
        OneSignal.isPushNotificationsEnabled(function(isEnabled) {
            if (isEnabled) {
                OneSignal.push(["getNotificationPermission", function(permission) {
                    console.log("Site Notification Permission:", permission);
                    // three cases here
                    if (permission === "default") {
                        $('#notifStatus').text("true " + permission)  
                    } else if (permission === "denied") {
                        $('#notifStatus').text("true " + permission)  
                    } else {
                        $('#notifStatus').text("true " + permission)
                    }
                }]);
            } else {
                // three cases here
                OneSignal.push(["getNotificationPermission", function(permission) {
                    console.log("Site Notification Permission:", permission);
                    // three cases here
                    if (permission === "default") {
                        $('#notifStatus').text("false " + permission)  
                    } else if (permission === "denied") {
                        $('#notifStatus').text("false " + permission)  
                    } else {
                        $('#notifStatus').text("false " + permission)
                    }
                }]);
            }
        });
    });
```

amankapoor:
Ohh, I forgot to add that this issue exists in Chrome. On Firefox, I can see denied. I don't have access to Safari, so no idea about it.

jasonpang:
Hey @amankapoor,

Sorry for the delay in getting back to you.

Due to recent changes in Chrome 62's permission API, we can no longer accurately detect a user's permission for HTTP sites or within a cross-origin iframe. The [first section of this pull request](https://github.com/ampproject/amphtml/pull/11480) explains the changes in more detail.

The link above describes permission changes for cross-origin iframes, but for non-iframes regular HTTP sites, the permission change is simpler but a little more drastic: the permission is always denied. It doesn't matter if you've granted permissions, denied permissions, or never been prompted -- `Notification.permission` will always return `denied`.

To solve this, our workaround is to assume the permission is `default` until we know for sure it is either `granted` or `denied`.

[...]

PR #400, opened 2018-09-01
on_session on every first page view

itrush:
- moved out of registration call, otherwise it was called with every re-subscription if done multiple times during the first page view
- now it's part of InitHelper.onSdkInitialized, works uniformly for all use-cases including safari which was before relying on registration call just for on_session reporting
- sends detected subscription status based on reported browser permission (environment aware)
- includes logic to handle special case with http and default/denied permission as described here in more details https://github.com/OneSignal/OneSignal-Website-SDK/issues/289
- added tests for new logic in InitHelper and MainHelper
- refactored pushsubscriptionchange in service worker test mostly due to one test breaking after replacing on_session with simple player update in registration call

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/400)

### DIFF
--- a/src/helpers/InitHelper.ts
+++ b/src/helpers/InitHelper.ts
@@ -137,6 +137,7 @@ export default class InitHelper {
       }
     }
 
+    // TODO: possibly wrap into Promise.all for parallel execution
     await InitHelper.processExpiringSubscriptions();
     await InitHelper.showNotifyButton();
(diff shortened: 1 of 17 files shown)


jwilm:
@itrush, do you have any estimate for how this will affect API load?

Issue #404, opened 2018-09-11
Chrome 69 disabled notification


tthdvd:
**Description:**

After we updated the chrome to 69 version, on websites with only http the notification is disabled and it's does not work. Nothing shows up (include the onesignal notification accept modal).  It's a little problem, our master site is https, only in local development server we use http. 
The main problem is that when we use the isPushNotificationsEnabled() function nothing response goes back. 

**Steps to Reproduce Issue:**

1. open an http site
2. set up OneSignal
3. use isPushNotificationsEnabled function, nothing response.

itrush:
@tthdvd Investigating. Will keep you posted.

itrush:
Seems to working again as normal. Just remember to clear browser cache. OneSignal.VERSION should help verify if you're on the latest one.

tthdvd:
Thank you guys! :)

PR #405, opened 2018-09-11
Fix for OneSignal init in Chrome 69 on http

itrush:
Chrome 69 introduced a change that navigator.serviceWorker returns undefined in an insecure context, which http is. Our logic to determine if push is supported relies on service worker to be present and returns prematurely not letting init to happen.

This PR adds another check to the isPushNotificationsSupported to cover for this new use-case. If any other required parameters, like subdomain, are missing, OneSignal will fail during init and at least give a better error message for what exactly is wrong with the setup.

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/405)

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -55,6 +55,13 @@ export function isPushNotificationsSupported() {
   const browser = redetectBrowserUserAgent();
   let userAgent = navigator.userAgent || '';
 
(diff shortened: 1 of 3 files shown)


itrush:
Fix for https://github.com/OneSignal/OneSignal-Website-SDK/issues/404

PR #418, opened 2018-10-08
Safari 12.1 prompt

itrush:
Prompting for push notifications in Safari will start requiring user interaction.
https://webkit.org/blog/8406/release-notes-for-safari-technology-preview-64/

Show our slide down prompt if auto prompt is on or if it is called directly from the site's JS without any interaction.

Safari interaction seems to have conflicts with accessing IndexedDB, added a workaround for specific version of safari in registerForPushNotification because it was affecting all prompts and this is the most logical place to do it.

Also made registerForPushNotifications async so that it can be awaited in tests if needed and it now works as setSubscription(true) if user previously subscribed, no need to call it separately. Yay!

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/418)

### DIFF
--- a/src/OneSignal.ts
+++ b/src/OneSignal.ts
@@ -1,6 +1,5 @@
 import bowser from 'bowser';
 
-
 import Environment from './Environment';
 import AlreadySubscribedError from './errors/AlreadySubscribedError';
 import { InvalidArgumentError, InvalidArgumentReason } from './errors/InvalidArgumentError';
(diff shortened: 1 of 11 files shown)


src/OneSignal.ts
```diff
@@ -448,29 +443,26 @@ export default class OneSignal {
...
       }
     }
 
-    if (!OneSignal.initialized)
-      OneSignal.emitter.once(OneSignal.EVENTS.SDK_INITIALIZED, () => __registerForPushNotifications());
-    else
-      return __registerForPushNotifications();
+    if (!OneSignal.initialized) {
+      await new Promise((resolve, _reject) => {
+        OneSignal.emitter.once(OneSignal.EVENTS.SDK_INITIALIZED, async () => {
```
rgomezp: @itrush why are we using an emitter here???

PR #471, opened 2019-01-15
Reading autoRegister flag from server config for typical setup

itrush:
Defaults to false if not passed.
Will be complemented with changes for dashboard but should first start supporting it in web sdk.

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/471)


itrush:
Closing. Decided to go with a different implementation.

PR #488, opened 2019-02-26
Support making direct calls to OneSignal with shim

jkasten2:
* This is to ensure OneSignal.init(...) and all other SDK methods can be called if the SDK is being loaded synchronously.
* This only is setup if the <script> tag is missing async and defer.
  - The number of production sites missing this should be minimal.
* Removed OneSignalPageSDKES5.js target stub, this is now built into the OneSignalSDK.js shim.

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/488)


itrush:
And one more thing. We probably should update our webPushIframe and webPushModal to load sdk with async. Otherwise we get those `Please load OneSignalSDK.js async and use 'var OneSignal = OneSignal || [];'` for our own files.

PR #505, opened 2019-03-20
fixed samsung internet browser icon bug

rgomezp:
- handles case where SDK doesn't detect what the browser is
- fixes icon issue for samsung internet browser

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/505)

### DIFF
--- a/src/popover/Popover.ts
+++ b/src/popover/Popover.ts
@@ -82,32 +82,20 @@ export default class Popover {
     }, true);
   }
 
-  isSamsungBrowser() {
-    if(navigator.userAgent.match(/SAMSUNG|Samsung|SGH-[I|N|T]|GT-[I|N]|SM-[A|N|P|T|Z]|SHV-E|SCH-[I|J|R|S]|SPH-L/i)) {
-      return true;
-    }
-    return false;
-  }
-
   getPlatformNotificationIcon(): string {
-    if (this.notificationIcons) {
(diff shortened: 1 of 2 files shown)


rgomezp:
can we assume navigator object will always be defined or is this something I should check for?

jkasten2:
@itrush What do you think about us changing this code to use `this.notificationIcons.chrome` as a default instead of `'default-icon'` when we don't know what the browser is?
Thinking we only want to use `'default-icon'` if `this.notificationIcons.chrome` is falsey or blank.
This way we don't need to check for the Samsung browser.

```javascript
  getPlatformNotificationIcon(): string {
    if (this.notificationIcons)
      return 'default-icon';

    if (bowser.safari && this.notificationIcons.safari)
      return this.notificationIcons.safari;
    else if (bowser.firefox && this.notificationIcons.firefox)
      return this.notificationIcons.firefox;
    else if (this.notificationIcons.chrome)
      return this.notificationIcons.chrome;

    return 'default-icon';
  }
```

jkasten2:
```javascript
  getPlatformNotificationIcon(): string {
    if (!this.notificationIcons)
      return 'default-icon';

    if (bowser.safari && this.notificationIcons.safari)
      return this.notificationIcons.safari;
    else if (bowser.firefox && this.notificationIcons.firefox)
      return this.notificationIcons.firefox;

    return this.notificationIcons.chrome ||
      this.notificationIcons.firefox ||
      this.notificationIcons.safari ||
      'default-icon';
  }
```

PR #521, opened 2019-05-23
Unsubscribe push instead of unregister sw

jkasten2:
## Stop unregistering ServiceWorkers before replacing

* Stop unregistering the existing 3rd party ServiceWorker before registering ours over the top.
* We did this to make sure we didn't end up using a possible existing push subscription from it.
* Instead we're now calling pushManager.getSubscription().unsubscribe() to unsubscribe from the old push keys if we can't replace the push subscription with our own keys.
* This different strategy has the following benefits;
  - Fixes compatibly with the ServiceWorkerContainer.register.prototype importScripts multiple serviceworker support that Akamai does.
    - Possibly some other libraries / SDKs do this.
  - The extra handling around catching VAPID key changes when subscribing for push means changing OneSignal app_ids won't result in a mis-match server key issue.
  - Under the hood in the browser this is most likely more efficient, it doesn't need to remove anything attached to the service worker.

### Test 1 - Changing OneSignal App Id issue
1. Visit https://onesignal.com
2. Subscribe for push
3. Go to https://onesignal.com/404
4. Copy the following in the javascript console
*Replace `YOUR_APP_ID_HERE` before doing so.*
```javascript
function addScriptToPage(url) {
  const scriptElement = document.createElement('script');
  scriptElement.src = url;
  document.head.appendChild(scriptElement);
}
addScriptToPage("https://cdn.onesignal.com/sdks/OneSignalSDK.js");
var OneSignal = window.OneSignal || [];
  OneSignal.push(function() {
    OneSignal.log.setLevel('trace');
    OneSignal.init({
      appId: "YOUR_APP_ID_HERE",
    });
  });
```
5. Go to 2nd app and send yourself a push
6. Observer it errors out

### Test 2 - Akamai compatibility
Tested on a live site with OneSignal and Akamai setup, used Charles proxy to test the issue was fixed.

## Unit tests
List of coverage and tests, new as well as existing that cover behavior already.
### test/unit/managers/SubscriptionManager
1. **NEW test, NEW behavior** - `applicationServerKey = undefined`
   -  "null applicationServerKey throws when subscribing"
2. **Modified test, NEW behavior** - `existingSubscription` & different `applicationServerKey`
   - 'resubscribe-existing strategy uses new subscription applicationServerKey'
3. **NEW test** - `SubscriptionStrategyKind.ResubscribeExisting` & `existingSubscription.options != null`
   -  "resubscribe existing strategy does not unsubscribes if options are not null"
4. `SubscriptionStrategyKind.ResubscribeExisting` & `existingSubscription.options = null`
   - 'resubscribe existing strategy unsubscribes and creates new subscription if existing subscription options are null'
5. `SubscriptionStrategyKind.SubscribeNew` & `existingSubscription`
   -  'subscribe new strategy unsubscribes existing subscription to create new subscription'

This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/521)

### DIFF
--- a/src/managers/SubscriptionManager.ts
+++ b/src/managers/SubscriptionManager.ts
@@ -422,7 +422,7 @@ export class SubscriptionManager {
    *
    * If the VAPID key isn't present, undefined is returned instead of null.
    */
-  public getVapidKeyForBrowser(): ArrayBuffer {
+  public getVapidKeyForBrowser(): ArrayBuffer | undefined {
     // Specifically return undefined instead of null if the key isn't available
     let key = undefined;
 
@@ -471,7 +471,7 @@ export class SubscriptionManager {
     const existingPushSubscription = await pushManager.getSubscription();
 
     /* Record the subscription created at timestamp only if this is a new subscription */
-    let shouldRecordSubscriptionCreatedAt = !existingPushSubscription;
+    let isNewSubscription = !existingPushSubscription;
 
     /* Depending on the subscription strategy, handle existing subscription in various ways */
     switch (subscriptionStrategy) {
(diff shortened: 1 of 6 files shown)


src/managers/SubscriptionManager.ts
```diff
@@ -565,6 +534,55 @@ export class SubscriptionManager {
     return pushSubscriptionDetails;
   }
 
+  private static async updateSubscriptionTime(updateCreatedAt: boolean, expirationTime: number | null): Promise<void> {
+    const bundle = await Database.getSubscription();
+    if (updateCreatedAt) {
+      bundle.createdAt = new Date().getTime();
```
itrush: talked in person about deleting it.

src/managers/SubscriptionManager.ts
```diff
@@ -565,6 +534,55 @@ export class SubscriptionManager {
...
+    return result;
+  }
+
+  // Subscribes the ServiceWorker for a pushToken.
+  // If there is an error doing so unsubscribe from existing and try again
+  //    - This handles subscribing to new server VAPID key if it has changed.
+  // return type - [PushSubscription, createdNewPushSubscription(boolean)]
+  private static async doPushSubscribe(
+    pushManager: PushManager,
+    applicationServerKey: ArrayBuffer | undefined)
```
itrush: throw if `applicationServerKey` undefined

src/managers/SubscriptionManager.ts
```diff
@@ -527,34 +500,30 @@ export class SubscriptionManager {
             Because of this, we should unsubscribe the user from push first and then resubscribe
             them.
           */
-          await existingPushSubscription.unsubscribe();
 
           /* We're unsubscribing, so we want to store the created at timestamp */
-          shouldRecordSubscriptionCreatedAt = false;
+          isNewSubscription = await SubscriptionManager.doPushUnsubscribe(existingPushSubscription);
```
itrush: talked in person, don't care about the result anymore. will have another check inside of `doPushSubscribe`

PR #568, opened 2019-10-31
Fix requireInteraction macos display issue

jkasten2:
### Removed persistNotification "force"
* "force" was always being set if the config had true set
* OneSignal.config.userConfig is populated from remote config when typical site setup is used.
* Updated test to match this new simplified logic.

### Set requireInteraction = false, Chrome macOS 10.15
* Setting requireInteraction to false on Chrome macOS 10.15 otherwise a notification will not display!
  - https://bugs.chromium.org/p/chromium/issues/detail?id=1007418
* This requireInteraction is also a know issue on Opera macOS 10.14+ which is now also handled by setting this to false
  - https://forums.opera.com/topic/31334/push-notifications-with-requireinteraction-true-do-not-display-on-macos
* Added tests to ensure these environments result in disabling requireInteraction
* Moved filterNoticationOptions.ts tests into ServicerWorkerManger.ts tests.
* Note double equal on persistNotification == null in intentional, to account for both null and undefined.
* Note intentionally removed Chrome 59+ macOS logic and comment as persistNotification was never true, it was either false or "force"
   so this check was useless.
   - Also until / if macOS or Chrome can fix the issues requireInteraction shouldn't be use enabled an any case anyway.

### Util and Misc
* Added padStart and isVersionAtLeast Util methods
* Fixed setUserAgent used in unit tests so it works with ServicerWorker tests too

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/568)


src/helpers/InitHelper.ts
```diff
@@ -414,15 +414,14 @@ export default class InitHelper {
...
-      } else {
-        opPromises.push(Database.put('Options', { key: 'persistNotification', value: true }));
-      }
-    }
+
+    const persistNotification = OneSignal.config.userConfig.persistNotification;
+    opPromises.push(
+      Database.put('Options', {
+        key: 'persistNotification',
+        value: persistNotification != null ? persistNotification : true
```
collingreen: the logic on this side is way more clear but won't it still allow force through? should we add `!!` to coerce the not-null-or-undefined value into true/false?

src/service-worker/ServiceWorker.ts
```diff
@@ -408,13 +408,14 @@ export class ServiceWorker {
 
     notification.heading = notification.heading ? notification.heading : defaultTitle;
     notification.icon = notification.icon ? notification.icon : (defaultIcon ? defaultIcon : undefined);
-    var extra: any = {};
+    const extra: any = {};
     extra.tag = notification.tag || appId;
-    if (persistNotification === 'force') {
+    if (persistNotification === 'force')
```
collingreen: removing braces can lead to dangerous merges (#gotofail)

src/utils/Utils.ts
```diff
@@ -98,6 +98,39 @@ export class Utils {
...
+   *   to version.
+   * Limited to only checking for major and minor version values, patch versions are ignored
+   * @param toCheck - Version we want to check
+   * @param compareTo - Version we want to be at or higher
+   * @returns {string} - Returns true if toCheck >= compareTo
+   */
+  public static isVersionAtLeast(toCheck: string | number, compareTo: number): boolean {
+    const osVersionParts = toCheck.toString().split(".");
+    const majorVersion = Utils.padStart(osVersionParts[0], 2, "0");
+    const minorVersion = Utils.padStart(osVersionParts[1], 2, "0");
```
collingreen: we should avoid crashing on single digit version numbers just to make this safer for the future. Maybe something like:

```
const minorVersionPart = osVersionParts.length >= 2 ? osVersionParts[1] : 0;
const minorVersion = Utils.padStart(minorVersionPart);
```

src/service-worker/ServiceWorker.ts
```diff
@@ -488,28 +489,30 @@ export class ServiceWorker {
...
+   * @param notificationOptions - Value passed to ServiceWorkerRegistration.prototype.showNotification
+   */
+  static fixPlatformSpecificDisplayIssues(notificationOptions: any): any {
+    const clone = { ...notificationOptions };
     const browser = OneSignalUtils.redetectBrowserUserAgent();
 
-    if (browser.chrome &&
-      browser.mac &&
-      clone) {
+    if (browser.chrome && browser.mac && Utils.isVersionAtLeast(browser.osversion, 10.15)) {
```
collingreen: it looks like before we were (trying) to set requireInteraction to false for all mac chrome browsers but now we are only doing it for 10.15+ 

Is this comment from the old code no longer true?

> * Due to Chrome 59+ notifications on Mac OS X using the native toast style
> * which limits the number of characters available to display the subdomain	
> * to 14 with requireInteraction and 28 without, we force Mac OS X Chrome
> * notifications to be transient.

src/service-worker/ServiceWorker.ts
```diff
@@ -408,13 +408,14 @@ export class ServiceWorker {
...
-    var extra: any = {};
+    const extra: any = {};
     extra.tag = notification.tag || appId;
-    if (persistNotification === 'force') {
+    if (persistNotification === 'force')
       extra.persistNotification = true;
-    } else {
+    else if (persistNotification == null)
+      extra.persistNotification = true;
+    else
```
collingreen: this whole block can probably be 

```
extra.persistNotification = (persistNotification === false) ? false : true;
```
```

src/service-worker/ServiceWorker.ts
jkasten2: This was removed intentionally, it was always 'force' or false. This logic was broken for a long time so this is now the expected behavior. Also can't enable anyway for macOS 10.15 if / util Chrome or macOS fixes this issue

src/service-worker/ServiceWorker.ts
jkasten2: Found `extra.persistNotification = persistNotification !== false;`

src/utils/Utils.ts
jkasten2: Fixed

luisjoserivera:
`requireInteraction` has been working in Chrome running on macOS for a long time. The forced false for the persistNotification option should be removed or updated to match the specific combination presenting the issue.

PR #581, opened 2019-12-20
Session refactoring cleanup

itrush:
### Motivation

Previously we were tracking sessions per tab, i.e. every new open tab was counted as a new session while page refresh was not even if page was left open in the background for a long time.

New way of session tracking should be much more similar to Android SDK and be more logical from user perspective:
- it will track the session across tabs/windows
- it can track how much time user spends on a site
- it should allow adding time triggers later based on time spent on a page

### Implementation

The refactored session tracking heavily relies on service workers (SW). Previously all api calls were performed from the page which made them somewhat fragile to page refreshes. The new approach moves api calls related to session tracking into the service worker.

#### Chrome and Firefox

For these browsers we already utilize SW since they are used for push notifications. It means there was no change needed to install SW as part of initialization flow. The main changes are in setting up additional messaging between page and SW to track session related events, with different handling of HTTPS and HTTP setups.
HTTP setup is implemented using the subscription workaround where we add an iframe pointing to a custom HTTPS OneSignal (OS) subdomain and install SW there. Page can talk to iframe, iframe talks to its SW and back.

#### Safari

Safari does not use SW for push but they became supported a few versions ago so we'll be installing SW on Safari for HTTPS sites (SW are not allowed for HTTP sites). Session support for HTTP sites in Safari is not implemented yet, will need a separate PR.

#### General flow

* When user is not subscribed, session is not tracked.
* When user is subscribed
   * and opens the page for the first time, page notifies SW to start a new session. SW creates a new session object and saves it in IndexedDb then sends an api call to on_session api endpoint to notify OneSignal.
   * when page becomes out of focus (blur event) or is hidden (visibilitychange event) or page is closed (beforeunload event), page notifies SW of session change
   * when page becomes focused (focus event) or becomes visible after being hidden (visibilitychange event), page notifies SW of session change
   * when SW gets a message about possible session change, it sets up a debounce for session refresh function. Debounce can help with situations when multiple events are incoming, e.g. 2 tabs are open pointing to the pages on the same origin and user switches between them, multiple tabs are closed when browser closes, etc.
   * session refresh logic queries the focused state of all controlled pages and decides whether to mark session as active or inactive
   * if session stays inactive for longer than 30 seconds, SW finalized the session (reports it to OS api to on_focus endpoint and cleans it up from IndexedDb)
* When user unsubscribes, page notifies SW of session change [need to verify] 

This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/581)


src/managers/ServiceWorkerManager.ts
```diff
@@ -423,6 +423,26 @@ export class ServiceWorkerManager {
     workerMessenger.on(WorkerMessengerCommand.NotificationDismissed, data => {
       Event.trigger(OneSignal.EVENTS.NOTIFICATION_DISMISSED, data);
     });
+
+    const isHttps = !OneSignalUtils.isUsingSubscriptionWorkaround();
+    // const isSafari = !!bowser.safari; or typeof window.safari !== "undefined"
+    const isSafari = false; // TODO: GET REAL VALUE
```
collingreen: important TODO

src/managers/SessionManager.ts
```diff
@@ -82,20 +107,31 @@ export class SessionManager {
...
   }
 
-  private async handleOnFocus(e: Event): Promise<void> {
+  async handleOnFocus(e: Event): Promise<void> {
     Log.debug("handleOnFocus", e);
+    /**
+     * Firefox has 2 focus events with different targets (document and window).
+     * While Chrome only has one on window.
+     * Target check is important to avoid double-firing of the event.
+     */
```
collingreen: great comment

src/managers/SessionManager.ts
```diff
@@ -13,36 +13,61 @@ export class SessionManager {
...
+    const isHttps = this.isHttps();
+
     const payload: UpsertSessionPayload = {
       deviceId,
       deviceRecord: deviceRecord.serialize(),
       sessionThreshold: OneSignal.config.sessionThreshold,
       enableSessionDuration: OneSignal.config.enableSessionDuration,
       sessionOrigin,
+      isHttps,
+      isSafari: this.isSafari(),
```
collingreen: why do we need to send isSafari in the session info? Is this a proxy the 'does this platform support service workers/active page events correctly' question?

src/utils/Utils.ts
```diff
@@ -152,6 +155,21 @@ export class Utils {
       throw new Error("Player id cannot be empty");
     }
   }
+
+  public static async enforceAppIdAndPlayerId<T>(
```
collingreen: nice refactor

src/utils/OneSignalUtils.ts
```diff
@@ -110,6 +111,17 @@ export class OneSignalUtils {
   public static logMethodCall(methodName: string, ...args: any[]) {
     return Log.debug(`Called %c${methodName}(${args.map(Utils.stringify).join(', ')})`, Utils.getConsoleStyle('code'), '.');
   }
+
+  static isHttps(): boolean {
+    if (OneSignalUtils.isSafari()) {
+      return window.location.protocol === "https:";
+    }
+    return !OneSignalUtils.isUsingSubscriptionWorkaround();
```
collingreen: was surprised this is part of the 'isHttps' check -- is there any chance of https with the subscription workaround (that isn't safari)?

src/OneSignal.ts
```diff
@@ -820,6 +820,7 @@ export default class OneSignal {
   static proxyFrameHost: ProxyFrameHost;
   static proxyFrame: ProxyFrame;
   static emitter: Emitter = new Emitter();
+  static cache: any = {};
```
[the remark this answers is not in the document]
collingreen: ^ sometimes it is functions, sometimes boolean flags about those functions -- added another question later to see if we need both, the answer to which should inform this `any` interface

src/helpers/InitHelper.ts
```diff
@@ -47,24 +47,30 @@ export default class InitHelper {
...
-        'visibilitychange',
-        (_: any, destroyEventListener: Function) => {
-          if (document.visibilityState === 'visible') {
-            destroyEventListener();
-            InitHelper.sessionInit();
-          }
-        },
-        true
-      );
+      InitHelper.postponeSesstionInitUntilPageIsInFocus();
```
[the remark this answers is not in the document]
itrush: thanks, fixed

src/helpers/InitHelper.ts
```diff
@@ -47,24 +47,30 @@ export default class InitHelper {
     await OneSignal.context.serviceWorkerManager.installWorker();
 
     OneSignal.context.pageViewManager.incrementPageViewCount();
+
+    Log.warn("document.visibilityState", document.visibilityState);
```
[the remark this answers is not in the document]
itrush: removed

src/models/Session.ts
```diff
@@ -15,3 +18,24 @@ export interface Session {
 }
 
 export const ONESIGNAL_SESSION_KEY = "oneSignalSession";
+
+export function initializeNewSession(
+  options: Partial<Session> & { deviceId: string; appId: string, deviceType: number; }
```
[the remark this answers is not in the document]
itrush: not with 2.8.2 ts version :) maybe later

src/helpers/ServiceWorkerHelper.ts
```diff
@@ -44,21 +46,84 @@ export default class ServiceWorkerHelper {
...
+      const session: Session = initializeNewSession(
+        { deviceId, appId: deviceRecord.app_id, deviceType:deviceRecord.device_type }
+      );
+      await Database.upsertSession(session);
+      /**
+       * Send on_session call on each new session initialization except the case
+       * when player create call occurs, e.g. first subscribed or re-subscribed cases after clearing cookies,
+       * since player#create call updates last_session field on player.
+       */
+      if (sessionOrigin !== SessionOrigin.PlayerCreate) {
```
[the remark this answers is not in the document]
itrush: I'd say it's not worth it at this point. It's just us trying to save one api call.

src/managers/SessionManager.ts
itrush: Exactly 'does this browser active page events correctly'. Will rename it, thanks.

src/utils/OneSignalUtils.ts
itrush: we now have new class `EnvironmentInfo` which contains correct set of things. https for actual https, etc. I'll refactor the code to use it instead of in-place checks.

PR #589, opened 2020-01-15
Removed global OneSignal any define, added imports

jkasten2:
* Correct all OneSignal typing so IDE hints and compile time checks work on the OneSignal object.
* TODOs
  * Fix test runtime errors
  * Fix `OneSignalSDKWork.js` bloated size
    * Attempt to use TS 2.9's [import() method to import types only](https://davidea.st/articles/typescript-2-9-import-types)
    * `declare var OneSignal: import('OneSignal');`

## 
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/589)

### DIFF
--- a/src/CustomLink.ts
+++ b/src/CustomLink.ts
@@ -3,6 +3,7 @@ import { AppUserConfigCustomLinkOptions } from "./models/AppConfig";
 import { ResourceLoadState } from "./services/DynamicResourceLoader";
 import Log from "./libraries/Log";
 import { RegisterOptions } from './helpers/InitHelper';
+import OneSignal from "./OneSignal";
 
 export class CustomLink {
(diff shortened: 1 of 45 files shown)


jkasten2:
See branch `fix_onesignal_any_typing_take2` as another attempt at this by using the new `import type` TypeScript feature. Unfortunately it comes with a some cavots too. `import type` isn't enough alone, you also need to add `declare var` with the type as well.

Considering using the static `OneSignal` in internal parts of the SDK mean circular references the direction we should take is removing the references instead of making them typed.  I recommend these branches just stay around as reference only and fixing the circular reference as the solution here.

jkasten2:
Some of this work is already being done as part of User Model in 16.x.x so closing this PR.

PR #599, opened 2020-01-24
Test cleanup around onPushReceived

jkasten2:
* Tests for ServiceWorkerManager.ts should test only the usage from a page.
* The new context/sw tests should be for the ServiceWorker itself.

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/599)

### DIFF
--- a/src/service-worker/ServiceWorker.ts
+++ b/src/service-worker/ServiceWorker.ts
@@ -199,7 +199,8 @@ export class ServiceWorker {
                       .then(() => {
                         return ServiceWorker.workerMessenger.broadcast(WorkerMessengerCommand.NotificationDisplayed, notif).catch(e => Log.error(e))
                       })
-                      .then(() => ServiceWorker.executeWebhooks('notification.displayed', notif).catch(e => Log.error(e)))
+                      .then(() => ServiceWorker.executeWebhooks('notification.displayed', notif)
+                      .then(() => ServiceWorker.sendConfirmedDelivery(notif)).catch(e => Log.error(e)));
                 }).bind(null, notification));
               }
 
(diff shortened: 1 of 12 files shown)


src/service-worker/ServiceWorker.ts
```diff
@@ -202,11 +196,11 @@ export class ServiceWorker {
                 // Never nest the following line in a callback from the point of entering from retrieveNotifications
                 notificationEventPromiseFns.push((notif => {
                   return ServiceWorker.displayNotification(notif)
-                      .then(() => ServiceWorker.updateBackupNotification(notif).catch(e => Log.error(e)))
                       .then(() => {
                         return ServiceWorker.workerMessenger.broadcast(WorkerMessengerCommand.NotificationDisplayed, notif).catch(e => Log.error(e))
```
rgomezp: Linter complains about line length

src/service-worker/ServiceWorker.ts
```diff
@@ -202,11 +196,11 @@ export class ServiceWorker {
                 // Never nest the following line in a callback from the point of entering from retrieveNotifications
                 notificationEventPromiseFns.push((notif => {
                   return ServiceWorker.displayNotification(notif)
```
rgomezp: Is `return` necessary here?

src/services/Database.ts
```diff
@@ -262,8 +261,6 @@ export default class Database {
       await this.put("Ids", {type: "WORKER1_ONE_SIGNAL_SW_VERSION", id: state.workerVersion});
     if (state.updaterWorkerVersion)
       await this.put("Ids", {type: "WORKER2_ONE_SIGNAL_SW_VERSION", id: state.updaterWorkerVersion});
-    if (state.backupNotification)
-      await this.put("Ids", {type: "backupNotification", id: state.backupNotification});
```
rgomezp: Nit: there should be a space after and before brackets

rgomezp: same for above calls

rgomezp:
LGTM except a couple nits. After, I think it's good to merge

src/service-worker/ServiceWorker.ts
jkasten2: This line wasn't changed in this PR, thinking we should leave as is for now as this whole method needs a refactor

src/services/Database.ts
jkasten2: 👍

PR #607, opened 2020-01-30
Temporarily disable confirmed delivery.

RyanAmos:
This temporarily disabled confirmed delivery.  This does the following:
1. No longer fetches the app config.
2. Sets up the `appConfig` value to an object that has the `receiveReceiptsEnable` value as false.  
3. Forces an early return in this function because of the `receiveReceiptsEnable` being false.

This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/607)


jkasten2:
This is no longer an issue, closing

PR #613, opened 2020-02-13
Outcomes events with new session logic

itrush:
**Web Outcomes implementation**

Relies on the new session logic. Has a few session related fixes that I discovered while testing outcomes. Will keep those fixes in the branch for easier review.

Not absolutely clear commit history, sorry! Some architecture decisions proved to be not ideal when tested in real situation.

**Detailed description of included changes**
- Each arriving notification gets stored into a new table called NotificationReceived. It gets cleaned up when outcome related logic is invoked in OutcomeHelper.getAttribution.
  _Open Question: currently it resides in indirect outcome section, probably should be refactored to clean up the table regardless of the type of outcome_

- Each clicked notification gets stored in a new table called NotificationClicked and gets cleaned up on session finalization stage
_Open Question: if session is already active when notification click happens, maybe not even store it?_

- Outcomes attribution happens in 2 places:
  - OneSignal.sendOutcome for any custom set outcome
  - included into on_focus call to attribute session correctly

- Updated IndexedDb version to make sure ondatabaseupgradeneeded is triggered and creates new tables

**Side notes**
Successfully tested https integration in Chrome. Firefox unfortunately does not respect long timeout for session finalization. I'll look into wrapping it into waitUntil.

This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/613)


itrush:
closing as duplicate

PR #616, opened 2020-03-03
Prompt Delays

rgomezp:
## Context
Many sites that implement OneSignal for web push ask for permission immediately when a user reaches their site for the first time. 

This annoying for the end-user (who doesn’t have context on if the site has useful content yet), and may be suboptimal for the site owner (who may lose potential opt-ins if users had more context)

Currently, implementing time or session based delays requires a site owner to write custom code and use the  ‘custom site setup’ instead of  ‘typical site setup’ 

Building time or session delay into ‘typical site setup’ makes OneSignal more useful to site owners and less frustrating to end users

### REMOTE PARAMS

The “sync” call already contains the following configuration parameters:

```
{
   config: {
      staticPrompts: {
         native: {},
         bell: {},
         slidedown: {}
      }
   }
}

```
We can add two new parameters to the prompt configuration objects: **`timeDelay`** and **`pageViews`**

Example:

```
{
   config: {
      staticPrompts: {
         native: {timeDelay: 5, pageViews: 3},
      }
   }
}
```

Initialization already handles getting the configuration options:
```

OneSignal.init

OneSignal.initializeConfig

ConfigManager().getAppConfig

```
These values can then be accessed from the OneSignal context with

```
OneSignal.config.userConfig.promptOptions.slidedown.<value>
" "                                      .native.pageViews
" "                                      .slidedown.timeDelay
...etc
```

##### `incrementPageViewCount`: 
added a "total" pageView count that persists past individual sessions. Gets written to Local Storage to then be used by prompting logic.

##### `getLocalPageViewCount` : 
getter

##### `setLocalPageViewCount` : 
setter

##### `DelayedPromptType`: 
defines two types of delayed prompts: 1) `native` and `slidedown`
##### `internalShowAutoPrompt`:
now checks `pageView` condition and uses `showDelayedPrompt` to prompt either the native or slidedown prompts

##### `showDelayedPrompt`: 
encapsulates internal prompting methods for both native or slidedown along with a `setTimeout` in case there is a `timeDelay` passed into the method
##### `isPageViewConditionMet`: 
checks whether the local `pageView` count is greater than or equal to the `pageView` value set for the site as part of the configuration options that are 1) either passed in from the server/backend or 2) from the user config (initialization options)
##### `getDelayedPromptOptions`: 
getter for either the `native` or `slidedown` prompting options from the `promptOptions` object

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/616)

### DIFF
--- a/src/models/AppConfig.ts
+++ b/src/models/AppConfig.ts
@@ -373,6 +373,10 @@ export interface ServerAppConfig {
         persist: boolean;
       };
     };
+    outcomes: {
+      influencedTimePeriodMin: number;
+      influencedNotificationsLimit: number;
+    },
     vapid_public_key: string;
     onesignal_vapid_public_key: string;
     http_use_onesignal_com?: boolean;
--- a/src/models/AppConfig.ts
+++ b/src/models/AppConfig.ts
@@ -1,3 +1,5 @@
+import { OutcomesConfig } from "./Outcomes";
+
 export interface AppConfig {
   /**
    * The OneSignal dashboard app ID. Although this value is provided, it isn't
@@ -102,20 +104,6 @@ export interface AppUserConfig {
   requiresUserPrivacyConsent?: boolean;
   pageUrl?: string;
   outcomes?: OutcomesConfig;
-}
-
-export interface OutcomesConfig {
-  direct: {
-    enabled: boolean;
-  }
-  indirect: {
-    enabled: boolean;
-    influencedTimePeriodMin: number;
-    influencedNotificationsLimit: number;
-  }
-  unattributed: {
-    enabled: boolean;
-  }
 }
 
 interface BasePromptOptions {
--- a/src/models/AppConfig.ts
+++ b/src/models/AppConfig.ts
@@ -101,6 +101,21 @@ export interface AppUserConfig {
   allowLocalhostAsSecureOrigin?: boolean;
   requiresUserPrivacyConsent?: boolean;
   pageUrl?: string;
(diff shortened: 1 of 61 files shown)


src/models/AppConfig.ts
```diff
@@ -108,16 +108,22 @@ export interface AppUserConfig {
 
 interface BasePromptOptions {
   enabled: boolean;
+}
+
+export interface DelayedPromptOptions extends BasePromptOptions {
   autoPrompt?: boolean;
+  timeDelay?: number;
+  pageViews?: number;
```
rgomezp: the reason I left these optional for now is because Typescript is fussy about the types not matching in some other files when running a test (breaks in several places). Will try to fix later

src/models/AppConfig.ts
```diff
@@ -230,6 +236,8 @@ export interface ServerAppConfigPrompt {
   native: {
     enabled: boolean;
     autoPrompt: boolean;
+    pageViews?: number;
+    timeDelay?: number;
```
rgomezp: same here

src/models/AppConfig.ts
```diff
@@ -270,6 +278,8 @@ export interface ServerAppConfigPrompt {
   slidedown: {
     enabled: boolean;
     autoPrompt: boolean;
+    pageViews?: number;
+    timeDelay?: number;
```
rgomezp: and here

PR #621, opened 2020-03-31
Building outcomes on top of new session logic

itrush:
Outcomes is a feature recently introduced in OneSignal dashboard and currently supported by ios and android only. This PR adds outcomes support to Web SDK.

How outcomes work:
- Browser tracks all incoming notifications saving them to local db table with timestamps.
- Browser tracks all clicked notifications. If a clicked notification results in a new session to be created, it's considered direct.
- Each app can configure its own attribution window for outcomes. I believe the default is now 1 hour. If a new session is created within that time frame since getting a notif, the session is considered indirectly attributed to 1 or more notifications.
- If session is not directly attributed and there are no notifications within outcomes attribution time frame, session is considered unattributed.
- Session reports both attribution and duration of the session which is translated into weighted outcome.
- Same logic applies to custom outcomes. The only difference is that they have to be triggered by user's code via `OneSignal.sendOutcome(value, optional_value)`. The attribution is decided the same way as described above.

New session logic and outcomes implementation heavily rely on service worker (SW) and having push as part of SW so it's best supported by browsers that implemented push via SW events.

Safari's way of handling push makes it impossible to track direct and indirect outcomes making all of them to be unattributed.

This branch also includes improvements and bug fixes for session logic that became obvious during testing.

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/621)

### DIFF
--- a/src/service-worker/ServiceWorker.ts
+++ b/src/service-worker/ServiceWorker.ts
@@ -1,7 +1,7 @@
 import bowser from "bowser";
 
 import Environment from "../Environment";
-import { WorkerMessenger, WorkerMessengerCommand } from "../libraries/WorkerMessenger";
+import { WorkerMessenger, WorkerMessengerCommand, WorkerMessengerMessage } from "../libraries/WorkerMessenger";
 import SdkEnvironment from "../managers/SdkEnvironment";
 import ContextSW from "../models/ContextSW";
 import OneSignalApiBase from "../OneSignalApiBase";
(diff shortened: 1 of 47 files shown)


src/service-worker/ServiceWorker.ts
```diff
@@ -802,7 +802,14 @@ export class ServiceWorker {
       }
       Log.info("NotificationClicked", notificationClicked);
       saveNotificationClickedPromise = (async (notificationClicked) => {
-        return await Database.put("NotificationClicked", notificationClicked)
+        try {
```
jkasten2, line 808: TOOD: @jkasten2 Look into if mobile upgrades session when a notification is clicked to direct

jkasten2: I checked the Android SDK and confirmed that the session is not changed if the app is not in focus.

It would more correct to detect here if any page is in focus when clicking though. This detection should be done here instead of the `getCurrentSession()` check. However not sure if the fact of clicking on a notification creates any sort of race condition, however the de-bounce logic in the service worker will probably account for this.

itrush:
@jkasten2 address all feedback. fixed tests. updated and tested indexeddb upgrade logic, works now. ready for a final (?) bug bash in the next couple of days.

PR #628, opened 2020-04-16
Outcomes fix

itrush:
Unfortunately I still was not able to find a reason for not seeing on_focus traffic to go up.

But I fixed one issue for staging where we were not checking correct domain for clients and it resulted in tracking active clients incorrectly. Should not be the case on production as far as I can see but shouldn't hurt either.

ALso added logging for service worker. Controlled by a boolean variable on SW which can be set through inspect of service worker directly, i.e. `self.shouldLog = true`, or via postMessage for browsers that don't allow inspect:
```
navigator.serviceWorker.controller.postMessage({
  command: "os.set_sw_logging",
  payload: {shouldLog: true}
})
```

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/628)

### DIFF
--- a/src/service-worker/ServiceWorker.ts
+++ b/src/service-worker/ServiceWorker.ts
@@ -17,7 +17,7 @@ import {
   UpsertSessionPayload, DeactivateSessionPayload,
   PageVisibilityRequest, PageVisibilityResponse, SessionStatus
 } from "../models/Session";
-import Log from "../libraries/Log";
+import Log from "../libraries/sw/Log";
 import { ConfigHelper } from "../helpers/ConfigHelper";
 import { OneSignalUtils } from "../utils/OneSignalUtils";
 import { Utils } from "../context/shared/utils/Utils";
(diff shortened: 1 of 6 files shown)


src/service-worker/ServiceWorker.ts
```diff
@@ -380,8 +389,8 @@ export class ServiceWorker {
       // Test if this window client is the HTTP subdomain iFrame pointing to subdomain.onesignal.com
       if (client.frameType && client.frameType === 'nested') {
         // Subdomain iFrames point to 'https://subdomain.onesignal.com...'
-        if (!Utils.contains(client.url, SdkEnvironment.getOneSignalApiUrl().host) &&
-            !Utils.contains(client.url, '.os.tc')) {
+        if (!Utils.contains(client.url, '.os.tc') &&
```
jkasten2, line 392: Does removing `SdkEnvironment.getOneSignalApiUrl().host` break dev evn? `localhost` testing?

PR #630, opened 2020-04-17
Prompt Delays

rgomezp:
Clean up https://github.com/OneSignal/OneSignal-Website-SDK/pull/616

## Context
Many sites that implement OneSignal for web push ask for permission immediately when a user reaches their site for the first time. 

This annoying for the end-user (who doesn’t have context on if the site has useful content yet), and may be suboptimal for the site owner (who may lose potential opt-ins if users had more context)

Currently, implementing time or session based delays requires a site owner to write custom code and use the  ‘custom site setup’ instead of  ‘typical site setup’ 

Building time or session delay into ‘typical site setup’ makes OneSignal more useful to site owners and less frustrating to end users

### REMOTE PARAMS

The “sync” call already contains the following configuration parameters:

```
{
   config: {
      staticPrompts: {
         native: {},
         bell: {},
         slidedown: {}
      }
   }
}

```
We can add two new parameters to the prompt configuration objects: **`timeDelay`** and **`pageViews`**

Example:

```
{
   config: {
      staticPrompts: {
         native: {timeDelay: 5, pageViews: 3},
      }
   }
}
```

Initialization already handles getting the configuration options:
```

OneSignal.init

OneSignal.initializeConfig

ConfigManager().getAppConfig

```
These values can then be accessed from the OneSignal context with

```
OneSignal.config.userConfig.promptOptions.slidedown.<value>
" "                                      .native.pageViews
" "                                      .slidedown.timeDelay
...etc
```

##### `incrementPageViewCount`: 
added a "total" pageView count that persists past individual sessions. Gets written to Local Storage to then be used by prompting logic.

##### `getLocalPageViewCount` : 
getter

##### `setLocalPageViewCount` : 
setter

##### `DelayedPromptType`: 
defines two types of delayed prompts: 1) `native` and `slidedown`
##### `internalShowAutoPrompt`:
now checks `pageView` condition and uses `showDelayedPrompt` to prompt either the native or slidedown prompts

##### `internalShowDelayedPrompt`: 
encapsulates internal prompting methods for both native or slidedown along with a `setTimeout` in case there is a `timeDelay` passed into the method
##### `isPageViewConditionMet`: 
checks whether the local `pageView` count is greater than or equal to the `pageView` value set for the site as part of the configuration options that are 1) either passed in from the server/backend or 2) from the user config (initialization options)
##### `getDelayedPromptOptions`: 
getter for either the `native` or `slidedown` prompting options from the `promptOptions` object

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/630)


src/helpers/page/WrapSetTimeoutInPromise.ts
```diff
@@ -0,0 +1,13 @@
+
+export async function wrapSetTimeoutInPromise(callback: Function, timeDelaySeconds: number): Promise<void> {
+    await new Promise((resolve, reject) => {
+    setTimeout(async () => {
+        try {
+            await callback();
+            resolve();
+        } catch(e) {
+            reject(e);
```
jkasten2, line 9: Don't think a catch and reject is needed here. If the callback throws it will bubble up in the same way I believe.

src/helpers/page/WrapSetTimeoutInPromise.ts
```diff
@@ -0,0 +1,13 @@
+
+export async function wrapSetTimeoutInPromise(callback: Function, timeDelaySeconds: number): Promise<void> {
```
jkasten2, line 2: I don't understand why this takes in a callback and also returns a `Promise`.

It looks like we are trying to have a version of `setTimeout` we can use with `await`. If this is the case then this would be a simpler way to do this.
```typescript
function awaitableTimeout(ms: number) :Promise<void> {
    return new Promise(resolve => setTimeout(resolve, ms));
}
```

src/helpers/page/WrapSetTimeoutInPromise.ts
```diff
@@ -0,0 +1,13 @@
+
```
jkasten2, line 1: Could use a comment on what this does and example.

test/support/sdk/TestEnvironment.ts
```diff
@@ -240,6 +241,18 @@ export class TestEnvironment {
     (windowDef as any).TextEncoder = TextEncoder;
     (windowDef as any).TextDecoder = TextDecoder;
     (windowDef as any).isSecureContext = isSecureContext;
+
+    if (config.stubSetTimeout) {
```
jkasten2, line 245: Does always stubing setTimeout cause issues with other tests?

src/OneSignal.ts
```diff
@@ -362,6 +364,19 @@ export default class OneSignal {
...
 
+  /**
+   * Shows delayed prompt
+   * @PublicApi
+   * @param type - type of prompt: "native" | "slidedown"
+   * @param timeDelaySeconds - number of seconds to delay the prompt by
+   * @Example
+   *  OneSignal.push(function() { OneSignal.showDelayedPrompt("native", 20); });
+   */
+  public static async showDelayedPrompt(type: DelayedPromptType, timeDelaySeconds: number): Promise<void> {
```
jkasten2, line 375: I don't see lot of use in this `showDelayedPrompt` public method. It would be the same as them calling.
```javascript
OneSignal.push(function() {
   setTimeout(OneSignal.showNativePrompt, 20000);
});
```

I would prefer we remove this for this release, otherwise we will be stuck with this public API which will have to be supported for awhile.

src/helpers/page/WrapSetTimeoutInPromise.ts
rgomezp: The callback is for the actual call to show the native or slidedown prompt after the setTimeout. It's wrapped in a promise that is returned to the higher level method so that the entire `showDelayedPrompt` method can be awaited on. 

See https://github.com/OneSignal/OneSignal-Website-SDK/pull/630/files#diff-1d83af775103f5a84ed90068ae19c422R133

PR #631, opened 2020-04-20
Check require-user-interaction in showDelayedPrompt and change type

rgomezp:
This fixes the public method by checking if the browser requires user-interaction. If so, force the prompt type to be "slidedown"

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/631)


src/managers/PromptsManager.ts
```diff
@@ -124,6 +125,11 @@ export class PromptsManager {
       return;
     }
 
+    const { requiresUserInteraction } = EnvironmentInfoHelper.getEnvironmentInfo();
+    if (requiresUserInteraction && type==="native") {
```
jkasten2, line 129: nit, `type==="native"` should be `type === DelayedPromptType.Native`

rgomezp: true

PR #632, opened 2020-04-21
Address Delayed Prompt Changes Regarding SetTimeout Implementation

rgomezp:
Requested changes found in https://github.com/OneSignal/OneSignal-Website-SDK/pull/630

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/632)


src/managers/PromptsManager.ts
```diff
@@ -124,19 +125,24 @@ export class PromptsManager {
...
     switch(type){
       case DelayedPromptType.Native:
         if (timeDelaySeconds === 0) {
-          await this.internalShowNativePrompt();
-          break;
+          return this.internalShowNativePrompt();
         }
-        return wrapSetTimeoutInPromise(this.internalShowNativePrompt.bind(this), timeDelaySeconds);
+        await awaitableTimeout(timeDelaySeconds*1000);
+        return this.internalShowNativePrompt();
```
jkasten2, line 139: `this.internalShowNativePrompt` is being called twice in this block. To fix this and simplify this code I would suggest only calling `awaitableTimeout` if timeDelaySeconds > 0,

jkasten2: Also we can move `timeDelaySeconds` outside of each case since it is duplicated for each in the switch.

jkasten2: `break`s in switch statements are normally preferred. Returns in a switch statements would be nice if it as a helper function with only that, but in other places it might not be clear.

<details>
  <summary>With comments noted above I would suggest this:

*(Feel free to wait to expand this if you want to take an attempt at my above comments)*</summary>

```typescript
    if (timeDelaySeconds > 0) {
       await awaitableTimeout(timeDelaySeconds * 1_000);
    }

   switch (type) {
      case DelayedPromptType.Native:
        this.internalShowNativePrompt();
        break;
      case DelayedPromptType.Slidedown:
        this.internalShowSlidedownPrompt(options);
        break;
      default:
        Log.error("Invalid Delayed Prompt type");
    }
```
</details>

PR #636, opened 2020-04-27
Loosens origin check to allow www. variant and stricter protocol (https)

rgomezp:
The following error is currently be thrown by 150300 of the Web SDK when only one .www is expected in the config but it not in the current URL the user in on.

```
OneSignalError.js:18 Uncaught (in promise) Error: OneSignal: This web push config can only be used on https://www.colourinvasion.pt. Your current origin is https://colourinvasion.pt.
t @ OneSignalError.js:18
SdkInitError @ SdkInitError.js:26
e.checkRestrictedOrigin @ ConfigManager.js:52
(anonymous) @ ConfigManager.js:32
(anonymous) @ Subscription.js:1
(anonymous) @ Subscription.js:1
s @ Subscription.js:1
Promise.then (async)
u @ EmailDeviceRecord.js:7
s @ EmailDeviceRecord.js:7
Promise.then (async)
u @ EmailDeviceRecord.js:7
(anonymous) @ EmailDeviceRecord.js:7
i @ EmailDeviceRecord.js:7
e.init @ OneSignal.js:159
(anonymous) @ (index):193
e.push @ OneSignal.js:664
t.oneSignalSdkInit @ pageSdkInit.js:31
(anonymous) @ sdk.js:12
n @ bootstrap:19
(anonymous) @ bootstrap:68
(anonymous) @ bootstrap:68
```

This restriction should be loosened so .www won't cause an issue.

In addition we should allows allow the HTTPS variant of their site if they enter http:// for their site URL. However for security not the other way around.

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/636)


src/helpers/ConfigHelper.ts
```diff
@@ -65,7 +65,18 @@ export class ConfigHelper {
 
   public static doesCurrentOriginMatchConfigOrigin(configOrigin: string): boolean {
     try {
-      return location.origin === new URL(configOrigin).origin;
+      const configUrl = new URL(configOrigin);
```
jkasten2, line 68: Variable names are mostly good here but still hard to read as the number check being done here. There are a few options to clean this up.
1. Do early returns.
2. Make separate private static function in this file.
3. If generic enough some logic might be a good fit as a Until function.

src/helpers/ConfigHelper.ts
```diff
@@ -65,7 +65,18 @@ export class ConfigHelper {
 
   public static doesCurrentOriginMatchConfigOrigin(configOrigin: string): boolean {
```
jkasten2, line 66: We should list the conditions we are checking as a comment here, since this method does a lot more now.

src/helpers/ConfigHelper.ts
```diff
@@ -62,10 +62,25 @@ export class ConfigHelper {
...
     try {
-      return location.origin === new URL(configOrigin).origin;
+      const configUrl = new URL(configOrigin);
+
+      if (location.origin === configUrl.origin) {
+        return true;
+      }
+
+      const configURLWithout3W = new URL(configUrl.origin.split("www.").join(''));
```
jkasten2, line 81: Interesting, never though of `.split("www.").join('')` to remove a string. However `.replace("www.", "")` is easier to read.

src/helpers/ConfigHelper.ts
```diff
@@ -62,10 +62,25 @@ export class ConfigHelper {
...
+      const configOriginWithHttps = `https://${configURLWithout3W.host}`;
+      return location.origin === configOriginWithHttps;
```
jkasten2, line 83: Couldn't this be simplified to?
```Typescript
return location.origin === configURLWithout3W.origin;
```

jkasten2: Also `location.origin` could have `www.` in it, we should remove it from this too is compare with without `www.` for both URLs.

src/helpers/ConfigHelper.ts
```diff
@@ -62,10 +62,25 @@ export class ConfigHelper {
       }
     }
   }
-
+  /**
+   * Checks if current origin matches:
+   *  - config origin exactly
+   *  - variants:
+   *    - config origin without 'www.'
+   *    - config origin with stricter protocol (we should allow http sites to initialize with https but not viceversa)
```
jkasten2, line 70: I don't see the HTTP config allowing both `location` being on an HTTP and HTTPS in this method.

test/unit/managers/ConfigManager.ts
```diff
@@ -1,17 +1,32 @@
...
     httpOrHttps: HttpHttpsEnvironment.Https
   });
 });
 
+test.afterEach(function (_t: TestContext) {
+  sinonSandbox.restore();
+
+  OneSignal._initCalled = false;
+  OneSignal.__initAlreadyCalled = false;
+  OneSignal._sessionInitAlreadyRunning = false;
```
jkasten2, line 27: It looks like we are cleaning these 3 in `test/unit/public-sdk-apis/init.ts`. We should put this in a test helper somewhere so this can be shared.

test/unit/managers/ConfigManager.ts
```diff
@@ -157,3 +172,108 @@ test('should have enableOnSession flag', t => {
...
+  } catch (e) {
+    // pass
+  }
+
+  t.true(doesOriginMatchSpy.calledOnce);
+  t.false(doesOriginMatchSpy.returnValues[0]);
+  t.true(checkRestrictedOriginSpy.threw());
+});
+
+test('should throw if origins do not match', async t => {
```
jkasten2, line 241: Test should explain testing config does not have `.www` but the page does.

test/unit/managers/ConfigManager.ts
```diff
@@ -157,3 +172,108 @@ test('should have enableOnSession flag', t => {
...
+  const checkRestrictedOriginSpy = sinonSandbox.spy(ConfigHelper, "checkRestrictedOrigin");
+  const doesOriginMatchSpy = sinonSandbox.spy(ConfigHelper, "doesCurrentOriginMatchConfigOrigin");
+  sinonSandbox.stub(OneSignal, "delayedInit");
+  try{
+    await initOneSignalOnUrlWithConfigOrigin("https://www.example.com", "https://example.com");
```
jkasten2, line 246: This is safe since host is the same for both and they are both HTTPS. However not sure if you can subscribe and / or have two different service workers with and without `www.`. After we figure this out we will know if we should support this or not a make a note here.

test/unit/managers/ConfigManager.ts
```diff
@@ -157,3 +172,108 @@ test('should have enableOnSession flag', t => {
...
+  };
+  const serverConfig = TestEnvironment.getFakeServerAppConfig(
+    ConfigIntegrationKind.TypicalSite,
+    true,
+    { features: {
+        restrict_origin: { enable: true }
+      },
+      config: { origin: configOrigin }
+    },
+      appId
```
jkasten2, line 273: nit, remove the 2 extra spaces.

test/unit/managers/ConfigManager.ts
```diff
@@ -157,3 +172,108 @@ test('should have enableOnSession flag', t => {
...
+async function initOneSignalOnUrlWithConfigOrigin(url: string, configOrigin: string) {
+  const appId = Random.getRandomUuid();
+  const testConfig = {
+    httpOrHttps: HttpHttpsEnvironment.Https,
+    url: new URL(url)
+  };
+  const serverConfig = TestEnvironment.getFakeServerAppConfig(
+    ConfigIntegrationKind.TypicalSite,
+    true,
+    { features: {
```
jkasten2, line 268: nit, move `features: {` down a line

src/helpers/ConfigHelper.ts
rgomezp: For your first, I don't think that would work. We want to allow HTTPS sites (location origin) to be able to load OneSignal configured with HTTP (config origin) but not viceversa.

If we only strip 3w from the config, we could have a case where:
- `location.origin` = "https://site.com"
- `configURLWithout3W.origin` = "http://site.com"

Then your proposed simplification would break.

But if we default the comparison to always make use of the config origin with HTTPS, then we would have:
1) in the case that the location origin is HTTPS and true config is HTTP
- `location.origin` = "https://site.com"
- `configOriginWithHttps` = "https://site.com" (even though the true value isn't https)

result: match. HTTPS -> HTTP = okay

2) in the case that the location origin is HTTP and true config origin is HTTPS
- `location.origin` = "http://site.com"
- `configOriginWithHttps` = "https://site.com" (matches true config value)

result: doesn't match. HTTP -> HTTPS = insecure

So the idea here is to strip 3w and then make a unidirectional comparison that only allows HTTPS->HTTP.

Per your second comment, I agree we should remove from both URLs

src/helpers/ConfigHelper.ts
rgomezp: I think I should have said, "weaker" not "stricter". Does that address your comment?

test/unit/managers/ConfigManager.ts
```diff
@@ -157,3 +170,125 @@ test('should have enableOnSession flag', t => {
...
+  } catch (e) {
+    // pass
+  }
+
+  t.true(doesOriginMatchSpy.calledOnce);
+  t.true(doesOriginMatchSpy.returnValues[0]);
+  t.false(checkRestrictedOriginSpy.threw());
+});
+
+test('should match origins with www. variant on page', async t => {
```
jkasten2, line 207: Seems www. and non-www are different origins according to this.
https://developer.mozilla.org/en-US/docs/Glossary/origin

HTTPS needs be strict if "My site is fully HTTPS" is used without a subdomain.

rgomezp:
Decided to handle this validation on frontend entirely

PR #643, opened 2020-05-18
fix installWorker() annoying error message

h3xby:
we have a lot of `"installWorker - SW's 'controllerchange' fired but no state change!"` errors in our error tracking system. i discovered that this happened during worker update. it updates successfully, but prints error. this happens due event listener doesn't removed and `this.installAlternatingWorker()` below will trigger this listener again, but we don't need it because promise has already been resolved

## 
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/643)


jkasten2:
@h3xby Thanks for the PR!
We made some big changes in #745 where the `controllerchange` is no longer used in the first place by this SDK now.

PR #666, opened 2020-07-10
Category slidedown feature 0710

itrush:
@jkasten2 @rgomezp Made a first pass at re-organizing the changes. Did not rebase on purpose. :)

Next steps would be:
- `git rebase -i --autosquash master` - to get rid of all fixups
- force push to this branch
- `git reset HEAD~1` to unstage all changes from the last commit and continue working on tests as necessary.

Happy to answer any questions and pair more on it.

Does not look so scary anymore :) only 30+ commits.

This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/666)


src/utils/TagUtils.ts
```diff
@@ -30,6 +30,10 @@ export default class TagUtils {
         // Going off local tags since it's our categories. Trying to find only changed tags and returning those
         // as a final object.
         Object.keys(localTags).forEach(key => {
+            if (remoteTags[key] === undefined) {
+                finalTags[key] = localTags[key];
+            }
```
itrush, line 36: this check already exists here `if (remoteTags[key] && ....) {`

src/utils/TagUtils.ts
```diff
@@ -66,8 +70,8 @@ export default class TagUtils {
             }
 
             return categoriesCopy.map(category => {
-                const existingTagValue = existingPlayerTags[category.tag];
-                category.checked = !!existingTagValue;
```
itrush: I don't see the reason for this rewrite. `category.checked = !!existingTagValue;` is even safer because it will always produce boolean. and it will handle undefined correctly too.

src/utils/TagUtils.ts
rgomezp: think about the case where there aren't any remote tags yet

src/utils/TagUtils.ts
rgomezp: same as above, think about the case where there aren't any remote tags yet. we want to default the checkbox to true. with the previous implementation, it would default to false

test/support/sdk/TestEnvironment.ts
```diff
@@ -28,6 +28,20 @@ import deepmerge = require("deepmerge");
 import { RecursivePartial } from '../../../src/context/shared/utils/Utils';
 import { EnvironmentInfo } from  '../../../src/context/browser/models/EnvironmentInfo';
 import { EnvironmentInfoHelper } from '../../../src/context/browser/helpers/EnvironmentInfoHelper';
+import { ExecutionContext } from 'ava';
```
itrush, line 31: recommended commit message structure
```
One line short summary
<empty line>
Detailed description of changes including motivation for it
for more complex scenarios.
```
Example:
```
Helper utils for OneSignal.init testing

1) TestEnvironment page setup with stubs helper
    Extracted logic we had in several places in tests into a helper function.
    Sets up stubs for common api calls occurring during initialization.
    Sets `OneSignal.initialized = false;` to simulate first time run.
2) getIcon nock util
    Simulates a successful request on icon api call, does not return actual image.
```

test/unit/prompts/Popover.ts
```diff
@@ -1,172 +0,0 @@
-import test from 'ava';
-import sinon, { SinonSandbox } from 'sinon';
-import Popover from '../../../src/popover/Popover';
```
itrush, line 3: there is a typo in the original commit message :( `Popever -> Popover`. We can fix it with reword at the final rebase.

src/slidedown/Slidedown.ts
```diff
@@ -17,13 +17,17 @@ import { getLoadingIndicatorWithColor } from './LoadingIndicator';
...
+import Log from 'src/libraries/Log';
 
 export default class Slidedown {
   public options: SlidedownPermissionMessageOptions;
   public notificationIcons: NotificationIcons | null;
-  private isInSaveState: boolean;
   public isShowingFailureState: boolean;
 
+  private isInSaveState: boolean;
+  private categoryOptions?: Categories;
```
itrush, line 29: `private categoryOptions?: Categories;` change this to be required because you always create them in constructor. then you don't need `isPrivateCategoryOptionsConfigured` check at all.

src/utils/TagUtils.ts
itrush: Got it. How about a shorter version of it then?
```
Object.keys(localTags).forEach(key => {
      // only if user's tag value did not change, skip it
      if (localTags[key] === remoteTags[key]) {
         return;
      }
      finalTags[key] = localTags[key];
});
```

src/utils/TagUtils.ts
itrush: got it. However it's not obvious from the code. How about a helper like this:
```
static getCheckedStatusForTagValue(tagValue: boolean | undefined): boolean {
        // If user does not have tag assigned to them, consider it selected
        if (tagValue === undefined) {
            return true;
        }

        return tagValue;
    }
```
It gives more room for explaining the logic and easy central place to change the logic in the future.

src/helpers/ConfigHelper.ts
```diff
@@ -312,7 +313,7 @@ export class ConfigHelper {
     };
     let categories;
     if (staticPrompts.slidedown.categories) {
-      categories = staticPrompts.slidedown.categories;
+      categories = TagUtils.limitCategoriesToMaxCount(staticPrompts.slidedown.categories, 10);
```
itrush, line 316: I'd rather do this as a separate commit, it is feature.

itrush: nevermind, it's a fixup to a correct commit.

src/slidedown/Slidedown.ts
rgomezp: Done.

src/utils/TagUtils.ts
rgomezp: Done.

src/utils/TagUtils.ts
```diff
@@ -76,4 +76,9 @@ export default class TagUtils {
                 return category;
             });
     }
+
+    static limitCategoriesToMaxCount(categories: Categories, max: number): Categories {
```
itrush, line 79: I'd recommend updating commit message
```
Limit the number of tag categories

For typical setup the number of categories is enforced in the dashboard.
However for custom code setup we can only do it on client side.
Limiting max number of categories to 10.
Added a util to take first 10 items from the provided array.
```

test/unit/prompts/Popover.ts
rgomezp: Done.

src/utils/TagUtils.ts
rgomezp: Done.

rgomezp:
# Closed in favor of #670

PR #674, opened 2020-07-22
Don't default to document title if no notification title present

gardenFiend138:
# Don't default to document title if no notification title present

### Detailed description

Notification titles used to be required in browsers. We allowed users to not provide a title, but if that happened, we defaulted to the document.title property. This is no longer a required field for notifications, and browsers handle it differently (i.e. Chrome says 'Google Chrome' if no title is present).

This continues to use the default title value if it is provided by the user by calling `OneSignal.setDefaultTitle`.

### Checklist:

- [x] I have included screenshots/recordings of the intended results

### Jira Ticket(s):
[OS-4666](https://onesignal.atlassian.net/browse/OS-4666)

### Screenshots/Gifs/Video of All Changes:
With no title or default title set:
no-title

Default title set, then with an actual notification title:
title

### Users Affected:

Estimated number of users this change will impact:

All end users who receive notifications without a set title

### Validation:

- Send notification without a title or default title
  - should see browser default title (i.e. 'Google Chrome' in Chrome)
- Send notification without a title, but after setting a default title (call `OneSignal.setDefaultTitle(<string>)`)
  - should see newly set default title
- Send notification with a title set
  - should see title set on notification

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/674)

### DIFF
--- a/src/OneSignal.ts
+++ b/src/OneSignal.ts
@@ -306,7 +306,6 @@ export default class OneSignal {
         MainHelper.checkAndTriggerNotificationPermissionChanged();
       });
 
-      await InitHelper.initSaveState(document.title);
       await InitHelper.saveInitOptions();
       if (SdkEnvironment.getWindowEnv() === WindowEnvironmentKind.CustomIframe)
         await Event.trigger(OneSignal.EVENTS.SDK_INITIALIZED);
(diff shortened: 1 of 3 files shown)


gardenFiend138:
Just realized need to address HTTP too

gardenFiend138:
This PR is being closed as @itrush came up with a solution that makes more sense. See [OS-4876](https://onesignal.atlassian.net/browse/OS-4876) for more details.

PR #681, opened 2020-07-29
Switch to use Web API in Slidedown and Tagging Container

rgomezp:
# Description
This change includes a transition from using hard-coded HTML strings to using the W3C Web API to create DOM elements programmatically. The main reason for this is for security purposes since this approach is an easy way to avoid injection vulnerabilities.

This PR also includes other tweaks like the addition of CSS and Element ID constants, 1 file rename, and other minor nits.

# Systems Affected
WebSDK only

# Validation

### Tests

Test coverage for these changes are not needed since the only change was a switch to how DOM elements are created.

### Screenshots

Nothing has visually changed in this PR.

---

This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/681)


src/slidedown/Slidedown.ts
```diff
@@ -127,83 +135,92 @@ export default class Slidedown {
...
       Log.debug("Slidedown private category options are not defined");
       return;
     }
 
     if (state) {
       // note: savingButtonText is hardcoded in constructor. TODO: pull from config & set defaults for future release
-      this.allowButton.innerHTML = this.getIndicatorHolderHtmlWithText(this.categoryOptions!.savingButtonText!);
-      addDomElement(this.buttonIndicatorHolder, 'beforeend',
-        getLoadingIndicatorWithColor(COLORS.whiteLoadingIndicator));
+      this.allowButton.innerHTML = this.getIndicatorHolderHtmlWithText(this.categoryOptions.savingButtonText).innerHTML;
```
jkasten2, line 146: nit, use [`replaceChild`](https://developer.mozilla.org/en-US/docs/Web/API/Node/replaceChild) over `innerHTML`.

src/slidedown/Slidedown.ts
```diff
@@ -127,83 +135,92 @@ export default class Slidedown {
...
+  setFailureState(state: boolean): void {
     if (!this.categoryOptions) {
       Log.debug("Slidedown private category options are not defined");
       return;
     }
 
     if (state) {
       // note: errorButtonText is hardcoded in constructor. TODO: pull from config & set defaults for future release
-      this.allowButton.innerHTML = this.getIndicatorHolderHtmlWithText(this.categoryOptions!.errorButtonText);
+      this.allowButton.innerHTML = this.getIndicatorHolderHtmlWithText(this.categoryOptions.errorButtonText).innerHTML;
```
jkasten2, line 169: nit, use [`replaceChild`](https://developer.mozilla.org/en-US/docs/Web/API/Node/replaceChild) over `innerHTML`.

src/slidedown/TaggingContainer.ts
```diff
@@ -8,102 +8,135 @@ import {
...
+        const wrappingDiv   = document.createElement("div");
+
+        addCssClass(labelElement, TAGGING_CONTAINER_CSS_CLASSES.categoryLabel);
+        addCssClass(labelSpan, TAGGING_CONTAINER_CSS_CLASSES.categoryLabelText);
+        addCssClass(inputElement, TAGGING_CONTAINER_CSS_CLASSES.categoryLabelInput);
+        addCssClass(checkmarkSpan, TAGGING_CONTAINER_CSS_CLASSES.checkmark);
+
+        labelElement.title = label;
+        labelSpan.innerText = label;
+        inputElement.type    = "checkbox";
```
jkasten2, line 102: nit,  inconsistent spaces on these lines

test/support/sdk/TestEnvironment.ts
```diff
@@ -221,13 +220,14 @@ export class TestEnvironment {
     }
 
     let html = '<!doctype html><html><head></head><body></body></html>';
+    global.document = jsdom.jsdom('');
```
jkasten2, line 223: Why was this line required?

rgomezp: the global document object must be defined for `getSlidedownElement` to work correctly (below). this line initializes the document object so that function works.

jkasten2: ah, gotcha 👍

src/slidedown/Slidedown.ts
jkasten2: Looking at this again [`replaceWith`](https://developer.mozilla.org/en-US/docs/Web/API/ChildNode/replaceWith) would probably be a better fit in this case.

src/slidedown/Slidedown.ts
jkasten2: Looking at this again [`replaceWith`](https://developer.mozilla.org/en-US/docs/Web/API/ChildNode/replaceWith) would probably be a better fit in this case.

rgomezp: Ah, yep that makes more sense

src/slidedown/Slidedown.ts
rgomezp: Ah, yep that makes more sense

src/slidedown/Slidedown.ts
rgomezp: I ended up switching to use `insertAdjacentElement` ([mdn docs](https://developer.mozilla.org/en-US/docs/Web/API/Element/insertAdjacentElement)) to individually insert the children nodes. I had to also separate the helper function into two to return the span element and indicator holder separately.

src/slidedown/Slidedown.ts
rgomezp: I ended up switching to use `insertAdjacentElement` ([mdn docs](https://developer.mozilla.org/en-US/docs/Web/API/Element/insertAdjacentElement)) to individually insert the children nodes. I had to also separate the helper function into two to return the span element and indicator holder separately.

src/slidedown/Slidedown.ts
```diff
@@ -143,14 +143,18 @@ export default class Slidedown {
 
     if (state) {
       // note: savingButtonText is hardcoded in constructor. TODO: pull from config & set defaults for future release
-      this.allowButton.innerHTML = this.getIndicatorHolderHtmlWithText(this.categoryOptions.savingButtonText).innerHTML;
       this.allowButton.disabled = true;
+      this.allowButton.textContent = null;
```
jkasten2, line 147: Is it necessary to remove the `textContent`?

rgomezp: Yeah, otherwise it will just get appended to the end and we'll have the previous text plus the new text and the spinning indicator

jkasten2: 👍 ah gotcha

rgomezp:
- [x] Visually normal slidedown.
- [x] Visually category slidedown.
- [x] Saving categories saves correct values
- [x] Visually category slidedown when updating.

PR #683, opened 2020-08-11
HTTP Fix Controller is Undefined

rgomezp:
# Description
## 1 Line Summary
Fixes issue on HTTP sites introduced during Outcomes release where access to the service worker from an HTTP site isn't available

## Details
This fix addresses issue where `WorkerMessenger.js:118 Uncaught (in promise) TypeError: Cannot read property 'controller' of undefined` is thrown on HTTP sites. The reason this fails is because `isSecureContext` is `false` in the iframe context due to ancestry rules. In other words, the iframe is not considered secure because it is a child of the host (page) which is insecure (http) and so can't establish the messaging channel. From MDN:

`For example, even for a document delivered over TLS within an <iframe>, its context is not considered secure if it has an ancestor that was not also delivered over TLS.`
([source](https://developer.mozilla.org/en-US/docs/Web/Security/Secure_Contexts))

To resolve, we switch to use `getInterpretedAmbiguousPermission` which is a workaround to Chrome 62+'s permission ambiguity for "denied" permissions, we assume the permission is "default" until we actually record the permission being "denied" or "granted". This allows our best-effort approach to subscribe new users, and upon subscribing, if we discover the actual permission to be denied, we record this for next time.

# Systems Affected
   - [x] WebSDK
   - [ ] Backend
   - [ ] Dashboard

# Validation
### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [ ] No Typescript warnings
   - [x] Avoid silencing null/undefined warnings by with the explanation point

Other:

### Checklist

---

## Related Tickets

---

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/683)

### DIFF
--- a/src/managers/sessionManager/page/SessionManager.ts
+++ b/src/managers/sessionManager/page/SessionManager.ts
@@ -41,7 +41,8 @@ export class SessionManager implements ISessionManager {
     ) {
       Log.debug("Notify SW to upsert session");
       await this.context.workerMessenger.unicast(WorkerMessengerCommand.SessionUpsert, payload);
-    } else if (this.context.environmentInfo.isUsingSubscriptionWorkaround) {
+    } else if (this.context.environmentInfo.canTalkToServiceWorker &&
+        this.context.environmentInfo.isUsingSubscriptionWorkaround) {
       Log.debug("Notify iframe to notify SW to upsert session");
       await OneSignal.proxyFrameHost.runCommand(OneSignal.POSTMAM_COMMANDS.SESSION_UPSERT, payload);
     } else { // http w/o our iframe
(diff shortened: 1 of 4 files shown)


src/managers/sessionManager/page/SessionManager.ts
```diff
@@ -222,6 +222,11 @@ export class SessionManager implements ISessionManager {
       return;
     }
 
+    if (!this.context.environmentInfo.canTalkToServiceWorker) {
+      Log.debug("Not setting session event listeners. No SW possible 2.");
```
jkasten2, line 226: Would be more useful with a more detailed error message such as.
```
Not setting session event listeners, can't talk to ServiceWorker due being hosted on an HTTP page.
```

src/managers/sessionManager/page/SessionManager.ts
```diff
@@ -193,7 +193,12 @@ export class SessionManager implements ISessionManager {
       this.context.environmentInfo.isBrowserAndSupportsServiceWorkers ||
       this.context.environmentInfo.isUsingSubscriptionWorkaround
     ) {
-      this.setupSessionEventListeners();
+      if (this.context.environmentInfo.canTalkToServiceWorker) {
```
jkasten2, line 198: This seems backwards, if we can talk to the sw shouldn't we call `this.setupSessionEventListeners()` to setup the message with the sw?

rgomezp: I think you're right. Fixed

src/managers/sessionManager/page/SessionManager.ts
jkasten2: Confirmed fixed

PR #684, opened 2020-08-13
Use site name as default notification title

itrush:
## One Line Summary
Use site name as default notification title

## Details
As described in [this epic](https://onesignal.atlassian.net/browse/OS-4876):

Currently if notification does not have a title specified in the dashboard, the web sdk first attempts to use `defaultTitle` provided via `setDefaultTitle` and then `pageTitle` which is set during OneSignal initialization as `document.title` of the page.

The problem is that if user subscribes for notifications for the whole site while being on some topic specific page, e.g. looking at the page with title `My Site Sports`, all their notifications will have this value used as a site's name going forward.

Solution:
Pass site name to web sdk (actually already happening) and use it in web sdk to save as the page title.
Order of title resolution:
actual notification title if present -> default title (provided via `setDefaultTitle`) -> page title

For http integration had to make this change on the dashboard - https://github.com/OneSignal/OneSignal/pull/4642.

## Screenshots
Screen Shot 2020-08-12 at 5 59 13 PM

---

## Related Tickets
https://onesignal.atlassian.net/browse/OS-4950

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/684)

### DIFF
--- a/src/OneSignal.ts
+++ b/src/OneSignal.ts
@@ -306,7 +306,7 @@ export default class OneSignal {
         MainHelper.checkAndTriggerNotificationPermissionChanged();
       });
 
-      await InitHelper.initSaveState(document.title);
+      await InitHelper.initSaveState();
       await InitHelper.saveInitOptions();
       if (SdkEnvironment.getWindowEnv() === WindowEnvironmentKind.CustomIframe)
         await Event.trigger(OneSignal.EVENTS.SDK_INITIALIZED);
--- a/src/OneSignal.ts
+++ b/src/OneSignal.ts
@@ -75,8 +75,8 @@ export default class OneSignal {
   }
 
   /**
-   * Sets the default title to display on notifications. Will default to the page's document.title
-   *  if you don't call this.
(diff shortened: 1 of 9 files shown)


src/OneSignal.ts
```diff
@@ -306,7 +306,9 @@ export default class OneSignal {
         MainHelper.checkAndTriggerNotificationPermissionChanged();
       });
 
-      await InitHelper.initSaveState(document.title);
+      await InitHelper.initSaveState(
+        OneSignal.config && OneSignal.config.siteName ? OneSignal.config.siteName : document.title
```
jkasten2, line 310: After this change it isn't clear that is setting. We could assign the ternary result to a var that is descriptive such as `defaultNotificationTitle`.
```typescript
let defaultNotificationTitle = OneSignal.config && OneSignal.config.siteName ? OneSignal.config.siteName : document.title;
await InitHelper.initSaveState(defaultNotificationTitle);
```

PR #706, opened 2020-09-23
Build and publish Staging- files to S3 in Travis

terrynsun:
Staging- files are currently served from rails on staging-01, but we'd
like them to be served by turbine instead, as in prod, and to simplify
URL routing. This change pushes Staging- files to S3, so turbine can
download and serve them.

# Systems Affected
* [x] WebSDK
* [ ] Backend
* [ ] Dashboard

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/706)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,7 @@ cache:
 script:
   - yarn test:noWatch
   - yarn build:prod
+  - yarn build:staging
   # thanks to https://graysonkoonce.com/getting-the-current-branch-name-during-a-pull-request-in-travis-ci/
   - BRANCH=$(if [ "$TRAVIS_PULL_REQUEST" == "false" ]; then echo $TRAVIS_BRANCH; else echo $TRAVIS_PULL_REQUEST_BRANCH; fi)
   - if [ "$BRANCH" = "master" ]; then ./.travis/s3-simple-upload-all.sh build/releases s3://onesignal-build/websdk/$TRAVIS_COMMIT && ./.travis/s3-simple-upload-all.sh build/amp s3://onesignal-build/websdk/$TRAVIS_COMMIT && echo Successfully uploaded Web SDK artifacts for version $TRAVIS_COMMIT; else echo Not uploading artifacts for this build; fi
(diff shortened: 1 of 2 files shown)


RyanAmos (commented):
Will let @jkasten2  review this since he knows better wether there's any issues here.

.travis.yml
```diff
@@ -12,6 +12,7 @@ cache:
 script:
   - yarn test:noWatch
   - yarn build:prod
+  - yarn build:staging
```
jkasten2, line 15: The staging URLs need to be pass into this command as they are not provided in this repo. I recommend we provide them via environment variables via TravisCI.

The full command should looks something like:
```bash
BUILD_ORIGIN=${STAGING_DOMAIN} API=staging API_ORIGIN=${STAGING_DOMAIN} yarn build:staging
```

jkasten2: There is an extra step that needs to be done to correct source maps that are being done manually today.

```bash
sed -i 's/sourceMappingURL=OneSignal/sourceMappingURL=Staging-OneSignal/' Staging-*.js
sed -i 's/sourceMappingURL=OneSignal/sourceMappingURL=Staging-OneSignal/' Staging-*.css
```

jkasten2 (changes requested):

build/scripts/publish.sh
```diff
@@ -26,3 +26,8 @@ cp build/bundles/OneSignalSDKWorker.js.map build/releases/$PREFIX"OneSignalSDKWo
 
 cp build/bundles/OneSignalSDKStyles.css build/releases/$PREFIX"OneSignalSDKStyles.css"
 cp build/bundles/OneSignalSDKStyles.css.map build/releases/$PREFIX"OneSignalSDKStyles.css.map"
+
+if [ "$ENV" = "staging" ]; then
+  sed -i 's/sourceMappingURL=OneSignal/sourceMappingURL=Staging-OneSignal/' Staging-*.js
+  sed -i 's/sourceMappingURL=OneSignal/sourceMappingURL=Staging-OneSignal/' Staging-*.css
```
jkasten2, line 32: TravisCI failed due to not being able to find these files. I think we need to add `build/releases/` to the files being edited here.
Thinking this should fix it.
```
	  sed -i 's/sourceMappingURL=OneSignal/sourceMappingURL=Staging-OneSignal/' build/releases/Staging-*.js
    sed -i 's/sourceMappingURL=OneSignal/sourceMappingURL=Staging-OneSignal/' build/releases/Staging-*.css
```

jkasten2 (changes requested):
See my last comment above

jkasten2 (approved):

PR #713, opened 2020-10-21
Updated setEmail() and setExternalUserId() call to include external_user_id

grevolution:
…e checking for external user id

# Description
## Details
# Systems Affected
* [x] WebSDK
* [ ] Backend
* [ ] Dashboard

# Validation
### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:

* [x] No Typescript warnings

Other:

### Checklist

- - -
## Related Tickets
- - -
## 
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/713)

### DIFF
--- a/src/OneSignal.ts
+++ b/src/OneSignal.ts
@@ -105,6 +105,7 @@ export default class OneSignal {
 
     const appConfig = await Database.getAppConfig();
     const { deviceId } = await Database.getSubscription();
+    const externalUserId = await Database.getExternalUserId();
     const existingEmailProfile = await Database.getEmailProfile();
 
     if (appConfig.emailAuthRequired && !(options && options.emailAuthHash)) {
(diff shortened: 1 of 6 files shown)


grevolution:
fixes #713

rgomezp:
Howdy,
You linked this same PR when noting the issue that it fixes. Can you please include the correct issue for context?

src/OneSignal.ts
```diff
@@ -117,19 +118,26 @@ export default class OneSignal {
     }
 
     const isExistingEmailSaved = !!existingEmailProfile.emailId;
+    var sanitizedExternalUserId: string | undefined;
```
rgomezp, line 121: use `let` or `const`

grevolution:
> Howdy,
> You linked this same PR when noting the issue that it fixes. Can you please include the correct issue for context?

Done.

src/OneSignal.ts
grevolution: Done.

src/OneSignal.ts
```diff
@@ -117,19 +118,26 @@ export default class OneSignal {
     }
 
     const isExistingEmailSaved = !!existingEmailProfile.emailId;
+    let sanitizedExternalUserId: string | undefined;
+    if(externalUserId == null){
```
jkasten2, line 122: nit, missing space after both if and ), should be:
```typescript
if (externalUserId == null) {
```

jkasten2 (changes requested):
Left some feedback below.
Also would helpful to add some tests to cover this logic chane

jkasten2:
This is now fixed in PR #797 and has been released in version [151403](https://github.com/OneSignal/OneSignal-Website-SDK/releases/tag/151403).

PR #723, opened 2020-11-18
External id auth

rgomezp:
# Description
## Details
# Systems Affected
* [ ] WebSDK
* [ ] Backend
* [ ] Dashboard

# Validation
### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:

* [ ] No Typescript warnings

Other:

### Checklist

- - -
## Related Tickets
- - -
## 
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/723)


rgomezp:
Closing in favor of https://github.com/OneSignal/OneSignal-Website-SDK/pull/723

PR #725, opened 2020-11-24
Adds Authentication Identifier to WebSDK

rgomezp:
# Description
## 1 Line Summary
Adds ability to send authentication identifier to `setExternalId` in WebSDK, rename key in `setEmail`

## Details
# Systems Affected
* [x] WebSDK
* [ ] Backend
* [ ] Dashboard

# Validation
### Tests
* Renamed `emailAuthHash` in `setEmail` unit tests
* Checks that malformed auth hash throws an error in `externalUserId`

### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:

* [ ] No Typescript warnings

Other:

### Checklist

Screen Shot 2020-11-24 at 5 48 14 PM

Screen Shot 2020-11-24 at 5 51 03 PM

Screen Shot 2020-11-24 at 6 05 11 PM

- - -
## Related Tickets
- - -
## 
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/725)


src/OneSignal.ts
```diff
@@ -55,8 +55,8 @@ import { EnvironmentInfoHelper } from './context/browser/helpers/EnvironmentInfo
 import { EnvironmentInfo } from './context/browser/models/EnvironmentInfo';
 import { SessionManager } from './managers/sessionManager/page/SessionManager';
 import OutcomesHelper from "./helpers/shared/OutcomesHelper";
-import { OutcomeAttributionType, SentUniqueOutcome } from "./models/Outcomes";
-import { DelayedPromptType, AppUserConfigNotifyButton } from './models/Prompts';
+import { OutcomeAttributionType } from "./models/Outcomes";
+import { AppUserConfigNotifyButton } from './models/Prompts';
```
rgomezp, line 59: unneeded imports

src/OneSignal.ts
```diff
@@ -92,32 +92,52 @@ export default class OneSignal {
...
-    if (options && options.emailAuthHash && options.emailAuthHash.length !== 64) {
-      throw new InvalidArgumentError('options.emailAuthHash', InvalidArgumentReason.Malformed);
     }
-    await awaitOneSignalInitAndSupported();
+
+    const isIdentifierAuthHashDefined = options && !!options.identifierAuthHash;
+    const isEmailAuthHashDefined      = options && !!options.emailAuthHash;
+
+    const authHash = isIdentifierAuthHashDefined ? options.identifierAuthHash :
+    (isEmailAuthHashDefined ? options.emailAuthHash : undefined);
```
rgomezp, line 106: For backwards compatibility, `emailAuthHash` should be acceptable as well as newer `identifierAuthHash` standard

src/OneSignal.ts
```diff
@@ -92,32 +92,52 @@ export default class OneSignal {
...
+      const isEmailAuthHashIsMalformed    = isEmailAuthHashDefined && options.emailAuthHash.length !== 64;
+
+      if ( isIdentifierAuthHashMalformed ) {
+        throw new InvalidArgumentError('options.identifierAuthHash', InvalidArgumentReason.Malformed);
+      }
+
+      if ( isEmailAuthHashIsMalformed ) {
+        throw new InvalidArgumentError('options.emailAuthHash', InvalidArgumentReason.Malformed);
+      }
+    }
```
rgomezp, line 124: check for malformed hashes

src/OneSignal.ts
```diff
@@ -92,32 +92,52 @@ export default class OneSignal {
...
+    await awaitOneSignalInitAndSupported();
 
     const appConfig = await Database.getAppConfig();
     const { deviceId } = await Database.getSubscription();
     const existingEmailProfile = await Database.getEmailProfile();
 
-    if (appConfig.emailAuthRequired && !(options && options.emailAuthHash)) {
-      throw new InvalidArgumentError('options.emailAuthHash', InvalidArgumentReason.Empty);
-    }
+    const newEmailProfile = new EmailProfile(existingEmailProfile.emailId, email, authHash);
```
rgomezp, line 133: Motivation: Add `authHash` param to `EmailProfile`

src/OneSignal.ts
```diff
@@ -92,32 +92,48 @@ export default class OneSignal {
...
-    }
-
-    const newEmailProfile = new EmailProfile(existingEmailProfile.emailId, email);
-    if (options && options.emailAuthHash) {
-      newEmailProfile.emailAuthHash = options.emailAuthHash
-    }
+    const newEmailProfile = new EmailProfile(existingEmailProfile.emailId, email, authHash);
 
     const isExistingEmailSaved = !!existingEmailProfile.emailId;
-    if (isExistingEmailSaved && appConfig.emailAuthRequired) {
```
rgomezp, line 120: Motivation: removed `emailAuthRequired` from remote config. The logic wasn't correct anyway. Whether or not this config was required shouldn't impact whether the call is an update call or a create call

src/OneSignal.ts
```diff
@@ -92,32 +92,48 @@ export default class OneSignal {
...
-    // emailAuthHash is expected to be a 64 character SHA-256 hex hash
-    if (options && options.emailAuthHash && options.emailAuthHash.length !== 64) {
-      throw new InvalidArgumentError('options.emailAuthHash', InvalidArgumentReason.Malformed);
     }
-    await awaitOneSignalInitAndSupported();
+
+    const isIdentifierAuthHashDefined = options && !!options.identifierAuthHash;
+    const isEmailAuthHashDefined      = options && !!options.emailAuthHash;
+
+    const authHash = isIdentifierAuthHashDefined ? options.identifierAuthHash :
```
rgomezp, line 105: Note: typescript warning thinks `identifierAuthHash` may not be defined

src/OneSignal.ts
rgomezp: Note: typescript warning thinks `emailAuthHash` may not be defined (discussed with Josh)

src/OneSignal.ts
jkasten2: 👍

src/OneSignal.ts
```diff
@@ -471,7 +487,7 @@ export default class OneSignal {
     if (emailProfile.emailId) {
       await OneSignalApi.updatePlayer(appId, emailProfile.emailId, {
         tags: tags,
-        email_auth_hash: emailProfile.emailAuthHash,
+        identifier_auth_hash: emailProfile.identifierAuthHash,
```
jkasten2, line 490: Just noting here, the OneSignal backend needs to be deployed before this PR is due to this change

jkasten2 (changes requested):
Need to add `exnteral_user_id_auth_hash` to all push player updates.
Such as here:
https://github.com/OneSignal/OneSignal-Website-SDK/pull/725/files#diff-5e7bac54c465703b51d9b17f06118b85544d03dfb319ea34bf01be84d2a3e6b1R500-R502

src/OneSignal.ts
```diff
@@ -471,16 +487,16 @@ export default class OneSignal {
...
-    var { deviceId } = await Database.getSubscription();
+    const { deviceId } = await Database.getSubscription();
     if (!deviceId) {
       await awaitSdkEvent(OneSignal.EVENTS.REGISTERED);
     }
     // After the user subscribers, he will have a device ID, so get it again
-    var { deviceId: newDeviceId } = await Database.getSubscription();
+    const { deviceId: newDeviceId } = await Database.getSubscription();
     await OneSignalApi.updatePlayer(appId, newDeviceId!, {
       tags: tags
```
rgomezp, line 501: I imagine we should also add `external_id_auth_hash` here correct @jkasten2 ?

jkasten2:
@rgomezp Some tests failing currently. Looks like we just need to update the tests.
  ✖ public-sdk-apis › sendTags › sendTags sends to email record and push record with email auth hash 
  ✖ public-sdk-apis › sendTags › sendTags sends to email record and push record without email auth hash 
  ✖ public-sdk-apis › sendTags › sendTags sends to push record only without email 
https://travis-ci.org/github/OneSignal/OneSignal-Website-SDK/builds/745988305#L663-L665

jkasten2 (changes requested):
Looked through all the calls to `OneSignalApi.updatePlayer`, we also need `external_user_id_auth_hash` for `setSubscription` as well like we have for `sendTags`.

src/OneSignal.ts
jkasten2: yes, confirmed you have added it now

src/OneSignal.ts
```diff
@@ -587,10 +587,8 @@ export default class OneSignal {
       Log.warn("User is not subscribed, cannot remove external user id.");
       return;
     }
-    await Promise.all([
-      OneSignal.database.setExternalUserId(undefined),
-      OneSignal.context.updateManager.sendExternalUserIdUpdate(undefined),
-    ]);
+    await OneSignal.context.updateManager.sendExternalUserIdUpdate(undefined),
```
jkasten2, line 590: Why did we remove `Promise.all` here?

jkasten2 (approved):

src/OneSignal.ts
jkasten2: Talked about this, the order matters. This is because we need to ensure that `external_id_auth_hash` is in the local db as it will be read and sent in the REST API call. Once it is done then we can remove it with the next line from the local db.

PR #727, opened 2020-12-01
Upgrade typescript

rgomezp:
# Description
## 1 Line Summary
Upgrades Typescript version to 3.9.7

# Systems Affected
* [x] WebSDK
* [ ] Backend
* [ ] Dashboard

# Validation
## Tests
All tests pass

### Info
Fixed Unit Tests that broke due to Typescript changes

### Checklist

**Programming Checklist**
Interfaces:

  * New interfaces are unique to each individual testing file

Functions:

Typescript:

* [x] No Typescript warnings

Other:

### Info
No relevant screenshots necessary

### Checklist

- - -
## Related Tickets
- - -
## 
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/727)

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -376,14 +376,10 @@ export function getSdkLoadCount() {
   return (<any>window).__oneSignalSdkLoadCount || 0;
 }
 
-export async function awaitSdkEvent(eventName: string, predicate?: Action<any>) {
+export async function awaitSdkEvent(eventName: string) {
   return await new Promise(resolve => {
     OneSignal.emitter.once(eventName, (event: Event) => {
-      if (predicate) {
-        const predicateResult = predicate(event);
-        if (predicateResult)
-          resolve(event);
-      } else resolve(event);
+      resolve(event);
     });
   });
 }
--- a/package.json
+++ b/package.json
@@ -81,7 +81,7 @@
     "svgo": "^0.7.2",
     "text-encoding": "^0.6.4",
     "timemachine": "^0.3.0",
-    "typescript": "^2.7.1",
+    "typescript": "3.0.0",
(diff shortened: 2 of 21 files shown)


src/utils.ts
```diff
@@ -379,11 +379,9 @@ export function getSdkLoadCount() {
 export async function awaitSdkEvent(eventName: string, predicate?: Action<any>) {
   return await new Promise(resolve => {
     OneSignal.emitter.once(eventName, (event: Event) => {
-      if (predicate) {
-        const predicateResult = predicate(event);
-        if (predicateResult)
-          resolve(event);
-      } else resolve(event);
+      if (!predicate) {
```
jmadler, line 382: This seems like a meaningful semantic shift -- what's the intent here?

rgomezp: See the commit for [more details](https://github.com/OneSignal/OneSignal-Website-SDK/commit/ed16f5710b57a268401ae29ae59b1f4bca08cf3b)

It looks like `predicateResult` is always of type `void` anyway.

jmadler: Makes sense, but it looks like we're no longer called predicate when present.  Is that avoidable?

rgomezp: Decided on removing predicate altogether....

```
export async function awaitSdkEvent(eventName: string) {
  return await new Promise(resolve => {
    OneSignal.emitter.once(eventName, (event: Event) => {
      resolve(event);
    });
  });
}
```

This function is not used very much and it never actually receives a `predicate`.

rgomezp:
Closing in favor of cleaned up PR:  https://github.com/OneSignal/OneSignal-Website-SDK/pull/735

PR #735, opened 2020-12-24
Upgrade Typescript to 3.9.7

rgomezp:
# Description
## 1 Line Summary
Upgrades Typescript version to 3.9.7

# Systems Affected
* [x] WebSDK
* [ ] Backend
* [ ] Dashboard

# Validation
## Tests
All tests pass

### Info
Fixed Unit Tests that broke due to Typescript changes

### Checklist

**Programming Checklist**
Interfaces:

  * New interfaces are unique to each individual testing file

Functions:

Typescript:

* [x] No Typescript warnings

Other:

### Info
No relevant screenshots necessary

### Checklist
* 
* 
* 

## Related Tickets
- - -
## 
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/727)

### DIFF
--- a/package.json
+++ b/package.json
@@ -81,7 +81,7 @@
     "svgo": "^0.7.2",
     "text-encoding": "^0.6.4",
     "timemachine": "^0.3.0",
-    "typescript": "^2.7.1",
+    "typescript": "^3.9.7",
     "tslint": "^5.10.0",
     "tslint-eslint-rules": "^5.2.0",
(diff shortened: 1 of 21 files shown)


itrush (approved):

jkasten2:
@rgomezp I believe we can removing some additional entries in `typings/globals/missing.d.ts` now that we are using TS 3.9.7. Some were specifically present since we were still on an old TS version.

PR #741, opened 2021-02-04
Added IANA timezone

gonzalonarbaiz:
# Description
Added IANA timezone ID 

## Details
This change is required to prevent a bug we are currently having with TZ

# Systems Affected
* [x] WebSDK

## 
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/741)

### DIFF
--- a/src/models/UpdatePlayerOptions.ts
+++ b/src/models/UpdatePlayerOptions.ts
@@ -7,7 +7,7 @@ export interface UpdatePlayerOptions {
     web_p256                    ?: string;
     language                    ?: string;
     timezone                    ?: number;
-    timezone_id                    ?: string;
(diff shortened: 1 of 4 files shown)


src/models/UpdatePlayerOptions.ts
```diff
@@ -7,6 +7,7 @@ export interface UpdatePlayerOptions {
     web_p256                    ?: string;
     language                    ?: string;
     timezone                    ?: number;
+    timezone_id                    ?: string;
```
rgomezp, line 10: Nit: spacing doesn't match other params

rgomezp (approved):

PR #745, opened 2021-02-12
Remove ServiceWorker Page Control Requirement

jkasten2:
# Description
## 1 Line Summary
Remove ServiceWorker page control requirement.

### Summary
**_This is a quick 2 minute summary, recommend also seeing the [full spike work here](https://github.com/OneSignal/OneSignal-Website-SDK/issues/615) after reading this summary._**
Removes the requirement of the service worker scope to be at or higher than the page you are viewing. (AKA the ServiceWorker has control of the page)

* _You may have known you could[ already customize the path and scope of the OneSignal ServiceWorker ](https://documentation.onesignal.com/docs/onesignal-service-worker-faq#custom-code-setup---service-worker-customizations) before this but it still meant the user has to be on a page in it's scope._

APIs that assumed only one ServiceWorker is on a domain has been replaced with `navigator.serviceWorker.getRegistration(scope)`.

* Examples; `navigator.serviceWorker.controller` & `navigator.serviceWorker.ready`

Dev Sandbox has been updated to a non-root scope to demonstrate these changes work.

#### What this means
* No need to merge ServiceWorker files per the  ["Integrating Multiple Service Workers"](https://documentation.onesignal.com/docs/onesignal-service-worker-faq#section-integrating-multiple-service-workers) guide to support the following.
* Compatibility with AMP, when using your own domain (AKA not Google proxied)
* Compatibility with PWA
* And any other service worker that requires page control

#### What is PR does NOT do
* Does not change the default scope
  * This is intended so there is no breaking changes to those who have already setup this SDK. A change in the default will be done in the future.
* Does not allow more than one OneSignal ServiceWorker per domain. Example `en/push/onesignal` and `es/push/onesignal`
     - This is because the same index.db, sessionStorage, and others would be shared.

### [Full Details](https://github.com/OneSignal/OneSignal-Website-SDK/issues/615)
See issue #615 for the full research done before this PR was created.

# Systems Affected
* [x] WebSDK
* [ ] Backend
* [ ] Dashboard

# Validation
## Tests
Tests have been updated and pass. Need to evaluate changes if we some new tests.

### Checklist
* [X] All the automated tests pass or I explained why that is not possible
* [X] I have personally tested this on my machine or explained why that is not possible
  * Tested Edge(Chromium), and Firefox on Windows 10.
  * Tested Chrome and Safari on macOS
  * Tested both HTTP and HTTPS sites
  * Tested ServiceWorker upgrading versions
  * Tested on_focus and on_session works
[...]

PR #747, opened 2021-02-19
Update ServiceWorker if its href or scope changes

jkasten2:
# Description
## 1 Line Summary
Update ServiceWorker if its href or scope changes. Allows a seamless transition for those who take advantage of the ["Remove ServiceWorker Page Control Requirement"](https://github.com/OneSignal/OneSignal-Website-SDK/pull/745) changes.

## Details
# Systems Affected
* [x] WebSDK
* [ ] Backend
* [ ] Dashboard

# Validation
## Tests
**Tested the following scenario:**

1. Subscribed for push with default OneSignal ServiceWorker settings.
   - OneSignalSDKWorker.js on the root scope
2. Closed site
3. Changed scope by setting `OneSignal.SERVICE_WORKER_PARAM` to `/push/onesignal/`
4. Also someone may add their own service work in the same deploy
    - Added a call to `navigator.serviceWorker.register("pwa-sw.js");` to replace OneSignal at the root scope.
5. Opened site, observed both ServiceWorkers installed at different scopes.
6. Sent a push and observed it displayed correctly.

Tested on Edge and Firefox on Windows 10.

### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:

* [x] No Typescript warnings

Other:

# Info
### Question if two different SW files have `self.addEventListener('push'` which one gets the event?
Every `ServiceWorkerRegistration` instance has its own subscription endpoint. So the answer is simply which one you sent the push to..

## Checklist
* [x] First time setup
* [x] Site was on root scope, changes to a non-root scope such as `/push/onesignal/`
* [x] Site has a root worked with both their SW `importScripts` and ours, want to migrate to OneSignal being under it own scope.
* [x] Ensure pushes display, and don't display twice.

- - -
## Related Tickets
PR #745
- - -
## 
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/747)

### DIFF
--- a/src/managers/ServiceWorkerManager.ts
+++ b/src/managers/ServiceWorkerManager.ts
@@ -136,34 +136,38 @@ export class ServiceWorkerManager {
   }
 
   private async shouldInstallWorker(): Promise<boolean> {
+    // 1. Does the browser support ServiceWorkers?
     if (!Environment.supportsServiceWorkers())
       return false;
 
+    // 2. Is OneSignal initialized?
     if (!OneSignal.config)
       return false;
 
+    // 3. Will the service worker be installed on os.tc instead of the current domain?
     if (OneSignal.config.subdomain) {
       // No, if configured to use our subdomain (AKA HTTP setup) AND this is on their page (HTTP or HTTPS).
       // But since safari does not need subscription workaround, installing SW for session tracking.
       if (
-        OneSignal.environmentInfo.browserType !== "safari" && 
+        OneSignal.environmentInfo.browserType !== "safari" &&
           SdkEnvironment.getWindowEnv() === WindowEnvironmentKind.Host
       ) {
         return false;
       }
     }
 
+    // 4. Is a OneSignal ServiceWorker not installed now?, if not and
+    //   notification permissions are enabled we should install.
+    //   This prevents an unnecessary install which saves bandwidth
     const workerState = await this.getActiveState();
-    // If there isn't a SW or it isn't OneSignal's only install our SW if notification permissions are enabled
-    // This prevents an unnessary install which saves bandwidth
     if (workerState === ServiceWorkerActiveState.None || workerState === ServiceWorkerActiveState.ThirdParty) {
       const permission = await OneSignal.context.permissionManager.getNotificationPermission(
         OneSignal.config!.safariWebId
       );
-
       return permission === "granted";
     }
 
+    // 5. We have a OneSignal ServiceWorker installed, is there an update?
     return this.workerNeedsUpdate();
   }
 
--- a/src/managers/ServiceWorkerManager.ts
+++ b/src/managers/ServiceWorkerManager.ts
@@ -156,9 +156,9 @@ export class ServiceWorkerManager {
       }
     }
 
-    // 4. Is a OneSignal ServiceWorker not installed now?, if not and
-    //   notification permissions are enabled we should install.
-    //   This prevents an unnecessary install which saves bandwidth
+    // 4. Is a OneSignal ServiceWorker not installed now?
+    // If not and notification permissions are enabled we should install.
+    // This prevents an unnecessary install of the OneSignal worker which saves bandwidth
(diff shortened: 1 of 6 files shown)


src/managers/ServiceWorkerManager.ts
```diff
@@ -136,37 +136,91 @@ export class ServiceWorkerManager {
...
+      const notificationsEnabled = permission === "granted";
+      if (notificationsEnabled) {
+        Log.info("[shouldInstallWorker] Notification Permissions enabled, will install ServiceWorker");
+      }
+      return notificationsEnabled;
+    }
 
-      return permission === "granted";
+    // 5. We have a OneSignal ServiceWorker installed, but did the path or scope of the ServiceWorker change?
+    if (await this.changedServiceWorkerParams()) {
```
jkasten2, line 176: Future PR: `ServiceWorkerActiveState.ThirdParty` check isn't really needed if we are checking for the name already in this `changedServiceWorkerParams` method.

jkasten2: Future PR: `ServiceWorkerActiveState.None` could also be cleaned up, since this checks for presence. Consider if it is worth keeping enum however so it is easy to search for.

src/managers/ServiceWorkerManager.ts
```diff
@@ -136,37 +136,91 @@ export class ServiceWorkerManager {
...
+      return true;
     }
 
+    // 6. We have a OneSignal ServiceWorker installed, is there an update?
     return this.workerNeedsUpdate();
   }
 
+  private async changedServiceWorkerParams(): Promise<boolean> {
```
jkasten2, line 184: Should change name to `haveParamsChanged`, to indicate it will return a `bool`

src/managers/ServiceWorkerManager.ts
```diff
@@ -136,37 +136,91 @@ export class ServiceWorkerManager {
...
+    }
+
+    // 3. Different href?, asking if (path + filename [A or B] + queryParams) is different
+    const availableWorker = ServiceWorkerUtilHelper.getAvailableServiceWorker(workerRegistration);
+    const serviceWorkerHrefs = ServiceWorkerHelper.getPossibleServiceWorkerHrefs(
+      this.config,
+      this.context.appConfig.appId
+    );
+    // We don't care if the only differences is between OneSignal's A(Worker) vs B(WorkerUpdater) filename.
+    if (serviceWorkerHrefs.indexOf(availableWorker?.scriptURL || "") === -1) {
```
jkasten2, line 213: This line is doing to much. It also isn't a good idea to check for `""` in array to get a try to force a not found result.

jkasten2 (commented):
Would be nice if there was more background about the problem this is trying to solve and ServiceWorker state it results. Such as keeping the existing worker incase it has more than just a OneSignal importScripts in it. And how getting calling getRegistration("/scope") will get higher scope if a specific one isn't present.

itrush (approved):
> Reviewed 1 of 1 files at r1, 5 of 6 files at r2, 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/747)* status: all files reviewed, 3 unresolved discussions (waiting on @jkasten2 and @rgomezp)

___
*[src/managers/ServiceWorkerManager.ts, line 176 at r2](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/747#-MTuVo6LxwKd8Hjo27y3-r2-176:-MUGhwAb6ch4GApDQpfC:b4c9lyr) ([raw file](https://github.com/onesignal/onesignal-website-sdk/blob/873a47a1e4e6eeffd51335d1776503d4928282de/src/managers/ServiceWorkerManager.ts#L176)):*

Future PR: `ServiceWorkerActiveState.None` could also be cleaned up, since this checks for presence. Consider if it is worth keeping enum however so it is easy to search for.
</blockquote></details>

:thumbsup:

PR #752, opened 2021-02-23
Fix update push endpoint immediately

### LINKED ISSUE #715: re-allow notifications doesn't work (most likely still an issue in v16 [160000])
**Description:**

Notifications work as expected until we block them for a while, and then when we decide to allow them again, we can't get notifications at all - and that's weird.

Typical use cases: user gets notifications, but for the next day decided to block notification just not to be distracted, and after that decided to allow again to be notified, but user will no longer be able to receive them.

**Steps to Reproduce Issue:**

1) Wait for prompt to appear
2) Click Allow
3) Send notification it will work
3) Block notification in site settings
4) Reload the page
5) Send notification it will not work (as expected)
5) Unblock notification in site settings - press allow
6) Reload the page
7) Send a message it will not work (but... notifications allowed)
Can't receive notification at all after these steps

**Anything else:**
After steps above I can see that my user in Onesignal dashboard has status "Not Subscribed", and can't get notifications.

jkasten2:
# Description
## 1 Line Summary
Fix issue where if there are subscription changes when a new tab is opened (before the 30 second new session) they are not sent until the next on_session will tigger.

## Details
The root issue is session logic is different from the page vs SW logic.
*  page - `onSessionSent`
* SW - `timeSinceLastDeactivatedInSeconds <= sessionThresholdInSeconds`

The path forward will probably be moving all player updates (and create) calls to the service worker so they can be made from one central location.

# Systems Affected
* [ ] WebSDK
* [ ] Backend
* [ ] Dashboard

# Validation
### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:

* [ ] No Typescript warnings

Other:

### Checklist

- - -
## Related Tickets
- - -
## 
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/752)

Fixes #715

### DIFF
--- a/src/helpers/ServiceWorkerHelper.ts
+++ b/src/helpers/ServiceWorkerHelper.ts
@@ -12,6 +12,7 @@ import { OutcomesConfig } from "../models/Outcomes";
 import OutcomesHelper from './shared/OutcomesHelper';
 import { cancelableTimeout, CancelableTimeoutPromise } from './sw/CancelableTimeout';
 import { OSServiceWorkerFields } from "../service-worker/types";
+import Utils from "../context/shared/utils/Utils";
 
 declare var self: ServiceWorkerGlobalScope & OSServiceWorkerFields;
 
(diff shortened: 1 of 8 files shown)


jkasten2:
It is no longer planned to land this PR in 15.x.x so closing this PR. The large changes around User Model in the 16.x.x release means the changes proposed here no longer apply.

PR #754, opened 2021-02-26
Fix SW scope handling of base html tag 

jkasten2:
# Description
## 1 Line Summary
Fixed regression introduced in #745 where we are no longer correctly handling relative URLs for the ServiceWorker scope when a `<base>` HTML tag is added to the page.

## Details
Fixes issue where the page can't find the ServiceWorker if it is registered. Fixes the following warning + error printed to the console.
```
[Service Worker Status] Error Checking service worker registration / DOMException: Failed to get a ServiceWorkerRegistration: The origin of the provided documentURL ('https://www.example.com') does not match the current origin ('https://localhost:4001').
[Worker Messenger] [Page -> SW] Could not get ServiceWorkerRegistration to postMessage!
```
Only API affected by the base tag we were using was `navigator.serviceWorker.getRegistration`.

# Systems Affected
   - [X] WebSDK

# Validation
### Checklist
      - Tested on Edge as a new user

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [X] No Typescript warnings

Other:

### Checklist
      - No visual changes.  

---

## Related Tickets
- PR #745 
---

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/754)

### DIFF
--- a/express_webpack/index.html
+++ b/express_webpack/index.html
@@ -1,5 +1,8 @@
 <!DOCTYPE html>
 <html lang="en">
+<!-- Some sites use this HTML base tag to change the origin for all relative links.
+    Ensure OneSignal SDK correctly handles this. -->
+<base href="https://www.example.com/">
 <script src="/sdks/Dev-OneSignalSDK.js" async=""></script>
 <script>
     const SERVICE_WORKER_PATH = "push/onesignal/";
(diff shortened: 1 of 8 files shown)


itrush (approved):
> Reviewed 1 of 1 files at r1, 7 of 7 files at r2.
> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/754)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @rgomezp)

PR #759, opened 2021-03-03
Prompting Logic - Web Prompts

rgomezp:
# Description
## 1 Line Summary
Prompting Logic changes needed for Web Prompts Implementation

### System Overview
1. The `InitHelper` `sessionInit` function calls a `PromptsManager` function that will "spawn" the various slidedowns based on their unique rules
1. The `PromptsManager` `spawnAutoPrompts` function will iterate over the configured slidedown options and call the appropriate `PromptsManager` show function (e.g: `internalShowDelayedPrompt`)
1. `internalShowDelayedPrompt` function will use the slidedown type to call the corresponding Slidedown type-specific function which will pass the relevant `AutoPromptOptions` to the `internalShowSlidedownPrompt` function
1. `internalShowSlidedownPrompt` will create the slidedown via the `SlidedownManager` `createSlidedown` function.
1. The `createSlidedown` function handles the rest, including mounting the slidedown and inserting any relevant containers such as the tagging container or channel capture containers (not included in this PR)

### `SlidedownManager`
The new manager has the functions

* `createSlidedown`
* `checkIfSlidedownShouldBeShown`
* `setIsSlidedownShowing`
* `showQueued`
* `enqueue`
* `dequeue`
* `mountSpecialContainers`
* `mountTaggingContainer`
* `mountChannelCaptureContainer` (not yet implemented)

### `Slidedown`
We are adding a new event "popoverQueued" that will trigger as soon as we enqueue a slidedown to the slidedown queue. This is useful in prod for debugging and in testing environment to unit test that specific numbers of slidedowns were enqueued correctly.

#### Members: `slidedownQueue`, `isSlidedownShowing`
This manager leverages the `context` object to store the latest manager state. This is needed to track a new `slidedownQueue` which is a private class member.

The class also uses `isSlidedownShowing` private member to set whether a slidedown is showing or not. This is then changed via a setter from the `PromptsManager` event listeners (e.g: SHOWN, CLOSED) to update the `SlidedownManager` private member boolean value.

# Systems Affected
* [x] WebSDK
* [ ] Backend
* [ ] Dashboard

# Validation
#### `SlidedownPrompting` Testing File
**Note:** test coverage not yet fully implemented. Changes should be added to new unit test file: `SlidedownPrompting`. These tests cover things like sequence of slidedown show events and other things like did OneSignal initialization "spawn" as many slidedown show function calls as configured.

#### `_helpers` testing file
In the future, I'd like to keep helper files near the files using them. Only when helper code is used by multiple files should it be moved to a higher directory. In this PR, I put the new `SlidedownPrompting` file in a directory of the same name along with the helper code in a `_helpers.ts` file. Ava ignores files with the `_helpers` for assertions...

Screen Shot 2021-03-02 at 6 25 37 PM

#### `EventTestHelper`
New function `simulateSlidedownAllow` to trigger the `ALLOW_CLICK` event from unit tests at a specific time (e.g: after initialization via a promise, as opposed to immediately after 'SHOWN' event)

### Checklist
  * **Test coverage is not yet fully implemented in this PR**

**Programming Checklist**
Interfaces:

Functions:

Typescript:

* [x] No Typescript warnings

Other:

### Checklist

https://user-images.githubusercontent.com/11739227/109737582-e8b86000-7b8b-11eb-80ee-1652dad89508.mov

- - -
## Related Tickets
- - -

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/759)


jkasten2 (changes requested):
> Reviewed 1 of 1 files at r1, 1 of 1 files at r2, 4 of 4 files at r3, 1 of 1 files at r4, 2 of 2 files at r5, 1 of 1 files at r9, 1 of 1 files at r10, 2 of 3 files at r11, 1 of 1 files at r16, 1 of 1 files at r17, 3 of 3 files at r19.
> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/759)* status: all files reviewed, 11 unresolved discussions (waiting on @itrush and @rgomezp)

___
*[src/OneSignal.ts, line 767 at r19](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/759#-MV0-IqK2ddGCqTOsBR_:-MV0-IqK2ddGCqTOsBRa:bmg9jlj) ([raw file](https://github.com/onesignal/onesignal-website-sdk/blob/6e23cbd70ecdac5058387a861ea26e57a2be6c8c/src/OneSignal.ts#L767)):*
> ```XML
>   }
> 
>   // TO DO: consider renaming to something like privateGetNotOptedOutStatus
> ```

Opening a comment thread so we don't forget about this :smile: 
___
*[src/managers/PromptsManager.ts, line 94 at r19](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/759#-MV04p3G8ssXxXTgVli9:-MV04p3G8ssXxXTgVliA:b-r6f0tp) ([raw file](https://github.com/onesignal/onesignal-website-sdk/blob/6e23cbd70ecdac5058387a861ea26e57a2be6c8c/src/managers/PromptsManager.ts#L94)):*
> ```XML
>     const prompts = userPromptOptions.slidedown?.prompts;
>     if (!!prompts && prompts?.length > 0) {
>       for (let i=0; i<prompts.length; i++) {
> ```

nit, needs spaces around `=` and `<`
___
*[src/managers/PromptsManager.ts, line 238 at r19](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/759#-MV057UECjiix3fgZbTq:-MV057UECjiix3fgZbTr:bxbkzcg) ([raw file](https://github.com/onesignal/onesignal-website-sdk/blob/6e23cbd70ecdac5058387a861ea26e57a2be6c8c/src/managers/PromptsManager.ts#L238)):*
> ```XML
>       if (slidedown) {
>         slidedown.close();
>         // called here for compatibility with unit tests (close function doesn't run fully in test env)
> ```

I think this should be a TODO to look into before requiring it in the source.
___
*[src/managers/slidedownManager/SlidedownManager.ts, line 154 at r19](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/759#-MV05OR19hHlzi-Tbv2y:-MV05OR19hHlzi-Tbv2z:bvrsqh2) ([raw file](https://github.com/onesignal/onesignal-website-sdk/blob/6e23cbd70ecdac5058387a861ea26e57a2be6c8c/src/managers/slidedownManager/SlidedownManager.ts#L154)):*
> ```XML
> 
>     private async mountChannelCaptureContainer(options: AutoPromptOptions): Promise<void> {
>         // to do
> ```

Making note of todo so we don't forget before merging 
___
*[src/managers/slidedownManager/SlidedownManager.ts, line 156 at r19](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/759#-MV05UYiD6uTYCS184Xp:-MV05UYiD6uTYCS184Xq:b-b3ssyy) ([raw file](https://github.com/onesignal/onesignal-website-sdk/blob/6e23cbd70ecdac5058387a861ea26e57a2be6c8c/src/managers/slidedownManager/SlidedownManager.ts#L156)):*
> ```XML
>         // to do
>     }
> }
> ```

nit, missing new line
___
*[src/slidedown/Slidedown.ts, line 137 at r19](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/759#-MV05vXnC4pCefYR_7ux:-MV05vXnC4pCefYR_7uy:b-4h55qo) ([raw file](https://github.com/onesignal/onesignal-website-sdk/blob/6e23cbd70ecdac5058387a861ea26e57a2be6c8c/src/slidedown/Slidedown.ts#L137)):*
> ```XML
> 
>           /**
>            * Remember to trigger closed event after invoking close()
> ```

Not sure I understand where close needs to be called from this comment. i see `Event.trigger(Slidedown.EVENTS.CLOSED);` was called before, it should be in the comment that this needs to be called.
___
*[test/unit/prompts/SlidedownPrompting/_helpers.ts, line 1 at r5](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/759#-MV01xreDRgmoO806T0b:-MV01xreDRgmoO806T0c:bn3e2bp) ([raw file](https://github.com/onesignal/onesignal-website-sdk/blob/74160c0ece7069886735259c985ce060cbbe69d6/test/unit/prompts/SlidedownPrompting/_helpers.ts#L1)):*
> ```XML
> import { ExecutionContext } from "ava";
> ```

You noted that you had to start the class with a `_` underscore to prevent ava from trying to run it as a test because it is under the `test/unit` folder. I looked for guidance from ava here and found they recommend a `/helpers/` folder. See there release notes:
https://github.com/avajs/ava/releases/tag/v2.0.0
___
*[test/unit/prompts/SlidedownPrompting/_helpers.ts, line 102 at r5](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/759#-MV03XgO1llcI-XlKcwF:-MV03XgO1llcI-XlKcwG:b-5amvmk) ([raw file](https://github.com/onesignal/onesignal-website-sdk/blob/74160c0ece7069886735259c985ce060cbbe69d6/test/unit/prompts/SlidedownPrompting/_helpers.ts#L102)):*
> ```XML
> };
> 
> export const smsAndEmailOptions: SlidedownPromptOptions = {
> ```

Making a note that this isn't used right now. We will want to remove or use it before merging this PR.
___
*[test/unit/prompts/SlidedownPrompting/SlidedownPrompting.ts, line 89 at r5](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/759#-MV045EECMW4Xqr58_f1:-MV045EF1dIl20r_ZcQ1:bijkavx) ([raw file](https://github.com/onesignal/onesignal-website-sdk/blob/74160c0ece7069886735259c985ce060cbbe69d6/test/unit/prompts/SlidedownPrompting/SlidedownPrompting.ts#L89)):*
> ```XML
> 
> /* -- TO DO: FIX RACE CONDITION --
> test.serial(`on slidedown dismiss with slidedown queue non-empty, show next slidedown`, async t => {
> ```

This is a TODO for me to try to get working
___
*[test/unit/prompts/SlidedownPrompting/SlidedownPrompting.ts, line 94 at r5](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/759#-MV04Eij9Dgv19CUSbsz:-MV04Eij9Dgv19CUSbt-:bfzrde1) ([raw file](https://github.com/onesignal/onesignal-website-sdk/blob/74160c0ece7069886735259c985ce060cbbe69d6/test/unit/prompts/SlidedownPrompting/SlidedownPrompting.ts#L94)):*
> ```XML
>     const eventsHelper = new EventsTestHelper(sinonSandbox);
> 
>     // spies
> ```

Comment seems unneeded, the variable names cover this.
___
*[test/unit/public-sdk-apis/showHttpPrompt.ts, line 14 at r3](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/759#-MV00L-e5FymruJOpHYD:-MV00L-e5FymruJOpHYE:bmg9jlj) ([raw file](https://github.com/onesignal/onesignal-website-sdk/blob/4c166da6813e355290f006408d2c0a2946b6ceea/test/unit/public-sdk-apis/showHttpPrompt.ts#L14)):*
> ```XML
> });
> 
> test.todo("fix test");
> ```

Opening a comment thread so we don't forget about this :smile:

rgomezp (commented):
> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/759)* status: all files reviewed, 9 unresolved discussions (waiting on @itrush, @jkasten2, and @rgomezp)

___
*[src/OneSignal.ts, line 767 at r19](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/759#-MV0-IqK2ddGCqTOsBR_:-MV2RysQ9ZmZfsuuErgl:b-15brtj) ([raw file](https://github.com/onesignal/onesignal-website-sdk/blob/6e23cbd70ecdac5058387a861ea26e57a2be6c8c/src/OneSignal.ts#L767)):*

Opening a comment thread so we don't forget about this :smile: 
</blockquote></details>

`privateGetNotOptedOutStatus` is more descriptive of the return boolean. However, this is used in the public api `getSubscription` function which is similarly misleading. Should we also update the public api function? 

Also, probably better to make a separate PR for this change
___
*[src/managers/PromptsManager.ts, line 94 at r19](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/759#-MV04p3G8ssXxXTgVli9:-MV2Sfs-07o7keiq7bM0:bb7f2mf) ([raw file](https://github.com/onesignal/onesignal-website-sdk/blob/6e23cbd70ecdac5058387a861ea26e57a2be6c8c/src/managers/PromptsManager.ts#L94)):*

nit, needs spaces around `=` and `<`
</blockquote></details>

done
___
*[src/managers/PromptsManager.ts, line 238 at r19](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/759#-MV057UECjiix3fgZbTq:-MV2SpWj4SPggNHm7M8t:b-chb6og) ([raw file](https://github.com/onesignal/onesignal-website-sdk/blob/6e23cbd70ecdac5058387a861ea26e57a2be6c8c/src/managers/PromptsManager.ts#L238)):*

I think this should be a TODO to look into before requiring it in the source.
</blockquote></details>

I think we'l know better once we try running some unit tests. I'll add a comment though
___
*[src/managers/slidedownManager/SlidedownManager.ts, line 154 at r19](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/759#-MV05OR19hHlzi-Tbv2y:-MV2T6wgCSbvHP6-H_Mr:b-55xdaq) ([raw file](https://github.com/onesignal/onesignal-website-sdk/blob/6e23cbd70ecdac5058387a861ea26e57a2be6c8c/src/managers/slidedownManager/SlidedownManager.ts#L154)):*

Making note of todo so we don't forget before merging 
</blockquote></details>

this should be covered in the next PR. this PR is only covering prompting logic changes, not visual changes to the actual slidedown, which this would be an example of
___
*[src/slidedown/Slidedown.ts, line 137 at r19](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/759#-MV05vXnC4pCefYR_7ux:-MV2TesX4Z72i-0Y0nm7:b1o3sen) ([raw file](https://github.com/onesignal/onesignal-website-sdk/blob/6e23cbd70ecdac5058387a861ea26e57a2be6c8c/src/slidedown/Slidedown.ts#L137)):*

Not sure I understand where close needs to be called from this comment. i see `Event.trigger(Slidedown.EVENTS.CLOSED);` was called before, it should be in the comment that this needs to be called.
</blockquote></details>

the trigger was previously where this comment is. however, in testing the `once` function would not run correctly so the emitter would not fire correctly. this leads to unit tests, say where we have a promise waiting on two separate close events (for two separate slidedowns) would never resolve. 
___
*[test/unit/prompts/SlidedownPrompting/_helpers.ts, line 1 at r5](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/759#-MV01xreDRgmoO806T0b:-MV2VDdw4tBwohDuks76:b-yt3hng) ([raw file](https://github.com/onesignal/onesignal-website-sdk/blob/74160c0ece7069886735259c985ce060cbbe69d6/test/unit/prompts/SlidedownPrompting/_helpers.ts#L1)):*

You noted that you had to start the class with a `_` underscore to prevent ava from trying to run it as a test because it is under the `test/unit` folder. I looked for guidance from ava here and found they recommend a `/helpers/` folder. See there release notes:
https://github.com/avajs/ava/releases/tag/v2.0.0
</blockquote></details>

I'm open to doing it that way as well if you like. But that wouldn't be very different to how we have it today (i.e. a higher scoped `support` directory with any and every piece of helper code thrown together in a seemingly unspecific subdirectory name like `tester` or `sdk`).

If the goal of this change is to bring helper code closer to the code using it, your proposal would lead to a bunch of `helpers` directories in lower levels.

Furthermore, we already have a `helpers` directory in the unit tests directory. This directory contains actual unit tests for actual helper classes such as `InitHelper` or `ConfigHelper`. So we would want to change that directory name as well.

Here are our options:
1. Leverage adding _ to beginning of helper files so that they live at the same level as other files using them.
2. Leverage the `helpers` specification in the `ava` configuration so that ava ignores helper files. This would also require a change to an existing `helpers` directory we wouldn't want ava ignoring AND would result in a bunch of `helpers` directories with single or a just a couple files at lower levels (assuming the goal is in fact to bring helper code closer to files using it)
___
*[test/unit/prompts/SlidedownPrompting/_helpers.ts, line 102 at r5](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/759#-MV03XgO1llcI-XlKcwF:-MV2XZT93qPia2X6zIjs:bxtidt7) ([raw file](https://github.com/onesignal/onesignal-website-sdk/blob/74160c0ece7069886735259c985ce060cbbe69d6/test/unit/prompts/SlidedownPrompting/_helpers.ts#L102)):*

Making a note that this isn't used right now. We will want to remove or use it before merging this PR.
</blockquote></details>

I commented it out since we may want to use this in the next PR once we have sms and email slidedowns to test
___
*[test/unit/prompts/SlidedownPrompting/SlidedownPrompting.ts, line 94 at r5](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/759#-MV04Eij9Dgv19CUSbsz:-MV2XnRUBAmYr0ZGduVG:b-896fix) ([raw file](https://github.com/onesignal/onesignal-website-sdk/blob/74160c0ece7069886735259c985ce060cbbe69d6/test/unit/prompts/SlidedownPrompting/SlidedownPrompting.ts#L94)):*

Comment seems unneeded, the variable names cover this.
</blockquote></details>

Done.

jkasten2 (changes requested):
> Reviewed 1 of 1 files at r20, 1 of 1 files at r22, 1 of 1 files at r23, 1 of 1 files at r24.
> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/759)* status: all files reviewed, 16 unresolved discussions (waiting on @itrush, @jkasten2, and @rgomezp)

___
*[src/OneSignal.ts, line 767 at r19](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/759#-MV0-IqK2ddGCqTOsBR_:-MVVi2lcCunh0Wcq4m5-:bjdgab7) ([raw file](https://github.com/onesignal/onesignal-website-sdk/blob/6e23cbd70ecdac5058387a861ea26e57a2be6c8c/src/OneSignal.ts#L767)):*
<details><summary><i>Previously, rgomezp (Rodrigo Gomez Palacio) wrote…</i></summary><blockquote>

`privateGetNotOptedOutStatus` is more descriptive of the return boolean. However, this is used in the public api `getSubscription` function which is similarly misleading. Should we also update the public api function? 

Also, probably better to make a separate PR for this change
</blockquote></details>

I think `isPushDisabled` is a better name than `GetNotOptedOutStatus`. This matches closer to what the mobile SDK naming.

We can name internally if it makes sense, I won't change any public names in this PR.
___
*[test/unit/prompts/SlidedownPrompting/_helpers.ts, line 1 at r5](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/759#-MV01xreDRgmoO806T0b:-MVVnBlvAk8HGBlCvZl0:b-xr9du4) ([raw file](https://github.com/onesignal/onesignal-website-sdk/blob/74160c0ece7069886735259c985ce060cbbe69d6/test/unit/prompts/SlidedownPrompting/_helpers.ts#L1)):*
<details><summary><i>Previously, rgomezp (Rodrigo Gomez Palacio) wrote…</i></summary><blockquote>

I'm open to doing it that way as well if you like. But that wouldn't be very different to how we have it today (i.e. a higher scoped `support` directory with any and every piece of helper code thrown together in a seemingly unspecific subdirectory name like `tester` or `sdk`).

If the goal of this change is to bring helper code closer to the code using it, your proposal would lead to a bunch of `helpers` directories in lower levels.

Furthermore, we already have a `helpers` directory in the unit tests directory. This directory contains actual unit tests for actual helper classes such as `InitHelper` or `ConfigHelper`. So we would want to change that directory name as well.

Here are our options:
1. Leverage adding _ to beginning of helper files so that they live at the same level as other files using them.
2. Leverage the `helpers` specification in the `ava` configuration so that ava ignores helper files. This would also require a change to an existing `helpers` directory we wouldn't want ava ignoring AND would result in a bunch of `helpers` directories with single or a just a couple files at lower levels (assuming the goal is in fact to bring helper code closer to files using it)
</blockquote></details>

ah right, forgot about all the "helpers" in the src directory such as `ConfigHelper` and `InitHelper`. Looks like we put most, if not all our test helpers in `test/support/tester` and the huge `test/support/sdk/TestEnvironment.ts` today. 

1. What do you think about naming this `_SidedownPromptTestHelpers.ts`? We have a bunch of files just named`_helper.ts` it's hard to read where it is being imported from and when file searching.
2. This could be something like `testers` or `testhelper` instead of `helpers` because we do any use `helpers` for code in src.

I think option 1 with a more specific name would be the best approach for now. I don't think we going to nail down a longer term standard in a fast enough before we finish this PR.

We should definitely keep adding test helpers in general though to make reusable and reable code!
___
*[test/unit/prompts/SlidedownPrompting/_helpers.ts, line 102 at r5](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/759#-MV03XgO1llcI-XlKcwF:-MVVu3cl7nK64PpkNI1k:bjzle3v) ([raw file](https://github.com/onesignal/onesignal-website-sdk/blob/74160c0ece7069886735259c985ce060cbbe69d6/test/unit/prompts/SlidedownPrompting/_helpers.ts#L102)):*
<details><summary><i>Previously, rgomezp (Rodrigo Gomez Palacio) wrote…</i></summary><blockquote>

I commented it out since we may want to use this in the next PR once we have sms and email slidedowns to test
</blockquote></details>

ok, we can keep this thread open until we determine that and this code is either moved to that PR or deleted a unneeded.
___
*[test/unit/prompts/SlidedownPrompting/_helpers.ts, line 20 at r24](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/759#-MVVwzAk3bznvDfR7lTf:-MVVwzAk3bznvDfR7lTg:b-kbfhi4) ([raw file](https://github.com/onesignal/onesignal-website-sdk/blob/8ebb4de86304471bba735b64979ad4e98ac97ac7/test/unit/prompts/SlidedownPrompting/_helpers.ts#L20)):*
> ```XML
>     testConfig: TestEnvironmentConfig,
>     t: ExecutionContext)
>     {
> ```

nit, this `{` should go on the line above it.
___
*[test/unit/prompts/SlidedownPrompting/_helpers.ts, line 25 at r24](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/759#-MVVvx3i6ij9WW54SZ-r:-MVVvx3i6ij9WW54SZ-s:bx7r9nq) ([raw file](https://github.com/onesignal/onesignal-website-sdk/blob/8ebb4de86304471bba735b64979ad4e98ac97ac7/test/unit/prompts/SlidedownPrompting/_helpers.ts#L25)):*
> ```XML
> }
> 
> export function getShownPromiseWithEventCounts(eventCounts: EventCounts, resolveAfter: number = 0): Promise<void> {
> ```

This seems like a good candidate to move to the existing `EventsTestHelper` class
___
*[test/unit/prompts/SlidedownPrompting/_helpers.ts, line 28 at r24](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/759#-MVVxSv98RkY1iorcmJ9:-MVVxSv98RkY1iorcmJA:b-z367k6) ([raw file](https://github.com/onesignal/onesignal-website-sdk/blob/8ebb4de86304471bba735b64979ad4e98ac97ac7/test/unit/prompts/SlidedownPrompting/_helpers.ts#L28)):*
> ```XML
>     return new Promise<void>(resolve => {
>         OneSignal.on(Slidedown.EVENTS.SHOWN, () => {
>             eventCounts.shown+=1;
> ```

nit, space before and after `+=`
___
*[test/unit/prompts/SlidedownPrompting/_helpers.ts, line 34 at r24](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/759#-MVVw3hQ4d3L_Wry28Ln:-MVVw3hQ4d3L_Wry28Lo:bx7r9nq) ([raw file](https://github.com/onesignal/onesignal-website-sdk/blob/8ebb4de86304471bba735b64979ad4e98ac97ac7/test/unit/prompts/SlidedownPrompting/_helpers.ts#L34)):*
> ```XML
> }
> 
> export function getClosedPromiseWithEventCounts(eventCounts: EventCounts, resolveAfter: number = 0): Promise<void> {
> ```

This seems like a good candidate to move to the existing `EventsTestHelper` class
___
*[test/unit/prompts/SlidedownPrompting/_helpers.ts, line 37 at r24](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/759#-MVVxOjX5C0RmI5PRpcq:-MVVxOjY-aM7vKmu0MDR:b-z367k6) ([raw file](https://github.com/onesignal/onesignal-website-sdk/blob/8ebb4de86304471bba735b64979ad4e98ac97ac7/test/unit/prompts/SlidedownPrompting/_helpers.ts#L37)):*
> ```XML
>     return new Promise<void>(resolve => {
>         OneSignal.on(Slidedown.EVENTS.CLOSED, () => {
>             eventCounts.closed+=1;
> ```

nit, space before and after `+=`
___
*[test/unit/prompts/SlidedownPrompting/_helpers.ts, line 43 at r24](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/759#-MVVw49vDiCQqtKUSmDd:-MVVw49vDiCQqtKUSmDe:bx7r9nq) ([raw file](https://github.com/onesignal/onesignal-website-sdk/blob/8ebb4de86304471bba735b64979ad4e98ac97ac7/test/unit/prompts/SlidedownPrompting/_helpers.ts#L43)):*
> ```XML
> }
> 
> export function getSubscriptionPromise(): Promise<void> {
> ```

This seems like a good candidate to move to the existing `EventsTestHelper` class
___
*[test/unit/prompts/SlidedownPrompting/_helpers.ts, line 61 at r24](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/759#-MVVtAmD8dXIKJjDsJ2a:-MVVtAmD8dXIKJjDsJ2b:b20bivq) ([raw file](https://github.com/onesignal/onesignal-website-sdk/blob/8ebb4de86304471bba735b64979ad4e98ac97ac7/test/unit/prompts/SlidedownPrompting/_helpers.ts#L61)):*
> ```XML
> export function getPromptOptions(prompts: SlidedownPromptOptions[]): AppUserConfigPromptOptions {
>     return {
>         slidedown: {
> ```

This helper so simplistic that I think the abstraction it provides actually makes the code where it is used harder to read.
___
*[test/unit/prompts/SlidedownPrompting/_helpers.ts, line 76 at r24](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/759#-MVVtWBL8WqXE_Lm2Jnt:-MVVtWBL8WqXE_Lm2Jnu:b-ndw0dw) ([raw file](https://github.com/onesignal/onesignal-website-sdk/blob/8ebb4de86304471bba735b64979ad4e98ac97ac7/test/unit/prompts/SlidedownPrompting/_helpers.ts#L76)):*
> ```XML
> }
> 
> export const pushSlidedownOptions: SlidedownPromptOptions = {
> ```

It's clear to me this is defining `pushSlidedownOptions` but would like something in the name about the values. Such as default or minimal?
___
*[test/unit/prompts/SlidedownPrompting/_helpers.ts, line 86 at r24](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/759#-MVVtwwm9Hh5VHZxCxv0:-MVVtwwm9Hh5VHZxCxv1:b-nzxy4w) ([raw file](https://github.com/onesignal/onesignal-website-sdk/blob/8ebb4de86304471bba735b64979ad4e98ac97ac7/test/unit/prompts/SlidedownPrompting/_helpers.ts#L86)):*
> ```XML
> };
> 
> export const categorySlidedownOptions: SlidedownPromptOptions = {
> ```

It's clear to me this is defining `categorySlidedownOptions` but would like something in the name about the values. Such as default or minimal?
___
*[test/unit/prompts/SlidedownPrompting/SlidedownPrompting.ts, line 20 at r24](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/759#-MVVwM9OBiTnRdQ6YXXQ:-MVVwM9OBiTnRdQ6YXXR:b-v08vez) ([raw file](https://github.com/onesignal/onesignal-website-sdk/blob/8ebb4de86304471bba735b64979ad4e98ac97ac7/test/unit/prompts/SlidedownPrompting/SlidedownPrompting.ts#L20)):*
> ```XML
>     getShownPromiseWithEventCounts,
>     addPromptDelays,
>     setupWithStubs
> ```

This doesn't seem ideal to have to maintain and read this list. First thing that comes to mind is to wrap all these in a `SlidedownPromptTestHelper` class in `_helper.ts` and just import that.

jkasten2 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/759)* status: all files reviewed, 17 unresolved discussions (waiting on @itrush, @jkasten2, and @rgomezp)

___
*[src/managers/slidedownManager/SlidedownManager.ts, line 106 at r24](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/759#-MVW7je2CsC3fzyqZktG:-MVW7je2CsC3fzyqZktH:b5dfb6g) ([raw file](https://github.com/onesignal/onesignal-website-sdk/blob/8ebb4de86304471bba735b64979ad4e98ac97ac7/src/managers/slidedownManager/SlidedownManager.ts#L106)):*
> ```XML
>         await OneSignal.slidedown.create(options.isInUpdateMode);
>         await this.mountSpecialContainers(options);
>         this.setIsSlidedownShowing(true);
> ```

I think this line should be moved to just after the`if (this.isSlidedownShowing) {` if above. This way show can't be `createSlidedown` can't be called too quickly that would result in more than one showing.

This fixes the  hanging `Promise` issue with in the "TO DO: FIX RACE CONDITION --", however more changes are needed to the test to get them to pass.
___
*[test/unit/prompts/SlidedownPrompting/SlidedownPrompting.ts, line 89 at r5](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/759#-MV045EECMW4Xqr58_f1:-MVW92t929o6fKmwhk0J:b-p1jkq2) ([raw file](https://github.com/onesignal/onesignal-website-sdk/blob/74160c0ece7069886735259c985ce060cbbe69d6/test/unit/prompts/SlidedownPrompting/SlidedownPrompting.ts#L89)):*

This is a TODO for me to try to get working
</blockquote></details>

See my `this.setIsSlidedownShowing(true);` comment in `SlidedownManager.ts` to fix the Promise handling issue. Asserts need to be adjusted after

jkasten2 (commented):
For the logic on queuing you might want to take a look at the IAM code that was written for Android for inspiration on how to handle "is displaying" logic.
https://github.com/OneSignal/OneSignal-Android-SDK/blob/d786b30c57737c4a59f94d185f8708c39c9b4cec/OneSignalSDK/onesignal/src/main/java/com/onesignal/OSInAppMessageController.java#L619

> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/759#-:-MVW9WHY1lUxR33h5zZ7:b-v7gfp7)* status: all files reviewed, 17 unresolved discussions (waiting on @itrush, @jkasten2, and @rgomezp)

jkasten2 (commented):
https://github.com/OneSignal/OneSignal-Android-SDK/blob/d786b30c57737c4a59f94d185f8708c39c9b4cec/OneSignalSDK/onesignal/src/main/java/com/onesignal/OSInAppMessageController.java#L746

> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/759#-:-MVW9zVB3A8cjVmGc3WU:b-8txkqb)* status: all files reviewed, 17 unresolved discussions (waiting on @itrush, @jkasten2, and @rgomezp)

jkasten2 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/759)* status: all files reviewed, 16 unresolved discussions (waiting on @itrush, @jkasten2, and @rgomezp)

___
*[src/OneSignal.ts, line 767 at r19](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/759#-MV0-IqK2ddGCqTOsBR_:-MVYUvRX7xdqDQpOHXdg:bxl6ubb) ([raw file](https://github.com/onesignal/onesignal-website-sdk/blob/6e23cbd70ecdac5058387a861ea26e57a2be6c8c/src/OneSignal.ts#L767)):*

I think `isPushDisabled` is a better name than `GetNotOptedOutStatus`. This matches closer to what the mobile SDK naming.

We can name internally if it makes sense, I won't change any public names in this PR.
</blockquote></details>

Ok to use `privateGetOptedStatus` for this PR. Could do the rename in it's own commit to be clear.
We could rename this to`isPushDsiabled` in a different PR for the mobile name matching.
___
*[src/managers/slidedownManager/SlidedownManager.ts, line 106 at r24](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/759#-MVW7je2CsC3fzyqZktG:-MVYTbUn8vxppJCF5vsW:b-6ogzdf) ([raw file](https://github.com/onesignal/onesignal-website-sdk/blob/8ebb4de86304471bba735b64979ad4e98ac97ac7/src/managers/slidedownManager/SlidedownManager.ts#L106)):*

I think this line should be moved to just after the`if (this.isSlidedownShowing) {` if above. This way show can't be `createSlidedown` can't be called too quickly that would result in more than one showing.

This fixes the  hanging `Promise` issue with in the "TO DO: FIX RACE CONDITION --", however more changes are needed to the test to get them to pass.
</blockquote></details>

If there is some way we could be throwing here we should set `setIsSlidedownShowing` back to `false`.

rgomezp:
___
*[src/OneSignal.ts, line 767 at r19](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/759#-MV0-IqK2ddGCqTOsBR_:-MVcMiE07yNZEgHVRhak:b-2mxaeh) ([raw file](https://github.com/onesignal/onesignal-website-sdk/blob/6e23cbd70ecdac5058387a861ea26e57a2be6c8c/src/OneSignal.ts#L767)):*

Ok to use `privateGetOptedStatus` for this PR. Could do the rename in it's own commit to be clear.
We could rename this to`isPushDsiabled` in a different PR for the mobile name matching.

</blockquote></details>

I dug into this a bit more. From a rough and dirty glance (e.g: I looked at the IndexedDB value for "optedOut"). It seems that this value only gets set when the user is subscribed. It gets set to `false`. For this reason, technically this function is actually returning the correct value and the naming matches. However, I still think something should get done because we shouldn't be relying on "optedOut" as a sort of proxy for the subscription state. If this means we should add a new field "isSubscribed" to the local DB then we should consider that option.

rgomezp (commented):
> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/759)* status: 13 of 17 files reviewed, 13 unresolved discussions (waiting on @itrush, @jkasten2, and @rgomezp)

___
*[src/managers/slidedownManager/SlidedownManager.ts, line 106 at r24](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/759#-MVW7je2CsC3fzyqZktG:-MVcT7_WEJbF5aydTOWm:b-896fix) ([raw file](https://github.com/onesignal/onesignal-website-sdk/blob/8ebb4de86304471bba735b64979ad4e98ac97ac7/src/managers/slidedownManager/SlidedownManager.ts#L106)):*

If there is some way we could be throwing here we should set `setIsSlidedownShowing` back to `false`.
</blockquote></details>

Done.
___
*[test/unit/prompts/SlidedownPrompting/SlidedownPrompting.ts, line 20 at r24](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/759#-MVVwM9OBiTnRdQ6YXXQ:-MVcazth5w4rSdB56JbW:b-896fix) ([raw file](https://github.com/onesignal/onesignal-website-sdk/blob/8ebb4de86304471bba735b64979ad4e98ac97ac7/test/unit/prompts/SlidedownPrompting/SlidedownPrompting.ts#L20)):*

This doesn't seem ideal to have to maintain and read this list. First thing that comes to mind is to wrap all these in a `SlidedownPromptTestHelper` class in `_helper.ts` and just import that.
</blockquote></details>

Done.
___
*[test/unit/prompts/SlidedownPrompting/_helpers.ts, line 1 at r5](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/759#-MV01xreDRgmoO806T0b:-MVcUzigDm94hXr2sVxp:b-896fix) ([raw file](https://github.com/onesignal/onesignal-website-sdk/blob/74160c0ece7069886735259c985ce060cbbe69d6/test/unit/prompts/SlidedownPrompting/_helpers.ts#L1)):*

ah right, forgot about all the "helpers" in the src directory such as `ConfigHelper` and `InitHelper`. Looks like we put most, if not all our test helpers in `test/support/tester` and the huge `test/support/sdk/TestEnvironment.ts` today. 

1. What do you think about naming this `_SidedownPromptTestHelpers.ts`? We have a bunch of files just named`_helper.ts` it's hard to read where it is being imported from and when file searching.
2. This could be something like `testers` or `testhelper` instead of `helpers` because we do any use `helpers` for code in src.

I think option 1 with a more specific name would be the best approach for now. I don't think we going to nail down a longer term standard in a fast enough before we finish this PR.

We should definitely keep adding test helpers in general though to make reusable and reable code!
</blockquote></details>

Done.
___
*[test/unit/prompts/SlidedownPrompting/_helpers.ts, line 25 at r24](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/759#-MVVvx3i6ij9WW54SZ-r:-MVcYWwo4kKDvLfjaOge:b-896fix) ([raw file](https://github.com/onesignal/onesignal-website-sdk/blob/8ebb4de86304471bba735b64979ad4e98ac97ac7/test/unit/prompts/SlidedownPrompting/_helpers.ts#L25)):*

This seems like a good candidate to move to the existing `EventsTestHelper` class
</blockquote></details>

Done.
___
*[test/unit/prompts/SlidedownPrompting/_helpers.ts, line 34 at r24](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/759#-MVVw3hQ4d3L_Wry28Ln:-MVcYY0l0uSkjCwopDxs:b-896fix) ([raw file](https://github.com/onesignal/onesignal-website-sdk/blob/8ebb4de86304471bba735b64979ad4e98ac97ac7/test/unit/prompts/SlidedownPrompting/_helpers.ts#L34)):*

This seems like a good candidate to move to the existing `EventsTestHelper` class
</blockquote></details>

Done.
___
*[test/unit/prompts/SlidedownPrompting/_helpers.ts, line 43 at r24](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/759#-MVVw49vDiCQqtKUSmDd:-MVcYhL982XGKHUkhs08:b-896fix) ([raw file](https://github.com/onesignal/onesignal-website-sdk/blob/8ebb4de86304471bba735b64979ad4e98ac97ac7/test/unit/prompts/SlidedownPrompting/_helpers.ts#L43)):*

This seems like a good candidate to move to the existing `EventsTestHelper` class
</blockquote></details>

Done.
___
*[test/unit/prompts/SlidedownPrompting/_helpers.ts, line 61 at r24](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/759#-MVVtAmD8dXIKJjDsJ2a:-MVcYzBa0gdhnHItJ5hG:b-896fix) ([raw file](https://github.com/onesignal/onesignal-website-sdk/blob/8ebb4de86304471bba735b64979ad4e98ac97ac7/test/unit/prompts/SlidedownPrompting/_helpers.ts#L61)):*

This helper so simplistic that I think the abstraction it provides actually makes the code where it is used harder to read.
</blockquote></details>

Done.
___
*[test/unit/prompts/SlidedownPrompting/_helpers.ts, line 76 at r24](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/759#-MVVtWBL8WqXE_Lm2Jnt:-MVcZ4IT3pfO7WqnncbQ:b-896fix) ([raw file](https://github.com/onesignal/onesignal-website-sdk/blob/8ebb4de86304471bba735b64979ad4e98ac97ac7/test/unit/prompts/SlidedownPrompting/_helpers.ts#L76)):*

It's clear to me this is defining `pushSlidedownOptions` but would like something in the name about the values. Such as default or minimal?
</blockquote></details>

Done.
___
*[test/unit/prompts/SlidedownPrompting/_helpers.ts, line 86 at r24](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/759#-MVVtwwm9Hh5VHZxCxv0:-MVcZ9oZATF-w70U7YxW:b-896fix) ([raw file](https://github.com/onesignal/onesignal-website-sdk/blob/8ebb4de86304471bba735b64979ad4e98ac97ac7/test/unit/prompts/SlidedownPrompting/_helpers.ts#L86)):*

It's clear to me this is defining `categorySlidedownOptions` but would like something in the name about the values. Such as default or minimal?
</blockquote></details>

Done.

rgomezp:
___
*[test/unit/prompts/SlidedownPrompting/_helpers.ts, line 102 at r5](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/759#-MV03XgO1llcI-XlKcwF:-MVw_XdGDNL7ZKJXafoF:b-896fix) ([raw file](https://github.com/onesignal/onesignal-website-sdk/blob/74160c0ece7069886735259c985ce060cbbe69d6/test/unit/prompts/SlidedownPrompting/_helpers.ts#L102)):*

ok, we can keep this thread open until we determine that and this code is either moved to that PR or deleted a unneeded.
</blockquote></details>

Done.

rgomezp:
___
*[test/unit/public-sdk-apis/showHttpPrompt.ts, line 14 at r3](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/759#-MV00L-e5FymruJOpHYD:-MVw_hsG7TboZHK1XFO6:baeumx3) ([raw file](https://github.com/onesignal/onesignal-website-sdk/blob/4c166da6813e355290f006408d2c0a2946b6ceea/test/unit/public-sdk-apis/showHttpPrompt.ts#L14)):*

Opening a comment thread so we don't forget about this :smile: 
</blockquote></details>

I'm thinking it's better to merge this PR and make a separate one that just fixes the tests since it's already kind of a big PR. Thoughts?

rgomezp:
___
*[test/unit/prompts/SlidedownPrompting/SlidedownPrompting.ts, line 89 at r5](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/759#-MV045EECMW4Xqr58_f1:-MVwap1h3otT2XVuImY-:b-rje61i) ([raw file](https://github.com/onesignal/onesignal-website-sdk/blob/74160c0ece7069886735259c985ce060cbbe69d6/test/unit/prompts/SlidedownPrompting/SlidedownPrompting.ts#L89)):*

See my `this.setIsSlidedownShowing(true);` comment in `SlidedownManager.ts` to fix the Promise handling issue. Asserts need to be adjusted after
</blockquote></details>

See my below comment regarding putting related test fixes in dedicated PR

rgomezp:
Is there something specifically about this implementation that you would change regarding the displaying logic? @jkasten2

rgomezp (commented):
> Reviewed 1 of 1 files at r2, 4 of 4 files at r3, 1 of 1 files at r4, 1 of 1 files at r9, 1 of 1 files at r10, 2 of 3 files at r11, 1 of 1 files at r17, 1 of 3 files at r19, 1 of 1 files at r22, 1 of 1 files at r25, 2 of 3 files at r27, 1 of 1 files at r28.
> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/759)* status: all files reviewed, 13 unresolved discussions (waiting on @itrush and @jkasten2)

PR #767, opened 2021-03-16
fix(urls): tiny urls fixed for init errors

devpato:
# Description
## Details
 Based on the ticket Web SDK was being tested locally and user ran into an error, which points to a URL https://goo.gl/xip6JB that no longer exists. As you can see I modified 2 links instead of one because both of the links were down.

# Systems Affected
* [ ] WebSDK
* [ ] Backend
* [ ] Dashboard

# Validation
### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:

* [ ] No Typescript warnings

Other:

### Checklist

- - -
## 
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/767)

### DIFF
--- a/src/errors/SdkInitError.ts
+++ b/src/errors/SdkInitError.ts
@@ -24,10 +24,10 @@ export class SdkInitError extends OneSignalError {
         break;
       case SdkInitErrorKind.AppNotConfiguredForWebPush:
         errorMessage = `OneSignal: This app ID does not have any web platforms enabled. Double check your app` +
-        ` ID, or see step 1 on our setup guide (https://documentation.onesignal.com/docs/web-push-quickstart).`;
+        ` ID, or see step 1 on our setup guide (https://tinyurl.com/2x5jzk83).`;
         break;
       case SdkInitErrorKind.MissingSubdomain:
-        errorMessage = `Non-HTTPS pages require the subdomainName parameter within the label set within the OneSignal Web configuration (https://documentation.onesignal.com/docs/web-push-sdk#initialization).`;
+        errorMessage = `Non-HTTPS pages require the subdomainName parameter within the label set within the OneSignal Web configuration (https://tinyurl.com/ry39x7mk).`;
         break;
       case SdkInitErrorKind.WrongSiteUrl:
         if (extra && extra.siteUrl) {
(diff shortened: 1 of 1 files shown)


jkasten2 (changes requested):
> Reviewed 1 of 1 files at r1.
> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/767)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @devpato, @jkasten2, and @rgomezp)

___
*[src/errors/SdkInitError.ts, line 27 at r1](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/767#-MVwEMak8u6oP9-B9rtV:-MVwEMak8u6oP9-B9rtW:b-anqc13) ([raw file](https://github.com/onesignal/onesignal-website-sdk/blob/dd99de74a5576f69adde39682bbaa2c624b51b7f/src/errors/SdkInitError.ts#L27)):*
> ```XML
>       case SdkInitErrorKind.AppNotConfiguredForWebPush:
>         errorMessage = `OneSignal: This app ID does not have any web platforms enabled. Double check your app` +
>         ` ID, or see step 1 on our setup guide ( https://documentation.onesignal.com/docs/web-push-typical-setup).`;
> ```

nit, looks like an extra space was added after `(`.
___
*[src/errors/SdkInitError.ts, line 30 at r1](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/767#-MVwEqOIEn3oEAu-Axc_:-MVwEqOIEn3oEAu-Axca:bbpstvs) ([raw file](https://github.com/onesignal/onesignal-website-sdk/blob/dd99de74a5576f69adde39682bbaa2c624b51b7f/src/errors/SdkInitError.ts#L30)):*
> ```XML
>         break;
>       case SdkInitErrorKind.MissingSubdomain:
>         errorMessage = `Non-HTTPS pages require the subdomainName parameter within the label set within the OneSignal Web configuration (https://documentation.onesignal.com/docs/web-push-sdk#initialization).`;
> ```

Wow forgot about this check, I believe if you setup an HTTP site with either as "Custom Code" or "Typical Site" the code snip doesn't show `subdomainName`.

We can update this link for now in this PR to keep this one simple however.

jkasten2 (approved):
> Reviewed 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/767)* status: all files reviewed, 1 unresolved discussion (waiting on @devpato and @rgomezp)

jkasten2 (approved):
> Reviewed 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/767)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @rgomezp)

rgomezp (approved):

PR #797, opened 2021-04-15
Added SecondaryChannelManager to handle Email and future channel player updates

jkasten2:
# Description
## 1 Line Summary
Created a SecondaryChannelManager to make adding more channels (such as SMS) easier. This PR moves Email logic into this new manager.

## Details
This PR adds a new SecondaryChannelManager which all email handling has been moved into. This allows us to easily add more channels such as SMS. Introducing a Channel Manager which only handles secondary channels allowed us to keep a small scope and risk affecting push logic very low. The architecture SecondaryChannelManager uses could be a pattern that is expanded to push in the future however.

# Validation
## Tests
Tests added:
* setExternalUserId
* on_session, from page and from SW
* on_focus
* sendTags, with email
### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:

* [x] No Typescript warnings

Other:

### Checklist

- - -
## Related Tickets
- - -
## 
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/797)


src/helpers/page/AuthHashOptionsValidatorHelper.ts
```diff
@@ -0,0 +1,45 @@
+import { InvalidArgumentError, InvalidArgumentReason } from "../../errors/InvalidArgumentError";
+import Log from "../../libraries/Log";
+
+export class AuthHashOptionsValidatorHelper {
+
+  static VALID_AUTH_HASH_LENGTH = 64;
+
+  // Ensure that if the provided options object exists and has any keys that exist in the provided
+  //   array that they are a length of VALID_AUTH_HASH_LENGTH. null and undefined do not throw.
+  static throwIfInvalidAuthHashOptions(options: any, keys: Array<string>): string | undefined {
```
rgomezp, line 10: `keys: string[]`

rgomezp: can we set a type for `options`?

src/managers/channelManager/shared/SecondaryChannelProfileProvider.ts
```diff
@@ -0,0 +1,11 @@
+import { SecondaryChannelProfile } from "../../../models/SecondaryChannelProfile";
+
+// Interface to do the following operations on a SecondaryChannelProfile
+// * New based on parameters
+// * Get existing from storage
+// * Save to storage
+export interface SecondaryChannelProfileProvider {
+  newProfile(playerId?: string | null, identifier?: string, identifierAuthHash?: string): SecondaryChannelProfile;
```
rgomezp: why would this be `null` (e.g: vs `undefined` from `?`)

src/managers/channelManager/shared/SecondaryChannelEmail.ts
```diff
@@ -0,0 +1,111 @@
...
+      return false;
+    }
+
+    await Database.setEmailProfile(new EmailProfile());
+    return true;
+  }
+
+  async setIdentifier(identifier: string, authHash?: string): Promise<string | null> {
+    const profileProvider = this.secondaryChannelIdentifierUpdater.profileProvider;
+    const existingEmailProfile = await profileProvider.getProfile();
```
rgomezp, line 55: I'm a bit confused here. Is it possible that there is no email profile yet?

If yes, maybe change to `possibleExistingEmailProfile` or something similar

src/OneSignalApiShared.ts
```diff
@@ -54,44 +55,44 @@ export default class OneSignalApiShared {
...
+    const response = await OneSignalApiBase.post(`players`, secondaryChannelRecord.serialize());
     if (response && response.success) {
       return response.id;
     } else {
       return null;
     }
   }
 
-  static async updateEmailRecord(
+  static async updateSecondaryChannelRecord(
```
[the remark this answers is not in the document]
jkasten2: Done. While writing them I found out that `pushDeviceRecordId` isn't used at all. 57d5dc0c72e17f7936870a0da335a07f8c23bd39

test/support/lib/NockHelper.ts
```diff
@@ -0,0 +1,86 @@
...
+type NockInterceptMethods = "post" | "put" | "get";
+
+interface NockOptions {
+  reply: NockResponse;
+  baseUrl: string;
+  path: string;
+  method: NockInterceptMethods;
+}
+
+export class NockHelper {
```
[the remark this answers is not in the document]
jkasten2: Deciding that this can be a comment in the file instead of when and why you should use this NockHelper. More details on how tests should be written should be done later in another PR>

test/support/tester/NockOneSignalHelper.ts
```diff
@@ -0,0 +1,77 @@
...
+interface NockRequestResultPayerPost extends NockRequestResult {
+  request: NockRequest;
+  response: NockResponsePlayerPost;
+}
+
+export interface NockScopeWithResultPromisePlayerPost extends NockScopeWithResultPromise {
+  result: Promise<NockRequestResultPayerPost>;
+}
+
+export class NockOneSignalHelper {
```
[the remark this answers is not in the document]
jkasten2: Done

rgomezp:
Can we put the providers in its own `provider` directory?

test/support/lib/NockHelper.ts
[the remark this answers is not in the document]
jkasten2: Done, resolve if the comment I added is clear.

src/managers/channelManager/shared/SecondaryChannelEmail.ts
rgomezp: Yep that works

rgomezp (approved):

PR #801, opened 2021-04-17
Eliminate layout shift in bell hover

jmadler:
The bell has a responsive effect on hover when the user is subscribed.
This was implemented by adding or removing the "small" css class for the
bell.  The small/medium/large classes are also used to expose size
options to customers.

This change transitions to an approach where we add a scale transform
CSS property, which does not re-trigger layout.

### DIFF
--- a/src/bell/Launcher.ts
+++ b/src/bell/Launcher.ts
@@ -94,16 +94,25 @@ export default class Launcher extends ActiveAnimatedElement {
     this.wasInactive = false;
   }
 
+  scaleDown() {
+      return this.element.style.transform = 'scale(0.95)';
+  }
+
+
+  scaleBack() {
+      return this.element.style.transform = '';
+  }
+
   inactivate() {
     return this.bell.message.hide()
       .then(() => {
         if (this.bell.badge.content.length > 0) {
           return this.bell.badge.hide()
-            .then(() => Promise.all([super.inactivate(), this.resize('small')]))
+            .then(() => Promise.all([super.inactivate(), this.scaleDown()]))
             .then(() => this.bell.badge.show());
         }
         else {
-          return Promise.all([super.inactivate(), this.resize('small')]);
+          return Promise.all([super.inactivate(), this.scaleDown()]);
         }
       });
   }
@@ -111,10 +120,10 @@ export default class Launcher extends ActiveAnimatedElement {
   activate() {
     if (this.bell.badge.content.length > 0) {
       return this.bell.badge.hide()
-        .then(() => Promise.all([super.activate(), this.resize(this.bell.options.size)]));
+        .then(() => Promise.all([super.activate(), this.scaleBack()]));
     }
     else {
-      return Promise.all([super.activate(), this.resize(this.bell.options.size)]);
+      return Promise.all([super.activate(), this.scaleBack()]);
     }
   }
 }


jmadler:
(There's a TS error I'm unsure how to fix)

jkasten2 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/801)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @jmadler)

___
*[src/bell/Launcher.ts, line 98 at r1](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/801#-MYWUgdq91oNOMBGfcRz:-MYWUgdq91oNOMBGfcS-:bqt2yr5) ([raw file](https://github.com/onesignal/onesignal-website-sdk/blob/139d31cd177fc013bd5db7c9593c0bee0aa3cf33/src/bell/Launcher.ts#L98)):*
> ```XML
> 
>   scaleDown() {
>       return this.element.style.transform = 'scale(0.95)';
> ```

@jmadler This looks like the TS error you are talking about.
```
src/bell/Launcher.ts:98:27 - error TS2339: Property 'style' does not exist on type 'Element'.
98       return this.element.style.transform = 'scale(0.95)';
```

This where `this.element` is defined:
https://github.com/OneSignal/OneSignal-Website-SDK/blob/151401/src/bell/AnimatedElement.ts#L193

It seems like this definition is wrong. It returns `Element` but since we know it will be HTML we should indicate it as such. `document.querySelector` supports generics so we can define the type that way.
```typescript
return document.querySelector<HTMLElement>(this.selector);
```
___
*[src/bell/Launcher.ts, line 103 at r1](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/801#-MYWVodY2E5QVyqjPU9U:-MYWVodY2E5QVyqjPU9V:b-s6b4da) ([raw file](https://github.com/onesignal/onesignal-website-sdk/blob/139d31cd177fc013bd5db7c9593c0bee0aa3cf33/src/bell/Launcher.ts#L103)):*
> ```XML
> 
>   scaleBack() {
>       return this.element.style.transform = '';
> ```

`this.element` could be `nullish`, TS will probably error or warn on you about this on your new functions after you fix the `style` issue.

jmadler:
TS errors were removed, and bell is now style-identical to previous.
old
new

jkasten2 (changes requested):
> Reviewed 3 of 3 files at r2.
> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/801)* status: all files reviewed, 1 unresolved discussion (waiting on @jmadler)

___
*[src/bell/Bell.ts, line 393 at r2](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/801#-M_352cM8fAWmJuHcLhX:-M_352cM8fAWmJuHcLhY:b-ztxz) ([raw file](https://github.com/onesignal/onesignal-website-sdk/blob/5ad6952ab2958c95883c12f32086ee8015f4645a/src/bell/Bell.ts#L393)):*
> ```XML
>     const doNotPrompt = DismissHelper.wasPromptOfTypeDismissed(DismissPrompt.Push);
> 
>     await this.launcher.resize(this.options.size || this.DEFAULT_SIZE);
> ```

`isPushEnabled` check was removed. I believe we should keep this so it is always small when the user is already subscribed.

jmadler (commented):
> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/801)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @jkasten2)

___
*[src/bell/Bell.ts, line 393 at r2](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/801#-M_352cM8fAWmJuHcLhX:-Ma1kGYcC-RXDnp9RqKR:b-896fix) ([raw file](https://github.com/onesignal/onesignal-website-sdk/blob/5ad6952ab2958c95883c12f32086ee8015f4645a/src/bell/Bell.ts#L393)):*

`isPushEnabled` check was removed. I believe we should keep this so it is always small when the user is already subscribed.
</blockquote></details>

Done.

jkasten2 (approved):
> Reviewed 3 of 3 files at r3.
> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/801)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jmadler)

PR #808, opened 2021-04-28
Add SecondaryChannelManager to email & sms Slidedown to make WebPrompts Functional

jkasten2:
# Description
## 1 Line Summary
Added SecondaryChannelManager to SlidedownManager and connected it to email and sms updates so WebPrompts is now functional!

## Details
See commits for details.

# Validation
## Tests
Tested the following prompts
* Email
* SMS
* EmailAndSMS

Found pre existing bug where `updateSMS` is not given a phone number in a E.164 format. However fixed this in PR #809
### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [X] No Typescript warnings

Other:

### Checklist
       - No visual changes

---

## Related Tickets

---

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/808)

### DIFF
--- a/src/managers/slidedownManager/SlidedownManager.ts
+++ b/src/managers/slidedownManager/SlidedownManager.ts
@@ -129,12 +129,12 @@ export class SlidedownManager {
   private async handleAllowForSmsType(): Promise<void> {
     const smsInputFieldIsValid = OneSignal.slidedown.channelCaptureContainer.smsInputFieldIsValid;
     const isSmsEmpty = ChannelCaptureContainer.isSmsInputFieldEmpty();
-    const sms = ChannelCaptureContainer.getValueFromSmsInput();
 
-    if (!sms || !smsInputFieldIsValid || isSmsEmpty) {
+    if (!smsInputFieldIsValid || isSmsEmpty) {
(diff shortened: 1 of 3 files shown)


src/managers/slidedownManager/SlidedownManager.ts
```diff
@@ -119,19 +123,19 @@ export class SlidedownManager {
...
+    this.updateEmail(email);
   }
 
   private async handleAllowForSmsType(): Promise<void> {
     const smsInputFieldIsValid = OneSignal.slidedown.channelCaptureContainer.smsInputFieldIsValid;
     const isSmsEmpty = ChannelCaptureContainer.isSmsInputFieldEmpty();
+    const sms = ChannelCaptureContainer.getValueFromSmsInput();
 
-    if (!smsInputFieldIsValid || isSmsEmpty) {
+    if (!sms || !smsInputFieldIsValid || isSmsEmpty) {
```
rgomezp, line 134: this is redundant since we already check for `isSmsEmpty`

the function `isSmsInputFieldEmpty` will return the value or an empty string

rgomezp (commented):
small nit

PR #811, opened 2021-04-29
Feat: WebPrompts - fix reshow web prompts bug

rgomezp:
# Description
## 1 Line Summary
If the channel is already "subscribed", do not allow re-showing of the slidedown for now.

--- 

Since the dismiss key-value auto-erases based on the back-off logic, `autoPrompt: true` would result in the prompt re-showing even though the end-user may already have inputted their info.

Instead, we go based off of whether the non-push channel is already "subscribed", similar to how the push-dependent slidedown logic is done.

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/811)

### DIFF
--- a/src/managers/slidedownManager/SlidedownManager.ts
+++ b/src/managers/slidedownManager/SlidedownManager.ts
@@ -24,6 +24,9 @@ import ConfirmationToast from "../../slidedown/ConfirmationToast";
 import { awaitableTimeout } from "../../utils/AwaitableTimeout";
 import { DismissPrompt } from "../../models/Dismiss";
 import { SecondaryChannelManager } from "../channelManager/shared/SecondaryChannelManager";
+import Database from "../../services/Database";
+import AlreadySubscribedError from "../../errors/AlreadySubscribedError";
+import ExistingChannelError from "../../errors/ExistingChannelError";
 
 export class SlidedownManager {
   private context: ContextInterface;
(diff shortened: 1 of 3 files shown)


src/managers/slidedownManager/SlidedownManager.ts
```diff
@@ -83,6 +87,24 @@ export class SlidedownManager {
...
+
+      if (emailSubscribed && (slidedownType === DelayedPromptType.Email)) {
+        Log.info(new ExistingChannelError(DelayedPromptType.Email));
+        return false;
+      }
+
+      if (bothSubscribed && (slidedownType === DelayedPromptType.SmsAndEmail)) {
+        Log.info(new ExistingChannelError(DelayedPromptType.SmsAndEmail));
+        return false;
+      }
```
jkasten2, line 107: Notice we are checking for `isInUpdateMode` below all this new logic. Does calling `OneSignal.showSmsSlidedown({force: true});` still work to change the email or SMS?

jkasten2 (approved):
Had one concern, if it isn't an issue feel free to merge in

src/managers/slidedownManager/SlidedownManager.ts
jkasten2: Seems like this would break the {force: true} behavior.
I think it make sense to allow force true to always show, even if we don't have an visual different for differences / autofill yet. This way we don't change the behavior later when we introduce that

jkasten2 (changes requested):

jkasten2 (approved):

PR #812, opened 2021-04-29
Use persisted external user id for player create

jkasten2:
# Description
## 1 Line Summary
If `OneSignal.externalUserId` is called on a different page before the user subscribes (to push, email, or sms) ensure we set it on the subscription.

## Details
This PR fixes the following scenario:
1. OneSignal.setExternalUserId is called.
2. Page is reloaded
3. OneSignal.setEmail (or any player create happens).
4. external_user_id is never set on the player.
* In this commit we are now sending external_user_id on /player create
REST API to cover this scenario.

# Validation
### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [X] No Typescript warnings

Other:

### Checklist

---

## Related Tickets

---

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/812)

### DIFF
--- a/src/OneSignal.ts
+++ b/src/OneSignal.ts
@@ -534,13 +534,8 @@ export default class OneSignal {
   public static async removeExternalUserId(): Promise<void> {
     await awaitOneSignalInitAndSupported();
     logMethodCall("removeExternalUserId");
-    const isExistingUser = await this.context.subscriptionManager.isAlreadyRegisteredWithOneSignal();
-    if (!isExistingUser) {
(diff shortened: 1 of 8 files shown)


rgomezp (commented):
Is there a test that could be written to simulate this behavior?

jkasten2:
@rgomezp Added tests!
Also noticed an issue with `removeExternalUserId` logic that differed from setExternalUserId so I corrected that in it's own commit.

rgomezp (approved):
Small nit on renaming that one test to be more clear

PR #817, opened 2021-05-04
Fix/allow set external user id for secondary channels, without push

jkasten2:
# Description
## 1 Line Summary
Fixes limitation where setExternalUserId did not work for email or SMS unless the user also accepted push.

## Details
* Moved logic from OneSignal.privateSetExternalUserId to UpdateManager, this way it can encapsulate push and secondary channels. 
* Review commit-by-commit for more details.

# Validation
### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [X] No Typescript warnings

Other:

### Checklist
      - No visual changes

---

## Related Tickets

---

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/817)

### DIFF
--- a/src/managers/UpdateManager.ts
+++ b/src/managers/UpdateManager.ts
@@ -35,7 +35,7 @@ export class UpdateManager {
     return MainHelper.createDeviceRecord(this.context.appConfig.appId);
   }
 
-  public async sendPlayerUpdate(deviceRecord?: PushDeviceRecord): Promise<void> {
+  public async sendPushDeviceRecordUpdate(deviceRecord?: PushDeviceRecord): Promise<void> {
     const existingUser = await this.context.subscriptionManager.isAlreadyRegisteredWithOneSignal();
     if (!existingUser) {
       Log.debug("Not sending the update because user is not registered with OneSignal (no device id)");
--- a/src/managers/UpdateManager.ts
+++ b/src/managers/UpdateManager.ts
@@ -9,8 +9,8 @@ import { ContextSWInterface } from '../models/ContextSW';
 import Utils from "../context/shared/utils/Utils";
 import { SessionOrigin } from "../models/Session";
 import { OutcomeRequestData } from "../models/OutcomeRequestData";
-import { logMethodCall } from '../utils';
-import { UpdatePlayerExternalUserId } from "../models/UpdatePlayerOptions";
+import { awaitSdkEvent, logMethodCall } from '../utils';
+import { UpdatePlayerExternalUserId, UpdatePlayerOptions } from "../models/UpdatePlayerOptions";
 import { ExternalUserIdHelper } from "../helpers/shared/ExternalUserIdHelper";
 
 export class UpdateManager {
@@ -21,6 +21,10 @@ export class UpdateManager {
   constructor(context: ContextSWInterface) {
     this.context = context;
     this.onSessionSent = context.pageViewManager.getPageViewCount() > 1;
+  }
+
+  private async isDeviceIdAvailable(): Promise<boolean> {
+    return (await Database.getSubscription()).deviceId != null;
   }
 
   private async getDeviceId(): Promise<string> {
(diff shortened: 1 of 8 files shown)


test/unit/public-sdk-apis/setExternalUserIdWithSecondaryChannel.ts
```diff
@@ -87,3 +87,22 @@ test("setExternalUserId before email, includes external_user_id in POST create c
...
+  // 1. Nock out email create
+  const emailPostNock = NockOneSignalHelper.nockPlayerPost();
+  await OneSignal.setEmail(TEST_EMAIL_ADDRESS);
+  const emailPlayerId = (await emailPostNock.result).response.body.id;
+
+  // 2. Call OneSignal.setExternalUserId and ensure email is updated
+  const externalUserIdUpdateOnEmail = NockOneSignalHelper.nockPlayerPut(emailPlayerId);
+  await OneSignal.setExternalUserId(TEST_EXTERNAL_USER_ID);
+
+  t.deepEqual(
```
rgomezp, line 106: add a comment noting what is different in the body (why is this failing prior to the change?)

src/managers/UpdateManager.ts
```diff
@@ -23,6 +23,10 @@ export class UpdateManager {
     this.onSessionSent = context.pageViewManager.getPageViewCount() > 1;
   }
 
+  private async isDeviceIdAvailable(): Promise<boolean> {
+    return (await Database.getSubscription()).deviceId != null;
```
rgomezp, line 27: when is this set to `null`? do you mean `undefined`?

or are you purposefully using `!=`  ?

src/managers/UpdateManager.ts
```diff
@@ -139,6 +141,24 @@ export class UpdateManager {
...
+    const pushUpdatePromise = this.sendPushPlayerUpdate(payload);
+    if (await this.isDeviceIdAvailable()) {
+      await pushUpdatePromise;
+    }
+  }
+
+  // Send REST API payload to update the push player record.
+  // Await until we have a push playerId, which is require to make a PUT call.
+  private async sendPushPlayerUpdate(payload: UpdatePlayerOptions): Promise<void> {
+    let deviceId =  (await Database.getSubscription()).deviceId;
```
rgomezp, line 153: I believe you can destructure here

src/managers/UpdateManager.ts
```diff
@@ -139,6 +141,24 @@ export class UpdateManager {
...
+    if (!deviceId) {
+      await awaitSdkEvent(OneSignal.EVENTS.REGISTERED);
+      deviceId = (await Database.getSubscription()).deviceId;
+    }
+    if (!deviceId) {
```
rgomezp, line 158: I think this makes more sense nested in the above conditional

src/managers/UpdateManager.ts
jkasten2: yes, `deviceId` maybe `undefined` or `null` based on the return type we defined.

src/managers/UpdateManager.ts
jkasten2: We can, but was trying to keep it the same as line 156 below so it is easier to spot we are doing the same thing. Thoughts?

src/managers/UpdateManager.ts
jkasten2: We have to do a final undefined / null check before calling `OneSignalApiShared.updatePlayer` since it requires a valid instance. The compiler thinks line 156 could still return a nullish value.

I'll wrap the call `OneSignalApiShared.updatePlayer` in an if since the early return here doesn't make that clear.

test/unit/public-sdk-apis/setExternalUserIdWithSecondaryChannel.ts
jkasten2: 👍 I'll add a comment

jkasten2: Done

src/managers/UpdateManager.ts
jkasten2: It seems the deconstruct syntax does work with reassignment, but requires an extra set of parentheses.
```typescript
    let { deviceId } =  await Database.getSubscription();
    if (!deviceId) {
      await awaitSdkEvent(OneSignal.EVENTS.REGISTERED);
      ({ deviceId } = await Database.getSubscription());
    }
```

This looks a bit better so I'll update it to this.

jkasten2: Done

src/managers/UpdateManager.ts
jkasten2: Done

src/managers/UpdateManager.ts
rgomezp: Ah interesting. I didn't know that one.

rgomezp (approved):

PR #828, opened 2021-05-12
Customlink - CLS fix, refresh, and refactor

rgomezp:
# Description
## 1 Line Summary
Eliminate layout shift for Custom Link by adding a fixed height. Includes style updates.

## Details
Style updates: remove box-shadow and add opacity in subscribed state.

# Systems Affected
   - [x] WebSDK
   - [ ] Backend
   - [ ] Dashboard

# Validation
### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [x] No Typescript warnings

Other:

### Info
May-12-2021 16-05-41

### Checklist

---

## Related Tickets

---

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/828)


jkasten2 (changes requested):
This PR is difficult to review to review the changes from`CustomLink.ts` to `CustomLinkManager.ts` due to the following reasons:
1. `CustomLinkManager.ts` was added on one commit then in another commit `CustomLink.ts` was removed, github and reviewable don't give you a side-by-side of the changes. In order to review the changes one has to copy both locally and use a local diff tool.
2. The finial `CustomLinkManager.ts`was added in one commit. It would have been easier if changes were broken down into many commits. Example: A. Rename/move `CustomLink.ts` -> `CustomLinkManager.ts`. B. Refactored initialization. C. Refactored injectMarkup. D. Ect...

> Reviewed 3 of 3 files at r1, 1 of 1 files at r2, 1 of 1 files at r3, 1 of 1 files at r4, 1 of 1 files at r5, 5 of 5 files at r6, 1 of 1 files at r7.
> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/828#-:-M_cp9FG-TJN8PiWK-Xk:b-uos869)* status: all files reviewed, 7 unresolved discussions (waiting on @rgomezp)

___
*[src/managers/CustomLinkManager.ts, line 35 at r7](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/828#-M_clvJu72yaYwikK-_j:-M_clvJu72yaYwikK-_k:bisqeo6) ([raw file](https://github.com/onesignal/onesignal-website-sdk/blob/4e7344e89436855074ec6a21aaff74919fc6490c/src/managers/CustomLinkManager.ts#L35)):*
> ```XML
>     }
> 
>     for (let i=0; i<this.customlinkContainerElements.length; i++) {
> ```

This would be cleaner as a `forEach`.
___
*[src/managers/CustomLinkManager.ts, line 38 at r7](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/828#-M_coQe7C5CKsp6CyQsJ:-M_coQe7C5CKsp6CyQsK:b3dnwmx) ([raw file](https://github.com/onesignal/onesignal-website-sdk/blob/4e7344e89436855074ec6a21aaff74919fc6490c/src/managers/CustomLinkManager.ts#L38)):*
> ```XML
>       await this.injectMarkup(this.customlinkContainerElements[i]);
>     }
>   }
> ```

There used to be for `explanationSelector` but this is gone now in the new file (seems we only need `explanationClass` now?). I see we moved its definition into `CUSTOM_LINK_CSS_SELECTORS` but is not referenced anywhere now:
https://github.com/OneSignal/OneSignal-Website-SDK/pull/828/files#diff-f99abf27ad26dbfc8186c9957ee4b62242ed1fd3d97d33db4c54f6b326d17d8bR139
___
*[src/managers/CustomLinkManager.ts, line 81 at r7](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/828#-M_cwgeD1aS-rGdmTPG0:-M_cwgeD1aS-rGdmTPG1:b2pljhv) ([raw file](https://github.com/onesignal/onesignal-website-sdk/blob/4e7344e89436855074ec6a21aaff74919fc6490c/src/managers/CustomLinkManager.ts#L81)):*
> ```XML
> 
>     if (this.config.text.subscribe) {
>       const subscribe = document.createElement("button");
> ```

This method is long enough where `subscribe` isn't descriptive enough for the context. I would recommend renaming this to `subscribeButton`
___
*[src/managers/CustomLinkManager.ts, line 174 at r7](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/828#-M_cstVaAsgNDf3Fbw-5:-M_cstVaAsgNDf3Fbw-6:b-8cciay) ([raw file](https://github.com/onesignal/onesignal-website-sdk/blob/4e7344e89436855074ec6a21aaff74919fc6490c/src/managers/CustomLinkManager.ts#L174)):*
> ```XML
>   }
> 
>   private async updateSubscriptionState(): Promise<void> {
> ```

This sort of "refresh" link pattern to update class instance variables is normally bad practice as one could forget to call the refresh. Why not just make two private getter methods or properties for these? `isPushEnabled`and `isOptedOut`.
___
*[src/managers/CustomLinkManager.ts, line 179 at r7](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/828#-M_cuqVZEkxx5PCrO7ho:-M_cuqVZEkxx5PCrO7hp:b-573xjp) ([raw file](https://github.com/onesignal/onesignal-website-sdk/blob/4e7344e89436855074ec6a21aaff74919fc6490c/src/managers/CustomLinkManager.ts#L179)):*
> ```XML
>   }
> 
>   private setCustomColors(element: HTMLElement): void {
> ```

This method and logic looks pre-existing so a change isn't require but from a quick look I see a way to simplify the logic here.
Both if statements are checking for if `this.config?.color` and  `this.config?.color.text` are truthy, we could make this an early return and removed them out of both if statements.
___
*[src/stylesheets/customlink.scss, line 42 at r4](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/828#-M_city91Lousynf870c:-M_city91Lousynf870d:bblwvc6) ([raw file](https://github.com/onesignal/onesignal-website-sdk/blob/7501c8bb54338688484c98d6506b4a49cf3e3c1d/src/stylesheets/customlink.scss#L42)):*
> ```SCSS
>     font-weight: 600px;
>     border-radius: 4px;
>     height: 20px;
> ```

Can we make the height based on the font size somehow? Also is there anything else that would make this static value no longer correct? Maybe there isn't anything else we can do but this seems fragile.
___
*[test/support/tester/utils.ts, line 124 at r6](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/828#-M_cjWmf1HN3cYtImRwc:-M_cjWmf1HN3cYtImRwd:b-z34ip6) ([raw file](https://github.com/onesignal/onesignal-website-sdk/blob/679a79e70a908818b7bb8ef0ab266cbf5bb0bc9f/test/support/tester/utils.ts#L124)):*
> ```XML
> }
> 
> export function simulateEventOfTypeOnElement(type: string, element: Element|null): void {
> ```

nit, space before and after `|`.

rgomezp (commented):
> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/828)* status: all files reviewed, 6 unresolved discussions (waiting on @jkasten2 and @rgomezp)

___
*[src/managers/CustomLinkManager.ts, line 35 at r7](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/828#-M_clvJu72yaYwikK-_j:-M_gTYS7-Qet7Qiw6VI_:b-7ggzd3) ([raw file](https://github.com/onesignal/onesignal-website-sdk/blob/4e7344e89436855074ec6a21aaff74919fc6490c/src/managers/CustomLinkManager.ts#L35)):*

This would be cleaner as a `forEach`.
</blockquote></details>

This is a weird case where we actually want a traditional for-loop.

A forEach here causes race-conditions that break our unit tests. The reason is that asynchronous forEachs "kick off" each callback which run independently returning their individual promises. However, these promises get thrown away with a traditional forEach.

The closest thing would be awaiting on all the promises returned by the `map` function - which does return the individual promises.

That would look like:
```
await Promise.all(this.customlinkContainerElements.map(async elem => {
      await this.injectMarkup(elem);
    }));
```
___
*[src/managers/CustomLinkManager.ts, line 38 at r7](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/828#-M_coQe7C5CKsp6CyQsJ:-M_gWFbY50WSgMdZKZSp:bcw33zx) ([raw file](https://github.com/onesignal/onesignal-website-sdk/blob/4e7344e89436855074ec6a21aaff74919fc6490c/src/managers/CustomLinkManager.ts#L38)):*

There used to be for `explanationSelector` but this is gone now in the new file (seems we only need `explanationClass` now?). I see we moved its definition into `CUSTOM_LINK_CSS_SELECTORS` but is not referenced anywhere now:
https://github.com/OneSignal/OneSignal-Website-SDK/pull/828/files#diff-f99abf27ad26dbfc8186c9957ee4b62242ed1fd3d97d33db4c54f6b326d17d8bR139
</blockquote></details>

The explanation selector is used in the unit tests, still.

The reason it is no longer in the source is that instead of hiding both the explanation element and the subscription button element individually, it is much easier to hide them together by hiding the parent node.
___
*[src/managers/CustomLinkManager.ts, line 81 at r7](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/828#-M_cwgeD1aS-rGdmTPG0:-M_gWRzW-Lx_p0dHzflZ:b-896fix) ([raw file](https://github.com/onesignal/onesignal-website-sdk/blob/4e7344e89436855074ec6a21aaff74919fc6490c/src/managers/CustomLinkManager.ts#L81)):*

This method is long enough where `subscribe` isn't descriptive enough for the context. I would recommend renaming this to `subscribeButton`
</blockquote></details>

Done.
___
*[src/managers/CustomLinkManager.ts, line 174 at r7](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/828#-M_cstVaAsgNDf3Fbw-5:-M_gWkF65ONRznM911VE:b-896fix) ([raw file](https://github.com/onesignal/onesignal-website-sdk/blob/4e7344e89436855074ec6a21aaff74919fc6490c/src/managers/CustomLinkManager.ts#L174)):*

This sort of "refresh" link pattern to update class instance variables is normally bad practice as one could forget to call the refresh. Why not just make two private getter methods or properties for these? `isPushEnabled`and `isOptedOut`.
</blockquote></details>

Done.
___
*[src/managers/CustomLinkManager.ts, line 179 at r7](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/828#-M_cuqVZEkxx5PCrO7ho:-M_gXSAI3iAJ4zcWFtRc:b-896fix) ([raw file](https://github.com/onesignal/onesignal-website-sdk/blob/4e7344e89436855074ec6a21aaff74919fc6490c/src/managers/CustomLinkManager.ts#L179)):*

This method and logic looks pre-existing so a change isn't require but from a quick look I see a way to simplify the logic here.
Both if statements are checking for if `this.config?.color` and  `this.config?.color.text` are truthy, we could make this an early return and removed them out of both if statements.
</blockquote></details>

Done.
___
*[test/support/tester/utils.ts, line 124 at r6](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/828#-M_cjWmf1HN3cYtImRwc:-M_gYG2yE6ZixUaEt4Y3:b-896fix) ([raw file](https://github.com/onesignal/onesignal-website-sdk/blob/679a79e70a908818b7bb8ef0ab266cbf5bb0bc9f/test/support/tester/utils.ts#L124)):*

nit, space before and after `|`.
</blockquote></details>

Done.

rgomezp:
___
*[src/stylesheets/customlink.scss, line 42 at r4](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/828#-M_city91Lousynf870c:-M_gYe69CQwy7w6TOJWy:bez1ewb) ([raw file](https://github.com/onesignal/onesignal-website-sdk/blob/7501c8bb54338688484c98d6506b4a49cf3e3c1d/src/stylesheets/customlink.scss#L42)):*

Can we make the height based on the font size somehow? Also is there anything else that would make this static value no longer correct? Maybe there isn't anything else we can do but this seems fragile.
</blockquote></details>

I'll try writing up some mixins

jkasten2 (approved):
> Reviewed 5 of 5 files at r8.
> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/828)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @rgomezp)

PR #837, opened 2021-06-21
Add back clients.claim to fix AMP update issue

jkasten2:
# Description
## 1 Line Summary
Fixes AMP WebPush issue where the subscribe button does not update after accepting notifications from the pop-up window.

## Details
* AMP WebPush (amp-web-push v0.1) requires clients.claim()
*  It depends on ServiceWorker having full control of the iframe,
requirement could be lifted by AMP in the future however.
   - Without this the AMP symptom is the subscribe button does not
    update right after accepting the notification permission from
    the pop-up.

# Validation
### Checklist
      - Tested on a standard WebPush setup to ensure it doesn't cause any regression issues.

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [X] No Typescript warnings

Other:

---

## Related Tickets

---

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/837)

### DIFF
--- a/src/service-worker/ServiceWorker.ts
+++ b/src/service-worker/ServiceWorker.ts
@@ -91,6 +91,7 @@ export class ServiceWorker {
    * Service worker entry point.
    */
   static run() {
+    self.addEventListener('activate', ServiceWorker.onServiceWorkerActivated);
     self.addEventListener('push', ServiceWorker.onPushReceived);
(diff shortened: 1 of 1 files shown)


rgomezp (approved):
LGTM!

PR #844, opened 2021-07-06
Move the bell manager function until after knowing whether to show sl…

rgomezp:
…idedown

# Description
## 1 Line Summary
this commit moves the function that manages the bell while slidedown shows until later (after we know whether we *should* show the slidedown to begin with

## Details
This fixes an issue where we were still hiding the bell when the slidedown attempted and failed to show (due to being previously dismissed, for example)

# Systems Affected
   - [x] WebSDK
   - [ ] Backend
   - [ ] Dashboard

# Validation
### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [x] No Typescript warnings

Other:

## Related Tickets

---

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/844)

### DIFF
--- a/src/managers/PromptsManager.ts
+++ b/src/managers/PromptsManager.ts
@@ -3,7 +3,7 @@ import OneSignalUtils from "../utils/OneSignalUtils";
 import { ContextInterface } from "../models/Context";
 import MainHelper from '../helpers/MainHelper';
 import { ResourceLoadState } from '../services/DynamicResourceLoader';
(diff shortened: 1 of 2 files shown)


jkasten2 (approved):
> Reviewed 2 of 2 files at r1.
> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/844)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @rgomezp)

PR #860, opened 2021-11-02
AMP Subscribe - Was previously not saving  to DB

rgomezp:
# Description

In AMP integrations (service worker code only) we were previously not saving the `appId` to the IndexeDB. This was previously only being done in the `InitHelper` which is called from the main context.

The `getAppId` function in [`ServiceWorker`](https://github.com/OneSignal/OneSignal-Website-SDK/blob/main/src/service-worker/ServiceWorker.ts#L151) was written in a way that it first tries to get the query param before trying the Database. Weirdly, on that line the variable `appId` loses its reference within the block scope and suddenly becomes `undefined` after stepping through to the following line (`match` also gets lost).

After spending some time trying to resolve this bizarre behavior with no luck, I figured we can simply add the same line from the `InitHelper` so that no further code changes are needed.

After trying this for myself, it works fine and the clicks are being recorded successfully.

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/860)

### DIFF
--- a/src/service-worker/ServiceWorker.ts
+++ b/src/service-worker/ServiceWorker.ts
@@ -196,6 +196,7 @@ export class ServiceWorker {
       const context = new ContextSW(appConfig);
       const rawSubscription = await context.subscriptionManager.subscribe(SubscriptionStrategyKind.ResubscribeExisting);
       const subscription = await context.subscriptionManager.registerSubscription(rawSubscription);
+      await Database.put('Ids', { type: 'appId', id: appId });
       ServiceWorker.workerMessenger.broadcast(WorkerMessengerCommand.AmpSubscribe, subscription.deviceId);
     });
     ServiceWorker.workerMessenger.on(WorkerMessengerCommand.AmpUnsubscribe, async () => {
@@ -245,7 +246,7 @@ export class ServiceWorker {
               //Display push notifications in the order we received them
               const notificationEventPromiseFns = [];
               const notificationReceivedPromises: Promise<void>[] = [];
-              const appId = await Database.get<string>("Ids", "appId");
+              const appId = await ServiceWorker.getAppId();
 
               for (const rawNotification of notifications) {
                 Log.debug('Raw Notification from OneSignal:', rawNotification);
@@ -785,7 +786,7 @@ export class ServiceWorker {
 
     const launchUrl: string = await ServiceWorker.getNotificationUrlToOpen(notificationData);
     const notificationOpensLink: boolean = ServiceWorker.shouldOpenNotificationUrl(launchUrl);
-    const appId = await Database.get<string>("Ids", "appId");
+    const appId = await ServiceWorker.getAppId();
     const deviceType = DeviceRecord.prototype.getDeliveryPlatform();
 
     let saveNotificationClickedPromise: Promise<void> | undefined;


jkasten2 (changes requested):
> In AMP integrations (service worker code only) we were previously not saving the `appId` to the IndexeDB. This was previously only being done in the `InitHelper` which is called from the main context.

ah interesting, that make sense since I believe the `WorkerMessengerCommand.AmpSubscribe` event would be the first opportunity we have to save the id. Everything run before this is build into AMP push and isn't OneSignal specific code.

> The `getAppId` function in [`ServiceWorker`](https://github.com/OneSignal/OneSignal-Website-SDK/blob/main/src/service-worker/ServiceWorker.ts#L151) was written in a way that it first tries to get the query param before trying the Database. Weirdly, on that line the variable `appId` loses its reference within the block scope and suddenly becomes `undefined` after stepping through to the following line (`match` also gets lost).

Two thoughts on this part;
1. I don't see how `appId` or `match` could be `undefined` in the code you linked to, it would be out of scope. I have seen the debugger not always show the correct value, you may want to `console.log` instead to confirm this.
2. I don't see any changes addressing this. However changes might not be needed if we find out the true value with my above note.

> After spending some time trying to resolve this bizarre behavior with no luck, I figured we can simply add the same line from the `InitHelper` so that no further code changes are needed.

The concussion here is no issue?

> After trying this for myself, it works fine and the clicks are being recorded successfully.

Thanks manually testing to confirm! I think the two lines you are changed should account for the issue and possibility others since the database value will be around going forward for new subscribers.

> Reviewed 1 of 1 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/860#-:-MnU2awv-00zYff04yaJ:b-810a6r)* status: all files reviewed, 2 unresolved discussions (waiting on @rgomezp)

___
*[src/service-worker/ServiceWorker.ts, line 199 at r1](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/860#-MnU66EUDxdb1vmDPrlu:-MnU66EUDxdb1vmDPrlv:bhxh2ft) ([raw file](https://github.com/onesignal/onesignal-website-sdk/blob/bef74d6d9ef7e76818423fbb85ee48d730321f7b/src/service-worker/ServiceWorker.ts#L199)):*
> ```XML
>       const rawSubscription = await context.subscriptionManager.subscribe(SubscriptionStrategyKind.ResubscribeExisting);
>       const subscription = await context.subscriptionManager.registerSubscription(rawSubscription);
>       await Database.put('Ids', { type: 'appId', id: appId });
> ```


:+1: , this will make sure we have the value for new subscribers going forward.
___
*[src/service-worker/ServiceWorker.ts, line 789 at r1](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/860#-MnU6EeBAttlsyoZRznp:-MnU6EeBAttlsyoZRznq:bq4ja0b) ([raw file](https://github.com/onesignal/onesignal-website-sdk/blob/bef74d6d9ef7e76818423fbb85ee48d730321f7b/src/service-worker/ServiceWorker.ts#L789)):*
> ```XML
>     const launchUrl: string = await ServiceWorker.getNotificationUrlToOpen(notificationData);
>     const notificationOpensLink: boolean = ServiceWorker.shouldOpenNotificationUrl(launchUrl);
>     const appId = await ServiceWorker.getAppId();
> ```

:thumbsup: , this change ensures we have the appId for both existing a new subscribers even when the DB value isn't there.

There is another another usage of `Database.get<string>("Ids", "appId");` in this file we should clean up, looks be on the received receipts code. I'll leave it up to you if you want to change in this PR or if you want to do it in another since this is out of scope of fixing the click issue.

jkasten2 (approved):
The manual testing rules out the `undefined` issue in the debugger.

:lgtm:

> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/860#-:-MnWjfh7DVcUxe4FgbpL:byc40hy)* status: all files reviewed, 2 unresolved discussions (waiting on @rgomezp)

PR #863, opened 2021-11-17
Single url setup for service worker

rgomezp:
# Description
## 1 Line Summary
Simplifies the service worker update mechanism to require only a single service worker file.

## Details
When pushing WebSDK updates, we want the service worker to update so that any potential changes propagate not only to the page context but also to the underlying worker. The browser will automatically do this if the service worker file contents are different between versions or if the 24-hour cache expiry is hit. However, the hosted service worker file used by OneSignal integrated sites is not actually the full service worker component of the SDK but rather a simple `importScripts` call to our CDN-hosted service worker file. E.g: 
   `importScripts('https://cdn.onesignal.com/sdks/OneSignalSDKWorker.js');`

Thus, the way to force the browser to re-install the service worker (and thus re-fetch the remote worker file) is to use two files:
- `OneSignalSDKWorker.js`
- `OneSignalSDKUpdaterWorker.js`

This ensures that the SW context and page-context are up to date and in sync (important if there are changes across both that require them to be up to date). **Note:** AMP integrations don't use (and have never used) the previous updater-worker file as limited by the AMP web push API. This was not a problem because the "page context" was virtually non-existent due to the AMP restriction of running 3rd party scripts. The SW updates automatically upon cache expiry.

Prior to this PR: Note that the contents of the two files are exactly the same (the import statement above). The browser also re-installs the worker if the file name is different which we take advantage of here.

A negative of this approach is that sites must host two separate files. Luckily, there is a third way to force-update the worker — by changing the query params at the end of the worker path. By leveraging this, we can simplify the integration to require hosting a single worker import file.

## Changes
The OneSignal internal initialization code path always calls `installWorker` on every page load. This is because this function contains the logic to check for a service worker update.

We use the `workerNeedsUpdate` helper function to determine whether we need to update. Currently, we use the `getWorkerVersion` helper to fetch the SW version number via the worker-messenger postmam call. We then compare this to `Environment.version()` which is the page-context SDK version. If they are different, we update. To update, we call the `installAlternatingWorker` which switches back and forth between the two files e.g: A -> B -> A -> B indefinitely.

The core logic-change comes in the `installAlternatingWorker` function. Since this is only called if we have determined we should install the worker, it is safe to modify this code to not alternate between worker files (current) but rather consistently install a single file with a query parameter appended to the end. We can remove ALL code related to this alternating mechanism. Everything else stays the same. E.g:

# Systems Affected
   - [x] WebSDK
   - [ ] Backend
   - [ ] Dashboard

# Validation
### Info
Tests were updated to no longer use the alternating Service Worker install regime.

### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [x] No Typescript warnings

Other:

### Checklist
      - No visual changes

---

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/863)

### DIFF
--- a/express_webpack/amp/index.html
+++ b/express_webpack/amp/index.html
@@ -22,7 +22,6 @@
     var OneSignal = window.OneSignal || [];
     OneSignal.push(function() {
       OneSignal.SERVICE_WORKER_PARAM = { scope: "/" + SERVICE_WORKER_PATH };
-      OneSignal.SERVICE_WORKER_UPDATER_PATH = SERVICE_WORKER_PATH + "OneSignalSDKUpdaterWorker.js";
       OneSignal.SERVICE_WORKER_PATH = SERVICE_WORKER_PATH + "OneSignalSDKWorker.js";
 
       OneSignal.init({ appId });
--- a/express_webpack/index.html
+++ b/express_webpack/index.html
@@ -24,7 +24,6 @@
     var OneSignal = window.OneSignal || [];
     OneSignal.push(function() {
       OneSignal.SERVICE_WORKER_PARAM = { scope: "/" + SERVICE_WORKER_PATH };
-      OneSignal.SERVICE_WORKER_UPDATER_PATH = SERVICE_WORKER_PATH + "OneSignalSDKUpdaterWorker.js";
       OneSignal.SERVICE_WORKER_PATH = SERVICE_WORKER_PATH + "OneSignalSDKWorker.js";
 
       OneSignal.init({
--- a/src/OneSignal.ts
+++ b/src/OneSignal.ts
@@ -899,7 +899,6 @@ export default class OneSignal {
    * However, the init options 'path' should be used to specify the folder path instead since service workers will not
    * auto-update correctly on HTTPS site load if the config init options 'path' is not set.
    */
-  static SERVICE_WORKER_UPDATER_PATH = 'OneSignalSDKUpdaterWorker.js';
   static SERVICE_WORKER_PATH = 'OneSignalSDKWorker.js';
 
   /**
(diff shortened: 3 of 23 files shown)


jkasten2 (approved):
> Reviewed 2 of 4 files at r1, 21 of 21 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/863)* status: all files reviewed, 2 unresolved discussions (waiting on @rgomezp)

___
*[src/managers/ServiceWorkerManager.ts, line 204 at r2](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/863#-MouO3CNDMk5xxnW-_k4:-MouO3CNDMk5xxnW-_k5:b6rft1a) ([raw file](https://github.com/onesignal/onesignal-website-sdk/blob/5655fa0ce1939c2950f5c915368752f68275ab90/src/managers/ServiceWorkerManager.ts#L204)):*
> ```XML
>     }
> 
>     // 3. Different href?, asking if (path + filename [A or B] + queryParams) is different
> ```

Remove "[A OR B]" from comment.
___
*[src/managers/ServiceWorkerManager.ts, line 256 at r2](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/863#-MouPTghAq0cg6FHM2Z3:-MouPTghAq0cg6FHM2Z4:b-35tfe8) ([raw file](https://github.com/onesignal/onesignal-website-sdk/blob/5655fa0ce1939c2950f5c915368752f68275ab90/src/managers/ServiceWorkerManager.ts#L256)):*
> ```XML
>     workerMessenger.off();
> 
>     workerMessenger.on(WorkerMessengerCommand.NotificationDisplayed, async data => {
> ```

These look to be some type cleanups but don't seem to be related to this PR. We can leave them here since you made the changes already and they are small, in it's own commit, and I see they won't affect any logic.

PR #867, opened 2021-12-10
Fix custom link user gesture requirement [custom-link] [safari]

rgomezp:
Motivation: instead of relying on the asynchronous IndexDB reads for the `isPushEnabled` and `isOptedOut` values, we switch to use the synchronous `LocalStorage` getters instead.

The reason for this is simple, asynchronous DB reads make use of the message queue which queues asynchronous calls and is outside of the event loop call stack.

As soon as the message queue is used, the entire call stack loses its designation as "user-initiated" even though it may have been. We must thus avoid using any executions that would require use of the [message queue](https://flaviocopes.com/javascript-event-loop/#the-message-queue) to ensure the requirement is met.

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/867)

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
     "build:dev-stag": "./build/scripts/build.sh -f development -t staging"
   },
   "config": {
-    "sdkVersion": "151510"
+    "sdkVersion": "151511"
   },
   "repository": {
     "type": "git",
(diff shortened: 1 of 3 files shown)


jkasten2 (approved):
:lgtm:

> Reviewed 1 of 1 files at r1, 1 of 1 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/867#-:-MqaZfNf6-tic5QcRGox:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @rgomezp)

PR #868, opened 2021-12-14
Install worker - should not use console.log, reorder operations

rgomezp:
This PR fixes issue where we were wrongly using `console.log` instead of `Log.info` AND the order of operations was wrong. We should only log that we are installing the worker _after_ checking whether we _should_ install the worker.

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/868)

### DIFF
--- a/src/managers/ServiceWorkerManager.ts
+++ b/src/managers/ServiceWorkerManager.ts
@@ -387,11 +387,11 @@ export class ServiceWorkerManager {
    */
 
   public async installWorker() {
-    console.log("Installing worker...");
     if (!await this.shouldInstallWorker()) {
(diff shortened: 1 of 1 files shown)


jkasten2 (approved):
:lgtm:

> Reviewed 1 of 1 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/868#-:-MqukvfB6TgwGyqorDJO:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @rgomezp)

PR #872, opened 2022-01-13
Channel subscription observers

rgomezp:
# Description
## 1 Line Summary
Adds channel subscription observers for email and SMS in the SDK.

## Details
We use event listeners/observers across our SDKs to track subscription changes, notification events, user actions, prompt behavior, and more. Adding events for status changes of our new channels allows users to do more with the OneSignal integration.

### New Events
- `emailSubscriptionChange`
- `smsSubscriptionChange`

The event object:
```
// email
{
   email: "josh@onesignal.com"
} 
// SMS

{
   sms: "+12149874829"
}
```

### Note:
We currently have no way to unsubscribe SMS and EMAIL, just log out which is basically the same as unlinking the browser. So the “subscription change” would really only go from unsubscribed to subscribed in the SDK’s current state.

# Systems Affected
   - [x] WebSDK
   - [ ] Backend
   - [ ] Dashboard

# Validation
### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [x] No Typescript warnings

Other:

### Checklist
   Screen Shot 2022-01-13 at 3 57 58 PM

---

---

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/872)

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
     "build:dev-stag": "./build/scripts/build.sh -f development -t staging"
   },
   "config": {
-    "sdkVersion": "151512"
+    "sdkVersion": "151513"
   },
   "repository": {
     "type": "git",
--- a/src/OneSignal.ts
+++ b/src/OneSignal.ts
@@ -1036,6 +1036,14 @@ export default class OneSignal {
      */
     PERMISSION_PROMPT_DISPLAYED: 'permissionPromptDisplay',
     /**
+     * Occurs when the email subscription changes
+     */
+    EMAIL_SUBSCRIPTION_CHANGED: 'emailSubscriptionChanged',
+    /**
+     * Occurs when the SMS subscription changes
+     */
+    SMS_SUBSCRIPTION_CHANGED: 'smsSubscriptionChanged',
+    /**
      * For internal testing only. Used for all sorts of things.
      */
     TEST_INIT_OPTION_DISABLED: 'testInitOptionDisabled',
--- a/src/managers/channelManager/shared/SecondaryChannel.ts
+++ b/src/managers/channelManager/shared/SecondaryChannel.ts
@@ -3,7 +3,7 @@ import { TagsObject } from "../../../models/Tags";
 // Interface provided outside of this module.
 //   Example: OneSignal.ts as a consumer.
 export interface SecondaryChannel {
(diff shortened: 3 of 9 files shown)


jkasten2 (approved):
2 small nits in the tests, otherwise looks good. Feel free to merge and release without a 2nd approval.

> Reviewed 1 of 1 files at r1, 1 of 1 files at r2, 1 of 1 files at r3, 2 of 2 files at r4, 1 of 1 files at r5, 2 of 2 files at r6, 1 of 1 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/872#-:-MuTqHq0-HLWWPMl8F0r:bxzhlar)* status: all files reviewed, 2 unresolved discussions (waiting on @rgomezp)

___
*[test/unit/public-sdk-apis/setEmail.ts, line 373 at r6](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/872#-MuTpRcW8LkJmAi-W96F:-MuTpRcW8LkJmAi-W96G:brvlm3j) ([raw file](https://github.com/onesignal/onesignal-website-sdk/blob/fc9e584dd07ea4a63c7f7b1b7b39fffef448369a/test/unit/public-sdk-apis/setEmail.ts#L373)):*
> ```XML
>   async t => {
>     const fakeUuid = Random.getRandomUuid();
>     await expectEmailRecordCreationRequest(
> ```

nit, I noticed we used `NockOneSignalHelper.nockPlayerPost();` instead for the SMS test which is clearer. Can we use it for this email test as well?
___
*[test/unit/public-sdk-apis/setEmail.ts, line 382 at r6](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/872#-MuTodOqCz5gyLgWFFPv:-MuTodOqCz5gyLgWFFPw:b4g5442) ([raw file](https://github.com/onesignal/onesignal-website-sdk/blob/fc9e584dd07ea4a63c7f7b1b7b39fffef448369a/test/unit/public-sdk-apis/setEmail.ts#L382)):*
> ```XML
>     const subscriptionChangeEventPromise = awaitSubscriptionChangeEvent("emailSubscriptionChanged");
> 
>     setUserAgent(BrowserUserAgent.FirefoxMacSupported);
> ```

The `setUserAgent` line seems unneeded, if so let's remove it.

PR #900, opened 2022-07-21
User model/system skeleton

rgomezp:
This PR is not intended to be reviewed with a high level of scrutiny as it will first be merged into the feature branch for user model v1.

The goal is to get some early feedback on the overall system skeleton design and how the different components interface with each other before starting to plug in the API systems.

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/900)


rgomezp:
Will make some more changes before reopening

PR #903, opened 2022-07-28
User model/core v1

rgomezp:
# Description
## 1 Line Summary
This is the Core v1 implementation of the basic system architecture to get early feedback.

## Details
This system is the Core module of the SDK and will be encapsulated into its instance separate from the rest of the SDK (API). The rest of the SDK (API) will interface with the Core module via the model repo. The APIs will change models in the model repo directly, triggering the remote updates where needed.

Untitled Diagram drawio (2)

The WebSDK Core module will be instantiated as part of the initialization process and will live on the OneSignal instance.

### Core Repos
The Core module is driven by logic in the two main repos: the model repo and the operation repo. These repos are subscribable, allowing them to broadcast data to subscribers. When a model changes due to an SDK API, the model repo will broadcast the delta (the change). The operation repo listens for these model repo deltas and uses these to prepare an operation and subsequently a request to the request service. The operation repo will then broadcast the result of the operation.

### Core Caches
The Core module has two caching systems for each repo. 

The model cache will continuously backup model changes using database writes to the browser’s local IndexedDB. We can use a [proxy observer](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy) in the model repo to trigger these synchronizations to the model cache as soon as they happen automatically.

The operation cache will backup individual operations before processing them. Once they complete successfully, they will be removed from the operation cache. This is useful in case the device goes offline or a request fails for whatever reason.

### Hydrator Bus
The hydrator bus is a communication layer between the model and operation repos. It subscribes to operations outcomes from the operation repo and will hydrate them as needed to save to the model repo. The hydrator bus will then use the model repo’s public `set` function to update the model.

---

# Systems Affected
   - [x] WebSDK
   - [ ] Backend
   - [ ] Dashboard

---
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/903)


emawby (commented):
This looks great and is clean! I have a question about the future vision of the Web SDK architecture in general. It looks like we will have a Core module and an API module. Is this the extent to which we will modularize the Web SDK (question for the user model implementation and looking forward into the future)? Do we intend to modularize by product feature (i.e. notifications, location, outcomes, future stuff like IAM)? If we do intend to modularize by product feature I would like to think about how we can make the core module product agnostic and move the product specific types/references outside of the core module. If that is not something we are interested in that is fine as well!

> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/903#-:-N8dnhVI0KdZdh9FPnQA:bxsyev2)* status: 0 of 9 files reviewed, all discussions resolved (waiting on @brismithers, @jkasten2, and @nan-li)

src/core/ModelRepo.ts
```diff
@@ -0,0 +1,114 @@
...
+        break;
+      case Model.Config:
+        targetModel = this.config;
+        break;
+      default:
+        break;
+    }
+
+    // Proxy observer does not trigger on nested objects
+    // We need to change each property individually
```
rgomezp, line 107: after more testing, it appears it does trigger on nested objects, but we still need this function when setting an entire model as to avoid reassignment and overwriting the proxy observer

src/core/caches/ModelCache.ts
```diff
@@ -0,0 +1,30 @@
+// import Database from "../../shared/services/Database";
```
rgomezp, line 1: remove

src/core/caches/ModelCache.ts
```diff
@@ -0,0 +1,30 @@
...
+
+  async load(): Promise<void> {
+    this.identity = await this.get<IdentityModel>(Model.Identity) || DEFAULT_MODELS.identity;
+    this.properties = await this.get<UserProperties>(Model.Properties) || DEFAULT_MODELS.properties;
+    this.subscriptions = await this.get<SubscriptionsCollection>(Model.Subscriptions) || DEFAULT_MODELS.subscriptions;
+    this.config = await this.get<AppConfig>(Model.Config) || DEFAULT_MODELS.config;
+  }
+
+  async put(model: Model, property: string, value: object): Promise<void> {
+    await Database.put("Models", { key: model, value: { [property]: value } });
```
rgomezp, line 24: test to make sure it doesn't replace entire value

src/core/OperationRepo.ts
```diff
@@ -0,0 +1,101 @@
...
+
+  private pushToOperationsQueue(deltaAggregator: DeltaAggregator): void {
+    const modelKeys = Object.keys(deltaAggregator) as Model[];
+
+    for (let i=0; i<modelKeys.length; i++) {
+      const key = modelKeys[i];
+      const delta = deltaAggregator[key];
+
+      if (Object.keys(delta).length) {
+
```
rgomezp, line 51: extra line

src/core/HydratorBus.ts
```diff
@@ -0,0 +1,35 @@
...
+import ModelRepo from "./ModelRepo";
+import OperationRepo from "./OperationRepo";
+import { IdentityModel, Model, UserProperties } from "./types";
+
+type ServerResponse = object;
+
+export class HydratorBus {
+  constructor(private modelRepo: ModelRepo, private operationRepo: OperationRepo) {
+    this.operationRepo.subscribe((res: ServerResponse) => {
+      this.hydrate(res);
```
rgomezp, line 12: how will the hydrator bus know which models to hydrate based on the response?

rgomezp:
> This looks great and is clean! I have a question about the future vision of the Web SDK architecture in general. It looks like we will have a Core module and an API module. Is this the extent to which we will modularize the Web SDK (question for the user model implementation and looking forward into the future)? Do we intend to modularize by product feature (i.e. notifications, location, outcomes, future stuff like IAM)? If we do intend to modularize by product feature I would like to think about how we can make the core module product agnostic and move the product specific types/references outside of the core module. If that is not something we are interested in that is fine as well!

@emawby the plan is for the API to interact with the core module via `OneSignal` and `OneSignal.user` API namespaces.

Right now we have two webpack entry points for the `page` SDK and the service worker file. A lot of this code is shared between both. My hope is to take this opportunity to reduce this shared code as much as possible.

Yes, the hope is to eventually modularize the SDK further by feature, as you mentioned. I agree that the Core Module should be feature agnostic. What feature-specific code in the Core Module should I think about moving? And what alternative do you propose? Currently, the only models in the model repo are
* identity
* properties
* subscriptions
* config

src/core/utils/DefaultModelObjects.ts
```diff
@@ -0,0 +1,50 @@
+import { SubscriptionsCollection } from "../../onesignal/subscriptions/SubscriptionsCollection";
+import { AppConfig } from "../../shared/models/AppConfig";
+import { IdentityModel, UserProperties } from "../types";
+
+/**
```
rgomezp, line 5: Check what browsers still use ES5, consider dropping support

jkasten2 (approved):
This PR covers the basic architecture that includes model and operation systems and how the related to each other. While this doesn't handle everything the PR covers what it outlines so future PRs can be based on this one to fill in more of the logic. I approve this PR based on that. 

> Reviewed 1 of 1 files at r1, 1 of 1 files at r2, 1 of 1 files at r3, 1 of 1 files at r4, 1 of 1 files at r5, 1 of 1 files at r6, 1 of 1 files at r7, 1 of 1 files at r8, 1 of 1 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/903#-:-N8gWgyFDUXWvjHAxz6P:by8z2hy)* status: all files reviewed, 7 unresolved discussions (waiting on @brismithers, @emawby, @nan-li, and @rgomezp)

___
*[`src/core/ModelRepo.ts` line 29 at r2](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/903#-N8gVzcMAa4K6xPhdp_h:-N8gVzcMAa4K6xPhdp_i:b-wnir83) ([raw file](https://github.com/onesignal/onesignal-website-sdk/blob/0eadd4423f00f9ff24f1d5d07a7627c8796b839d/src/core/ModelRepo.ts#L29)):*
> ```XML
>   constructor(private modelCache: ModelCache) {
>     super();
>     this.identity = this.createObserver(Model.Identity, modelCache.identity);
> ```

We talked about it would be nice to not have the model repo to know about these, however not a blocker for this PR, or even for beta1.

src/core/utils/DefaultModelObjects.ts
rgomezp: https://caniuse.com/es5

With very low numbers of ES5 browsers, we will drop support

PR #910, opened 2022-09-14
User model/core v2

rgomezp:
# Description
## 1 Line Summary
Version 2 of the Core Module.

## Details

This PR is an overhaul of the v1 version of the core module.

### Overview
We drop the previous dependence on the ObservableSlim proxy observer in this PR to trigger changes. We also introduce the concept of executors to handle model-specific logic.

**Core v1 problem:** while making model changes was very easy, it was difficult to map the model change to an eventual REST call. 

**Core v2 solution:** we expose a new way to trigger model changes that will know exactly what change type we want end-to-end.

### Change types
There are three main change types:
1. Add
2. Remove
3. Update

**Add** and **remove** change types describe model-level changes. For example, if we call "add," we are adding an entire model (e.g: subscription model object) to a user—same thing with remove. If we call "update," we update a specific property in a specified model.

**Update** is not called directly (like add & remove) but rather triggered via calling `set` on an `OSModel` instance.

### Model-agnostic nature
This core module version was designed to be as model-agnostic as possible. Model-specific logic exists only in the executors and the core director class.

#### Model Repo
Contains a model store for each model name (e.g: "IdentityModel"). Also contains top-level change-processing logic.
The model repo listens to changes, then processes them based on the change type (add, remove, update).

#### Model Cache
Used by the model repo to synchronize models in memory with disk.

#### Operation Repo
In this much-simplified version, the operation repo will be used to tell executors what to do. Operations contain a unique id to find it after the executor completes, the payload, a reference to the relevant model, and a timestamp used for sorting.

#### Operation Cache
Used by the operation repo to back-up any operations in case they fail or device is disconnected.

#### Executor
An executor contains the delta and operation queues and the processing logic for deltas and operations. The `OSExecutorBase` class is the parent class to model-specific executor types. Model-specific logic mainly resides in the `processDeltaQueue` function for each executor.

#### Core Module Director
The core module director provides user-model-aware interface through which we can operate on the core module from outside while avoiding heavily nested calls.

Screen Shot 2022-09-14 at 5 59 20 PM

More details in individual commits.

# Systems Affected
   - [x] WebSDK
   - [ ] Backend
   - [ ] Dashboard

# Validation
### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [x] No Typescript warnings

Other:

### Checklist

---

## Related Tickets

---

- - -
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/910)


src/core/modelRepo/OSModelStore.ts
```diff
@@ -0,0 +1,36 @@
...
+
+export class OSModelStore<Model>
+  extends Subscribable<ModelStoreChange<SupportedModel>>
+  {
+    public models: {[key: string]: OSModel<Model>} = {};
+
+    constructor(modelArray: OSModel<Model>[] = []) {
+      super();
+      modelArray.forEach(model => {
+        this.models[model.id] = model;
```
rgomezp, line 15: subscribe models

src/core/modelRepo/OSModelStore.ts
```diff
@@ -0,0 +1,36 @@
...
+      });
+    }
+
+    public add(model: OSModel<Model>, noRemoteSync?: boolean): void {
+      this.subscribeUpdateListener(model);
+      this.models[model.id] = model;
+      this.broadcast(new ModelStoreAdded(model.id, model, noRemoteSync));
```
rgomezp, line 22: use unsubscribe callback to unsub after model removal

src/core/operationRepo/Operation.ts
```diff
@@ -0,0 +1,50 @@
...
+import { isPropertyDelta, isPureObject } from "../utils/typePredicates";
+
+export class Operation<Model> {
+  readonly id: string;
+  readonly payload: {[key: string]: any};
+  readonly model: OSModel<Model>;
+  readonly timestamp: number;
+
+  constructor(readonly changeType: CoreChangeType, readonly modelName: ModelName, deltas: CoreDelta<Model>[]) {
+    this.id = Math.random().toString(36).substring(2);
```
rgomezp, line 14: consider alternatives

src/core/executors/OSExecutorBase.ts
```diff
@@ -0,0 +1,111 @@
...
+    return finalChangeType;
+  }
+
+  protected _processOperationQueue(): void {
+    this._operationQueue = [...OperationCache.getOperations<Model>(), ...this._operationQueue];
+    while (this._operationQueue.length > 0) {
+      const operation = this._operationQueue.shift();
+
+      if (!operation) {
+        continue;
```
rgomezp, line 90: return

src/core/executors/OSExecutorBase.ts
```diff
@@ -0,0 +1,111 @@
...
+        res = this._executeRemove?.call(operation);
+      } else if (operation?.changeType === CoreChangeType.Update) {
+        res = this._executeUpdate?.call(operation);
+      }
+      // HYDRATE
+      if (res.success) {
+        operation?.model.hydrate(res.result);
+        OperationCache.delete(operation?.id);
+      } else {
+        this._enqueueOperation(operation);
```
rgomezp, line 107: add a comment regarding potential changes to retry logic

src/core/executors/OSExecutorBase.ts
```diff
@@ -0,0 +1,111 @@
...
+  protected _processOperationQueue(): void {
+    this._operationQueue = [...OperationCache.getOperations<Model>(), ...this._operationQueue];
+    while (this._operationQueue.length > 0) {
+      const operation = this._operationQueue.shift();
+
+      if (!operation) {
+        continue;
+      }
+
+      let res: ExecutorResult = { success: false };
```
rgomezp, line 93: add something like `retriable`

src/core/executors/OSExecutorBase.ts
```diff
@@ -0,0 +1,111 @@
...
+    } else {
+      throw new Error("Unsupported change type");
+    }
+
+    return finalChangeType;
+  }
+
+  protected _processOperationQueue(): void {
+    this._operationQueue = [...OperationCache.getOperations<Model>(), ...this._operationQueue];
+    while (this._operationQueue.length > 0) {
```
rgomezp, line 86: clone the array before the iteration

src/core/executors/OSExecutorBase.ts
rgomezp: remove if not fixing a typescript warning

rgomezp:
prefer `unknown` over `any`, or if you know it'll be non-null use `object`

rgomezp:
drop `OS` prefix from classes/files

rgomezp:
make unit test for add, remove model

src/core/models/ExecutorResult.ts
```diff
@@ -0,0 +1,4 @@
+export type ExecutorResult = {
+  success: boolean;
+  result?: any;
```
rgomezp, line 3: see if you can fix the typescript warning and make this non-optional

test/unit/core/ModelRepo.ts
```diff
@@ -0,0 +1,130 @@
...
+test("Update model properties -> delta is broadcasted", async t => {
+  const broadcastSpy = sinonSandbox.spy(ModelRepo.prototype as any, "broadcast");
+  const newSub = generateNewSubscription();
+  await global.coreDirector.add(ModelName.EmailSubscriptions, newSub);
+  newSub.set("rooted", true); // trigger model update
+  t.true(broadcastSpy.calledTwice);
+});
+
+test("Add subscription -> delta is broadcasted with correct change type and payload", async t => {
+  const broadcastSpy = sinonSandbox.spy(ModelRepo.prototype as any, "broadcast");
```
rgomezp, line 100: instead of using spies, we have the API to subscribe directly

test/unit/core/OSModel.ts
```diff
@@ -0,0 +1,58 @@
...
+
+test("`Set` function updates data", async t => {
+  const newSub = generateNewSubscription();
+  t.is(newSub.data?.rooted, undefined);
+  newSub.set("rooted", true);
+  t.is(newSub.data?.rooted, true);
+});
+
+test("`Set` function broadcasts update event", async t => {
+  const broadcastSpy = sinonSandbox.spy(OSModel.prototype as any, "broadcast");
```
rgomezp, line 18: subscribe directly

test/unit/core/OSModel.ts
```diff
@@ -0,0 +1,58 @@
...
+    id: "123",
+    token: "myToken",
+  };
+  const decodedSub = OSModel.decode(encodedSub);
+  t.deepEqual(decodedSub, new OSModel(ModelName.EmailSubscriptions, "333333", {
+    type: SubscriptionType.Email,
+    id: "123",
+    token: "myToken",
+  }));
+});
```
rgomezp, line 58: Add test for catching corrupted data not properly decoded

test/unit/core/OperationRepo.ts
```diff
@@ -0,0 +1,167 @@
...
+  modelRepo?.broadcast(delta1);
+  modelRepo?.broadcast(delta2);
+
+  clock.tick(1001);
+
+  t.true(broadcastSpy.calledTwice);
+  t.true(processDeltaSpy.calledTwice);
+  t.is(executor?.operationQueue.length, 1);
+  t.is(executor?.deltaQueue.length, 0);
+});
```
rgomezp, line 125: see if it makes sense to also inspect the changes on the operation itself (alias values)

build/config/tsconfig.base.json
```diff
@@ -34,6 +34,7 @@
     "lib": [
       "webworker",
       "es6",
+      "es2017",
```
rgomezp, line 37: Add details in commit on why this change is needed

test/unit/core/OSModel.ts
rgomezp: TS should prevent this at compile time

jkasten2 (changes requested):
> Reviewed 59 of 59 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/910)* status: all files reviewed, 1 unresolved discussion (waiting on @brismithers, @emawby, @nan-li, and @rgomezp)

___
*[`src/OneSignalDeprecated.ts` line 1 at r1](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/910#-NDAFZ3Z1BzGDPVCw-CB:-NDAFZ3Z1BzGDPVCw-CC:b4o0h9t) ([raw file](https://github.com/onesignal/onesignal-website-sdk/blob/8807b9174f2a5a8e79ab18ca8f517458e4423120/src/OneSignalDeprecated.ts#L1)):*
> ```XML
> // import bowser from 'bowser';
> ```

Every line in this file has been commented out. Should delete instead if we don't need anything from it.

PR #914, opened 2022-09-30
User model/notifications namespace

rgomezp:
# Description
## 1 Line Summary
Migrate OneSignal notification functions into separate namespace.

## Details
This PR migrates the notification-related functions into their own namespace. Moving them results in having to make changes to test files also.

More info in commits.

# Systems Affected
   - [x] WebSDK
   - [ ] Backend
   - [ ] Dashboard

# Validation
### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [x] No Typescript warnings

Other:

### Checklist
      - No UI changes

---

## Related Tickets

---

- - -
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/914)


test/unit/modules/entryInitialization.ts
```diff
@@ -455,7 +455,7 @@ test("Existing OneSignal array before OneSignalSDK.js loaded ES5", async t => {
   let didCallFunction = false;
   const preExistingArray = [() => {
     didCallFunction = true;
-    (<any>window).OneSignal.setDefaultNotificationUrl("test");
+    (<any>window).OneSignal.showNativePrompt();
```
rgomezp, line 458: move and rename to `requestPermission`

PR #920, opened 2022-10-19
[User model] migrate more onesignal functions

rgomezp:
# Description
## 1 Line Summary
With the exception of outcomes functions, all other functions on `OneSignal` class are moved to respective namespaces or removed.

### Slidedown Namespace
New slidedown namespace will be home to relevant slidedown functions:
* `promptPush`
* `promptSms`
* `promptEmail`
* `promptSmsAndEmail`
* `promptPushCategories`

### Privacy Consent Functions
Implemented privacy consent functions with names to match other SDKs:
* `setPrivacyConsent`
* getPrivacyConsent`
* setRequiresPrivacyConsent`

### Notification Subscription Functions
Moved notification subscription related functions previously on `OneSignal` elsewhere such as `isPushNotificationsEnabled` which now resides on the `SubscriptionManager`.

**`addListenerForNotificationOpened`** will be replaced by the `notification` namespace's `on` listener function

**`optOut`** will be replaced by the `notification` namespace's `disable` function

**`getSubscription`** was a misnomer since it simply returned `!isOptedOut` which is not entirely accurate since they could not be actually subscribed E2E.

### Removes tags & external id functions
Now handled by core-module / user namespace.

### Removes identifier functions like `getIdsAvailable`
Previously deprecated

# Systems Affected
   - [x] WebSDK
   - [ ] Backend
   - [ ] Dashboard

# Validation
### Checklist
      - many broken tests for now
      - I have tested that it **builds**

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [x] No Typescript warnings

Other:

---

- - -
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/920)


src/onesignal/OneSignal.ts
```diff
@@ -659,8 +652,6 @@ export default class OneSignal {
   static DeviceRecord = DeviceRecord;
   static SecondaryChannelDeviceRecord = SecondaryChannelDeviceRecord;
 
-  static notificationPermission = NotificationPermission;
```
rgomezp, line 774: (unused)

rgomezp (commented):
Moving `disable` from notification namespace to `user.pushSubscription` as `enable`

jkasten2 (approved):
:lgtm:

> Reviewed 1 of 1 files at r6, 2 of 4 files at r14, 1 of 2 files at r15, 4 of 8 files at r16, 2 of 2 files at r17, 1 of 1 files at r18, 1 of 2 files at r19, 1 of 1 files at r20, 1 of 1 files at r21, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/920#-:-NEriyjk2Z0rF3n5JSsC:bnfp4nl)* status: all files reviewed, 3 unresolved discussions

PR #930, opened 2022-11-18
Feat/safari 16

rgomezp:
# Description
## 1 Line Summary
Adds support for Safari 16 (VAPID).

## Details
Safari 16 is introducing the standard [PushAPI ](https://developer.apple.com/documentation/usernotifications/sending_web_push_notifications_in_safari_and_other_browsers)that Chrome, Firefox, and other browsers use today. 

Safari 16 still supports it’s proprietary “Safari Push Notifications” on macOS, but the standard API is preferred since it supports more features. iOS/iPadOS 16.4 also supports the standard PushAPI. Apple will almost certainly drop the old API for macOS, but it may be a number of years before then so we will continue supporting it. We still want to continue supporting macOS 12.x and older, since it requires end users update to macOS 13 to get it.

### Device Type
Introducing `device_type: 17` for W3C version of Safari. Since the two Safari push types are fundamentally different, a new device_type value would clearly separate them and would keep things consistent in our system.

### Data Purge Problem:
Safari has a 7 day window where after the user doesn’t visit the site, it will purge all data like IndexedDB and localStorage. We introduce a way to detect whether the user was previously subscribed to legacy notifications and avoid creating a second subscription.
 
### iOS Support
With the introduction of web push for Safari on iOS, these changes will also work for that feature. However, it _will_ require the site to host a manifest.json file. This PR updates our sandbox environment with that file.

# Systems Affected
   - [x] WebSDK
   - [x] Backend
   - [ ] Dashboard

# Validation
### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [x] No Typescript warnings

Other:

### Info
Screen Shot 2023-03-06 at 2 19 21 PM
Currently Safari 16 does not display the custom icon. This is an upstream limitation.

### Checklist

---

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Website-SDK/930)

### DIFF
--- a/src/helpers/MainHelper.ts
+++ b/src/helpers/MainHelper.ts
@@ -214,4 +214,13 @@ export default class MainHelper {
     const subscription = await OneSignal.database.getSubscription();
     return subscription.deviceId || undefined;
   }
+
+  static async getCurrentPushToken(): Promise<string | undefined> {
+    const registration = await OneSignal.context.serviceWorkerManager.getRegistration();
+    if (!registration) {
+      return undefined;
+    }
+    const subscription = await registration.pushManager.getSubscription();
+    return subscription?.endpoint;
+  }
 }
--- a/build/scripts/build.sh
+++ b/build/scripts/build.sh
@@ -66,7 +66,17 @@ echo "Unknown options -> ${POSITIONAL}"
 set -- "${POSITIONAL[@]}" # restore positional parameters
 
 # build local SW file for dev env
-./build/scripts/buildServiceWorker.sh $BUILD_ORIGIN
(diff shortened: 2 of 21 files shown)


src/helpers/MainHelper.ts
```diff
@@ -214,4 +214,13 @@ export default class MainHelper {
     const subscription = await OneSignal.database.getSubscription();
     return subscription.deviceId || undefined;
   }
+
+  static async getCurrentPushToken(): Promise<string | undefined> {
```
jkasten2, line 218: This function was added in this PR but I don't see it being used anywhere.

PR #935, opened 2022-11-29
`setAppId` function implementation

rgomezp:
Motivation: to match mobile, we want to provide `setAppId` function that simply calls `OneSignal.init(<app_id>);`

- - -
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/935)


rgomezp:
Let's hold off on this until we have more clarity on how we want to handle `setAppId`

PR #936, opened 2022-11-30
[User model] create user until we have at least 1 subscription

rgomezp:
Motivation: to avoid creating tons of users we will never use, we should wait until we have at least 1 sub

- - -
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/936)

### DIFF
--- a/src/onesignal/UserNamespace.ts
+++ b/src/onesignal/UserNamespace.ts
@@ -28,7 +28,7 @@ export default class UserNamespace {
           await this.coreDirector.getSmsSubscriptionModels(),
           await this.coreDirector.getEmailSubscriptionModels(),
         );
-        await this._currentUser.setupNewUser();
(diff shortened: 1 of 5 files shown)


src/onesignal/UserNamespace.ts
```diff
@@ -147,7 +149,9 @@ export default class UserNamespace {
 
     const newSubscription = new OSModel<SupportedModel>(ModelName.SmsSubscriptions, subscription);
 
-    this.coreDirector.add(ModelName.SmsSubscriptions, newSubscription, true).catch(e => {
+    this.coreDirector.add(ModelName.SmsSubscriptions, newSubscription, true).then(() => {
+      User.singletonInstance?.sendUserCreate();
```
rgomezp, line 153: this._currentUser

PR #940, opened 2022-12-02
Fix: we should only create user if the user doesn't already exist

rgomezp:
We should only create user if it doesn't already exist when adding SMS and Email

- - -
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/940)

### DIFF
--- a/src/onesignal/UserNamespace.ts
+++ b/src/onesignal/UserNamespace.ts
@@ -12,7 +12,7 @@ export default class UserNamespace {
   private _currentUser?: User;
   public userLoaded: Promise<void>;
 
-  readonly pushSubscription = new PushSubscriptionNamespace();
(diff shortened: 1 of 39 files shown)


src/onesignal/UserNamespace.ts
```diff
@@ -125,11 +125,19 @@ export default class UserNamespace {
     };
     const newSubscription = new OSModel<SupportedModel>(ModelName.EmailSubscriptions, subscription);
 
-    this.coreDirector.add(ModelName.EmailSubscriptions, newSubscription, true).then(() => {
-      this._currentUser?.sendUserCreate();
-    }).catch(e => {
-      throw e;
-    });
+    if (User.singletonInstance?.identified) {
```
rgomezp, line 128: rename to `hasOneSignalId`

PR #946, opened 2022-12-08
Add JWT token functionality

rgomezp:
* Add JWT setter and getter to IndexedDB
* Create helper function to get JWT token from database
* Add JWT header to appropriate RequestService functions

- - -
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/946)

### DIFF
--- a/src/core/requestService/RequestService.ts
+++ b/src/core/requestService/RequestService.ts
@@ -17,10 +17,10 @@ export class RequestService {
    * @param requestMetadata - { appId }
    * @param requestBody - The user's properties, identity, and subscriptions
    */
-  static createUser(requestMetadata: RequestMetadata, requestBody: Partial<UserData>)
+  static async createUser(requestMetadata: RequestMetadata, requestBody: Partial<UserData>)
     : Promise<OneSignalApiBaseResponse> {
       const { appId } = requestMetadata;
-      return OneSignalApiBase.post(`apps/${appId}/users`, requestBody);
+      return OneSignalApiBase.post(`apps/${appId}/users`, requestBody, requestMetadata.jwtHeader);
   }
 
   /**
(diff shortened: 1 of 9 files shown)


src/core/requestService/helpers.ts
```diff
@@ -64,3 +65,12 @@ export function processIdentityOperation<Model>(operation: Operation<Model>): {
   };
 }
 
+export async function getJWTHeaderIfIdentityVerificationEnabled(): Promise<{ "Authorization": string }[] | undefined> {
+  if (OneSignal.config?.userConfig?.identityVerificationEnabled) {
```
rgomezp, line 69: remove conditional

src/core/requestService/helpers.ts
```diff
@@ -64,3 +65,12 @@ export function processIdentityOperation<Model>(operation: Operation<Model>): {
...
+    const jwtToken = await Database.getJWTToken();
```
rgomezp, line 70: if undefined, return []

src/core/requestService/helpers.ts
rgomezp: rename function to getJWTHeader

src/core/requestService/RequestService.ts
```diff
@@ -176,10 +197,13 @@ export class RequestService {
    * @param subscriptionId - subscription id
    * @param identity - identity label & id
    */
-  static identifyUserForSubscription(requestMetadata: RequestMetadata, subscriptionId: string, identity: IdentityModel):
+  static async identifyUserForSubscription(
```
rgomezp, line 200: needs JWT

rgomezp (commented):
We actually want to copy the JWT token into operations so that if we switch users after calling addTag for example, we will still send the correct JWT token.

PR #952, opened 2022-12-12
User model/event listeners

rgomezp:
More details in commits

- - -
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/952)

### DIFF
--- a/src/shared/models/StructuredNotification.ts
+++ b/src/shared/models/StructuredNotification.ts
@@ -1,7 +1,7 @@
 type StructuredNotification = {
-  id?: string;
+  id: string;
+  content: string;
   heading?: string;
-  content?: string;
   url?: string;
   data?: object;
   rr?: string;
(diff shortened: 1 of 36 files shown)


src/shared/models/StructuredNotification.ts
```diff
@@ -0,0 +1,21 @@
+type StructuredNotification = {
```
jkasten2, line 1: Are all these fields really optional? Can we make some of the known ones required?

jkasten2 (changes requested):

src/shared/models/StructuredNotification.ts
rgomezp: we could probably make `id` `heading` and `content` required. the other ones I'm not 100 percent sure on since the `rawNotification` is of `any` type

jkasten2: Ya, I think the two things we know for sure is `id` and `content` will be in the payload.

rgomezp: Fixed

jkasten2 (approved):

PR #953, opened 2022-12-15
User model - Add deferred loading API, remove stub API

jkasten2:
# Description
New `OneSignalDeferred.push` API to replace `OneSignal.push` which allows removing all stub code.

## Details
With User Model we are adding nested objects (acting like namespaces), example `OneSignal.User.addTags()`. The stub code we have today wasn't designed to account for this and would be complex to add it. Instead we can make a breaking change and remove all stubbing logic. This gives us an opportunity to make sure everyone is using a new `OneSignalDeferred.push` interface (added in this PR) when they upgrade.

The new public API will looks this:
```Typescript
OneSignalDeferred.push(function(onesignal: OneSignal) {
   onesignal.User.addTag("key", "value");
});
```

We will be providing a `OneSignal` instance instead of telling the consumer to defined a `window.OneSignal` instance. Due note that we are still defining `window.OneSignal` in this PR in `OneSignalPageSDKES6.js` (pageSdkInit.ts) as not all internals have been upgraded, and may even want to keep it for easier debugging in the JS console.

In this PR we are not renaming any files.

### Added
```Typescript
type OneSignalDeferredLoadedCallback = (onesignal: OneSignal) => void;
// On OneSignalDeferred
static push(item: OneSignalDeferredLoadedCallback) { }
```
### Removed
* Using `OneSignal.functionName` directly is no longer supported, due to new deferred loading
    - Must use the new `OneSignal` instance instead provided by `OneSignalDeferred.push`.
* `object[]` support from `OneSignal.push`
*  Removed `OneSignal.SERVICE_WORKER_PARAM` and `OneSignal.SERVICE_WORKER_PATH`
   - Use existing API of passing `serviceWorkerParam` and `serviceWorkerPath` as init params

### Changed
* `OneSignal.push` doesn't work until `OneSignalPageSDKES6.js` is loaded.
   - The intent is `OneSignalDeferred.push` should be used instead, using the provided instance.
   - Only use the `OneSignal` global object for JS console debugging.

# Validation
## Tests
* Updated the entryInitialization.ts tests, most were removed since most covered stubs which is now removed.
   - Could not run them to due tests not building on the `user-model/v1` branch.
 * Manually tested on the provided sandbox index.html, tested to ensure user gets creating after accepting notification permission.
### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [X] No Typescript warnings

Other:

### Checklist

---

## Related Tickets

---

- - -
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/953)


rgomezp (changes requested):

src/entries/sdk.ts
```diff
@@ -1,10 +1,9 @@
 /**
  * This is OneSignalSDK.js (ES5)
- *   * This is an entry point for pages and older service workers.
- *       - Developers are now instructed to use OneSignalSDKWorker.js directly for service workers
+ *   * This is an entry point for pages
+ *       - Developers are instructed to use OneSignalSDKWorker.js in their Service Worker instead
```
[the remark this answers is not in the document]
jkasten2: I removed this, since we don't even need to bring up the service worker in this context.

src/page/models/OneSignalDeferredLoadedCallback.ts
```diff
@@ -0,0 +1,3 @@
+import OneSignal from "../../onesignal/OneSignal";
+
+export type OneSignalDeferredLoadedCallback = (onesignal: OneSignal) => void;
```
rgomezp, line 3: newline

rgomezp (approved):
just one tiny nit, missing a newline. but other than that LGTM

src/page/models/OneSignalDeferredLoadedCallback.ts
jkasten2: done

PR #964, opened 2022-12-20
Update `sendSelfPush` to send `onesignalId` instead of device id

rgomezp:
Should use new standard

- - -
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/964)

### DIFF
--- a/src/onesignal/NotificationsNamespace.ts
+++ b/src/onesignal/NotificationsNamespace.ts
@@ -58,9 +58,8 @@ export default class NotificationsNamespace extends EventListenerBase {
     return true;
   }
 
-  /**
-   * @PublicApi
-   */
+
+  /*
   async sendSelfPush(title: string = 'OneSignal Test Message',
                               message: string = 'This is an example notification.',
                               url: string = `${new URL(location.href).origin}?_osp=do_not_open`,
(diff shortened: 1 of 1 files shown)


src/shared/api/OneSignalApiShared.ts
```diff
@@ -21,11 +21,11 @@ export default class OneSignalApiShared {
     return OneSignalApiBase.put(`players/${playerId}`, { app_id: appId, ...options });
   }
 
-  static sendNotification(appId: string, playerIds: Array<string>, titles, contents, url, icon, data, buttons) {
+  static sendNotification(appId: string, onesignalId: string, titles, contents, url, icon, data, buttons) {
     var params = {
       app_id: appId,
       contents: contents,
-      include_player_ids: playerIds,
+      include_aliases: { onesignalId: [onesignalId] },
```
jkasten2, line 28: This own be allowed without auth.

jkasten2 (changes requested):

src/shared/api/OneSignalApiShared.ts
rgomezp: I'm not sure what this means

jkasten2: Sorry meant to say `include_aliases` requires auth to use. Specifically only the app's REST API key will work for this. No plans to support a client side JWT for this field.

jkasten2 (approved):

PR #966, opened 2022-12-20
Update Index.html of Web Sandbox

rgomezp:
Update with User-model-specific code

- - -
This change is [Reviewable](https://reviewable.io/reviews/onesignal/onesignal-website-sdk/966)

### DIFF
--- a/express_webpack/index.html
+++ b/express_webpack/index.html
@@ -6,6 +6,7 @@
 <script src="https://localhost:4001/sdks/web/v16/Dev-OneSignalSDK.page.js" defer></script>
 <script>
     const SERVICE_WORKER_PATH = "push/onesignal/";
+    let showEventAlertToggleSetting = false;
 
     function getUrlQueryParam(name) {
         var url = window.location.href;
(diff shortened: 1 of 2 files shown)


express_webpack/index.html
```diff
@@ -99,50 +103,94 @@
...
+      alert(`OneSignal event ${eventName} fired!\n\nPayload:\n${JSON.stringify(payload, null, 2)}`);
+    }
+
     function sendUniqueOutcome() {
-        OneSignalDeferred.push(function(onesignal) {
-            const outcomeName = document.querySelector('#unique_outcome_name').value;
-            onesignal.sendUniqueOutcome(outcomeName);
-        });
+      const outcomeName = document.querySelector('#unique_outcome_name').value;
+      OneSignal.Session.sendUniqueOutcome(outcomeName);
```
jkasten2, line 166: Possible race condition with OneSignal not being loaded yet. All OneSignal calls should go through `OneSignalDeferred.push`

jkasten2 (changes requested):

express_webpack/index.html
rgomezp: These are all triggered by button clicks so that's a non issue. Also, the problem with putting these in deferred, is the way you implemented it, there is no continuous processing of the functions pushed onto it, so they just get added to the array and nothing happens.

Maybe we need to revisit that.

jkasten2 (approved):

PR #983, opened 2023-01-18
User model/retry login

rgomezp:
# Description
## 1 Line Summary
Adds retry logic on login and includes some minor refactoring/cleanup to login related functions.

See commits for more details

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Website-SDK/983)

### DIFF
--- a/src/page/managers/LoginManager.ts
+++ b/src/page/managers/LoginManager.ts
@@ -9,14 +9,7 @@ import { OSModel } from "../../core/modelRepo/OSModel";
 import { SupportedIdentity } from "../../core/models/IdentityModel";
 import MainHelper from "../../shared/helpers/MainHelper";
 import { awaitableTimeout } from "../../shared/utils/AwaitableTimeout";
-
-const RETRY_BACKOFF: { [key: number]: number } = {
-  5: 10_000,
-  4: 20_000,
-  3: 30_000,
-  2: 30_000,
-  1: 30_000,
-};
+import { RETRY_BACKOFF } from "../../shared/api/RetryBackoff";
 
 export default class LoginManager {
   static setExternalId(identityOSModel: OSModel<SupportedIdentity>, externalId: string): void {
--- a/src/page/managers/LoginManager.ts
+++ b/src/page/managers/LoginManager.ts
@@ -172,4 +172,34 @@ export default class LoginManager {
     const newIdentity = { external_id };
     userData.identity = newIdentity;
   }
+
+  /**
+   * Transfer subscription when identifyUser fails with 409
+   * @param appId
+   * @param pushSubscriptionId
+   * @param identity
+   * @returns Promise<Partial<UserData>>
+   */
+  static async transferSubscription(appId: string, pushSubscriptionId: string, identity: SupportedIdentity):
+    Promise<Partial<UserData>> {
+      Log.info(`identifyUser failed: externalId already exists. Attempting to transfer push subscription...`);
+
+      const retainPreviousOwner = false;
+      const transferResponse = await RequestService.transferSubscription(
+        { appId },
+        pushSubscriptionId,
+        identity,
+        retainPreviousOwner
+      );
+      const transferResponseStatus = transferResponse?.status;
+      const tansferResult = transferResponse?.result;
+
+      if (transferResponseStatus && transferResponseStatus >= 200 && transferResponseStatus < 300) {
+        Log.info("transferSubscription succeeded");
+        const transferResultIdentity = tansferResult?.identity;
+        return { identity: transferResultIdentity };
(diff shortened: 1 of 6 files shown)


src/page/managers/LoginManager.ts
```diff
@@ -36,8 +37,13 @@ export default class LoginManager {
...
     if (status && status >= 200 && status < 300) {
       Log.info("Successfully created user", result);
-    } else {
-      Log.error("Error creating user", result);
+    } else if (status && status >= 400 && status < 500) {
+      Log.error("Malformed request", result);
+    } else if (status && status >= 500) {
+      // retry indefinitely
+      Log.error("Server error. Retrying...");
+      await awaitableTimeout(1000);
```
rgomezp, line 61: Probably want to set a longer await + backoff:

10 secs, 20, 30, 30, 30

Add max retry count: 5

src/page/managers/LoginManager.ts
```diff
@@ -83,6 +89,13 @@ export default class LoginManager {
       } else {
         throw new OneSignalError(`transferSubscription failed: ${JSON.stringify(tansferResult)}}`);
       }
+    } else if (identifyResponseStatus >= 400 && identifyResponseStatus < 500) {
+      throw new OneSignalError(`identifyUser: malformed request: ${JSON.stringify(identifyUserResponse?.result)}`);
+    } else if (identifyResponseStatus >= 500) {
+      // retry indefinitely
+      Log.error("identifyUser failed: server error. Retrying...");
+      await awaitableTimeout(1000);
+      return this.identifyUser(userData, pushSubscriptionId);
```
rgomezp, line 99: See above comment

rgomezp:
## Autosquash prior to merging

src/page/managers/LoginManager.ts
```diff
@@ -43,64 +51,72 @@ export default class LoginManager {
       return result;
   }
 
-  static async upsertUser(userData: Partial<UserData>): Promise<UserData> {
+  static async upsertUser(userData: Partial<UserData>, retry: number = 5): Promise<UserData> {
     logMethodCall("LoginManager.upsertUser", { userData });
+
+    if (retry === 0) {
+      throw new OneSignalError("Login: upsertUser failed: max retries reached");
```
jkasten2, line 58: Do we have a catch for this so it doesn't bubble outside of the SDK? If we do catch some where, can we ensure that it doesn't create any side effects? That is does the SDK continue to operate as expected.

jkasten2 (changes requested):

src/page/managers/LoginManager.ts
rgomezp: We do have a **[catch here](https://github.com/OneSignal/OneSignal-Website-SDK/blob/user-model/retry-login/src/onesignal/OneSignal.ts#L160)**

I just added a commit to remove the outermost throw to prevent just that

rgomezp: > create any side effects
Currently, it does create a side-effect: namely we never hydrate a `onesignal_id` so any operation dependent on that will fail.

We should consider what to do in this circumstance.

jkasten2: Catching and rethrowing there is ok for now. Longer term the SDK will need to keep this operation in cache, so the operation isn't lost though.

src/page/managers/LoginManager.ts
```diff
@@ -43,64 +51,72 @@ export default class LoginManager {
...
 
-    if (status && status >= 200 && status < 300) {
+    if (status >= 200 && status < 300) {
       Log.info("Successfully created user", result);
-    } else if (status && status >= 400 && status < 500) {
+    } else if (status >= 400 && status < 500) {
       Log.error("Malformed request", result);
-    } else if (status && status >= 500) {
-      // retry indefinitely
+    } else if (status >= 500) {
```
jkasten2, line 71: What status or result do we get on timeout? We should treat it like a 5xx so we can retry it.

jkasten2 (changes requested):

src/page/managers/LoginManager.ts
rgomezp: Typically a 408

src/page/managers/LoginManager.ts
```diff
@@ -94,7 +99,7 @@ export default class LoginManager {
 
       // identify user
       const identifyUserResponse = await RequestService.addAlias({ appId }, aliasPair, identity);
-      const identifyResponseStatus = identifyUserResponse?.status;
+      const identifyResponseStatus = 500;
```
rgomezp, line 102: Remove

PR #990, opened 2023-02-09
feat: refactor images location for static assets

niels-s:
# Description

Point to our new static assets location `media.onesignal.com/web-sdk` to retrieve static assets like, for example, our bell images.

Depending on https://github.com/OneSignal/media.onesignal.com/pull/5

## Details

# Systems Affected
   - [x] WebSDK
   - [x] Backend
   - [ ] Dashboard

# Validation
### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [ ] No Typescript warnings

Other:

### Checklist

---

## Related Tickets

---

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Website-SDK/990)

### DIFF
--- a/src/bell/Dialog.ts
+++ b/src/bell/Dialog.ts
@@ -90,7 +90,7 @@ export default class Dialog extends AnimatedElement {
 
         let instructionsHtml = '';
         if (imageUrl) {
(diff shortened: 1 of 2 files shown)


rgomezp (approved):

jkasten2 (commented):
LGTM

Issue #992, opened 2023-02-18
[Feedback]: "Web Push Troubleshooting" documentation appears out-of-date


onethirtyfive:
### What's on your mind?

My friend, a OneSignal infrastructure eng, was kind enough to refer me for a product engineering role this past week. I spent the evening working on a conversation piece demo I hope to have ready if and when. (Yes, I'm shameless. :))

At one point, I had made an easily-overlooked TS misconfiguration that broke my integration. While fixing, I ran into ["Debugging not receiving chrome notifications"](https://documentation.onesignal.com/docs/troubleshooting-web-push#debugging-not-receiving-chrome-notifications) in your "Web Push Troubleshooting" doc.

Following step 4 of this section, I opened `chrome://serviceworker-internals`, clicking "Inspect" on my app's service worker. In the resulting popout devtools console, when I started to enter`OneSignal.log.setLevel('trace')`, I saw:

<img width="369" alt="image" src="https://user-images.githubusercontent.com/4127/219852342-041f118e-5b55-4a0f-9be0-5c1c8045dc5b.png">

Curious, I cloned the repo and went digging. Looks like a parallel logging implementation solely for `ServiceWorker` was introduced by https://github.com/OneSignal/OneSignal-Website-SDK/commit/2550ae5eb4d9818faefdb8e824bdee5459b86cf3.

After familiarizing myself with SDK internals, I tried this in my app's console and it worked:

`OneSignal.context.workerMessenger.directPostMessageToSW('os.set_sw_logging', { shouldLog: true })`

That section remains useful, imho: great exposure to Chrome's service worker tooling. Thanks for writing it.

rgomezp:
Thanks for the issue.

So to confirm, you're saying that the normal `setLogLevel` function is not also setting it on the SW side?

onethirtyfive:
Right!

Two scenarios for you. Before each, I cleared site data and stopped/unregistered service worker.

Scenario 1: only `OneSignal.log.setLevel('trace')`:

![image](https://user-images.githubusercontent.com/4127/220797513-11896275-51ef-428b-9129-2689835be264.png)

Scenario 2, with a _stunning_ "mask" effect courtesy of `Preview.app`:

1. `OneSignal.log.setLevel('trace')`, then
2. `OneSignal.context.workerMessenger.directPostMessageToSW('os.set_sw_logging', { shouldLog: true })`:

[...]

Issue #980, opened 2023-01-10
🚧 OneSignal WebSDK v16 Released (Beta 8)! 🚧 -- seeking feedback


rgomezp:
### What's on your mind?

In this major version beta release for the OneSignal SDK, we are making a significant shift from a device-centered model to a user-centered model. A user-centered model allows for more powerful omni-channel integrations within the OneSignal platform.

For information please see the [migration guide](https://github.com/OneSignal/OneSignal-Website-SDK/blob/user-model/v1/MIGRATION_GUIDE.md).

**🚧 Beta 2 release. Please test thoroughly prior to production use. 🚧**

Please post any feedback related to the beta under this issue.

rgomezp:
@netizen-ais , 
Great question! Currently, no. But this feature is coming soon.

However, instead it will take a JWT token.

lmeurs:
Hi all, thanks for v11, the new user model is exactly what we need for our client. We experienced some issues, your support deskrefered to this Github issue.

We are new to OneSignal, so we had to start from scratch. This did not work out for v11 due to the lack of updated documentation, so we decided to build a proof of concept based on v9. We wanted to explore all possibilities, so implemented all events and methods. This worked like a charm, then we tried migrating to v11 and ran into an issue.

1. At https://noti.bloen.nl/ is a working example using our v11 script from https://noti.bloen.nl/js/notifications.js. Method OneSignal.Notifications.getPermissionStatus() seemed to no longer work, so I implemented testing property OneSignal.Notifications.permissionNative. This value seems not updated right away, I had to implement a small timeout to make it work, but there must be a better way! 😀 Could not find a way to use a promise of callback, did I miss something? The same script without timeout can be found at https://noti.bloen.nl/without-timeout/. NB: this only 

2. Also in the final part of the script I call OneSignal.login() which results in error: “Login: Error while identifying/upserting user: identifyUser failed: no identity found”, which did not occur in the v9 version calling OneSignal.setExternalUserId().

3. Users appear to be able to alter their external ID's client side. I assume using JWT tokens mentioned by @rgomezp can help prevent this, is this documented somewhere?

All help is well appreciated! 😀

BTW: When do you expect a stable v11 version? The 15th of March Marc said “we expect it within weeks rather than months”. 😀

rgomezp:
[...]

PR #999, opened 2023-03-04
feat: OneSignal.setSWLogging; improved testability

onethirtyfive:
# Description

## One Line Summary

Adds inability to turn on/off SW logging from page without directly using internals (#992).

## Details

This PR:

- has minimal impact on internal API contracts
- introduces [`SWLog`](https://github.com/onethirtyfive/OneSignal-Website-SDK/blob/sync-page-and-sw-logger-enabling/src/libraries/SWLog.ts), previously sharing the name of Page `Log` but not its implementation. `SWLog`:
    - follows the singleton pattern used elsewhere in repo
    - has good tests
    - `SWLog` is now unambiguous vs Page `Log` use--lower cognitive overhead
- Added `OneSignal.setSWLogging(enabled: boolean)`:
    - small refactor extracts `WorkerMessengerCommandHandlers` from `ServiceWorker` for testing surface area, used in this PR
    - uses existing `workerMessenger` API to send a `SetLogging` message to SW
    - `workerMessenger` now logs whether message receipt enables or disables `SWLog`

Because Service Workers are persistent, SW logging is unaffected by page refresh. This isn't new, but it behaves like `setLevel` (which is persisted in `Database`). Happy coincidence. :)

Associated [documentation](https://documentation.onesignal.com/docs/troubleshooting-web-push#debugging-not-receiving-chrome-notifications) would need updating.

# Systems Affected
   - [x] WebSDK
   - [ ] Backend
   - [ ] Dashboard

# Validation

See screenshots.

## Tests

- thorough unit testing of `SWLog` now possible with singleton pattern
- small refactor enabled good `SetLogging` message handling tests

### Checklist

**Programming Checklist**
Interfaces:
    **note: I did leave it in place on `SWLog`**
   **please confirm interfaces are in right place**

Functions:

Typescript:
   - [x] No Typescript warnings

Other:

## Screenshots

Note that `workerMessenger`, which I this is shared between page and sw browser contexts, uses the Page log and always logs. The Service Worker logs more, pictured below, only if activated.

Disabling:

image

Enabling:

image

### Checklist

---

## Related Tickets

#992 

---

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Website-SDK/999)


rgomezp:
Hey Joshua, thanks for opening this PR!

So unfortunately we're not planning on making any API changes to this SDK. In part this is due to the new [SDK currently in beta.](https://github.com/OneSignal/OneSignal-Website-SDK/issues/980)

That said, we also wouldn't add this function to the new SDK since the ideal behavior would more simply be that the one existing logging function `setLogLevel` sets it on both the page and sw contexts.

We really do appreciate your work here, and if you would like to use your changes, you always can! Simply build the SDK and host the SDK files yourself. You can follow the instructions [in this README](https://github.com/OneSignal/OneSignal-Website-SDK/tree/main/express_webpack).

Lastly, if you haven't taken a look at the new beta, I highly encourage that you do so that you can start the migration to the new user-based model, which will allow you to make the most of OneSignal's omni-channel offering to target users from multiple channels.

onethirtyfive:
No problem, it was fun learning about the SDK's internals, which was how I
discovered the issue.

I'll take a look at that beta.

Thanks!

Issue #560, opened 2019-10-23
Chrome Desktop with Dev iOS View - TypeError: Cannot read property 'pushNotification' of undefined


watzak:
**Description:**

There is an js error when u are visiting following url = https://www.watson.ch/wirtschaft/schweiz/930659620-moebel-pfister-von-xxxlutz-gekauft-schweizer-tradition-geht-nach-oesterreich
in mobile view

TypeError: Cannot read property 'pushNotification' of undefined
    at Function.getSafariNotificationPermission (PermissionManager.js:83)
    at Je.<anonymous> (PermissionManager.js:69)
    at Generator.next (<anonymous>)
    at SubscriptionError.js:3
    at new Promise (<anonymous>)
    at Ye (SubscriptionError.js:3)
    at Je.getReportedNotificationPermission (PermissionManager.js:67)
    at Je.<anonymous> (PermissionManager.js:39)
    at Generator.next (<anonymous>)
    at SubscriptionError.js:3

**Steps to Reproduce Issue:**

1. https://www.watson.ch/wirtschaft/schweiz/930659620-moebel-pfister-von-xxxlutz-gekauft-schweizer-tradition-geht-nach-oesterreich
1. on mobile or with user-agent mobile on chrome developers mode
1. check console output and u will find a js error
![](https://user-images.githubusercontent.com/8505970/67390129-4d2cdd80-f59c-11e9-9fae-433e60d5ae98.png)

**Anything else:**

Thanx for your help in advance ;)

jkasten2:
@watzak Thanks for reporting. Selecting an iOS device (iPhone or iPad) in the Chrome tools ends up mocking the `navigator.userAgent` string to an iOS device, however it does not mock `window.safari` so this error occurs in that testing environment. On a real iOS device in either the Chrome or Safari browser this error does not occur.

This is an error we will be fixing, in the meantime you can use "responsive" or an Android device in the drop down to simulate a non-iOS device to keep it from causing any issues in your mobile testing from a desktop browser.

svem1:
This error occurs in Google Chrome on Windows 10 too (site - https://mygoal.one, need autorize)

InitHelper.js:289 TypeError: Cannot read property 'pushNotification' of undefined
    at Function.getSafariNotificationPermission (PermissionManager.js:83)
    at Re.<anonymous> (PermissionManager.js:69)
    at Generator.next (<anonymous>)
    at tslib.es6.js:65
[...]

PR #1002, opened 2023-03-09
Feat/safari vapid

jkasten2:
# Description
## One Line Summary
Add support for Safari browsers that only support the VAPID PushAPI (iOS), Legacy Safari push will continue to be preferred where it is available (macOS).

## Details
# Validation
## Tests
macOS 13.1 with Safari 16.2
* Subscribed with Safari Legacy APIs and it receives notifications
* Tested with `window.safari = undefined` to simulate an iOS WebApp env, ensured it used VAPID push and received pushes
* Tested with `delete PushSubscriptionOptions` to simulate Safari on pre-macOS13, ensured it subscribed and received notifications

iOS 16.4 beta 3
* SDK didn't prompt until it was A2HS, as expected.
* Notification prompting worked and notification was received.

macOS 12.6.3 with Safari 15.6.1
* Prompted and notification received

Windows 11 22H2 Chrome 110.0.5481.105
* Prompted and notification received 

### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [X] No Typescript warnings

Other:

### Checklist

---

## Related Tickets
Due to Safari on macOS 13+ bugs we are holding off on PR #930, which attempted to use VAPID by default and migrate existing Legacy Safari players to it.

This also resolves issue ["Chrome Desktop with Dev iOS View - TypeError: Cannot read property 'pushNotification' of undefined" #560](https://github.com/OneSignal/OneSignal-Website-SDK/issues/560)

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Website-SDK/1002)

### DIFF
--- a/express_webpack/index.html
+++ b/express_webpack/index.html
@@ -3,8 +3,12 @@
 <!-- Some sites use this HTML base tag to change the origin for all relative links.
     Ensure OneSignal SDK correctly handles this. -->
 <base href="https://www.example.com/">
-<script src="sdks/Dev-OneSignalSDK.js" async=""></script>
+<script src="sdks/Dev-OneSignalSDK.js" defer=""></script>
 <script>
+    // NOTE: Uncomment and open site in Safari on macOS 13+ to simulate
+    // a similar JS API as an iOS 16.4 WebApp.
+    // window.safari = undefined;
+
     const SERVICE_WORKER_PATH = "push/onesignal/";
 
     function getUrlQueryParam(name) {
--- a/src/Environment.ts
+++ b/src/Environment.ts
@@ -1,6 +1,7 @@
 import SdkEnvironment from './managers/SdkEnvironment';
 import { WindowEnvironmentKind } from './models/WindowEnvironmentKind';
 import bowser from 'bowser';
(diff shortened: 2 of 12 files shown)


express_webpack/index.html
```diff
@@ -3,8 +3,13 @@
 <!-- Some sites use this HTML base tag to change the origin for all relative links.
     Ensure OneSignal SDK correctly handles this. -->
 <base href="https://www.example.com/">
-<script src="https://localhost:4001/sdks/Dev-OneSignalSDK.js" async=""></script>
+<script src="https://localhost:4001/sdks/Dev-OneSignalSDK.js" defer=""></script>
 <script>
+    // NOTE: Uncomment and open site in Safari on macOS 13+ to simulate
+    // a similar JS API as an iOS 16.4 WebApp.
+    // TEMP: Re-comment out
```
rgomezp, line 10: did you forget to uncomment?

rgomezp (approved):
Just one comment, otherwise LGTM

express_webpack/index.html
jkasten2: fixed, I removed this TEMP line

PR #1003, opened 2023-03-09
[Sandbox] Add web app manifest for ios push testing (for main branch)

jkasten2:
# Description
## One Line Summary
Required changes to our Sandbox localhost testing to try out iOS WebPush.

## Details

# Validation
## Tests
Tested on iPadOS 16.4 beta 2 (a) with these steps:
1. Open site in an HTTPS URL
2. Press the Share button > "Add to HomeScreen"
3. Open new homescreen web app
4. Ensured notification prompting and receiving worked.
### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [X] No Typescript warnings

Other:

### Checklist
      - No visual changes

---

## Related Tickets
Is the same as PR #997 but for the main branch.

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Website-SDK/1003)

### DIFF
--- a/express_webpack/index.html
+++ b/express_webpack/index.html
@@ -1,5 +1,8 @@
 <!DOCTYPE html>
 <html lang="en">
+<!-- This Web Application Manifest is required for iOS 16.4 WebPush -->
+<link rel="manifest" href="manifest.json" />
+
 <!-- Some sites use this HTML base tag to change the origin for all relative links.
     Ensure OneSignal SDK correctly handles this. -->
 <base href="https://www.example.com/">
--- a/express_webpack/index.html
+++ b/express_webpack/index.html
@@ -5,7 +5,8 @@
 
 <!-- Some sites use this HTML base tag to change the origin for all relative links.
     Ensure OneSignal SDK correctly handles this. -->
-<base href="https://www.example.com/">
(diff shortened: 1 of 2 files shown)


express_webpack/index.html
```diff
@@ -1,9 +1,15 @@
 <!DOCTYPE html>
 <html lang="en">
+<!-- This Web Application Manifest is required for iOS 16.4 WebPush -->
+<link rel="manifest" href="manifest.json" />
+
 <!-- Some sites use this HTML base tag to change the origin for all relative links.
     Ensure OneSignal SDK correctly handles this. -->
-<base href="https://www.example.com/">
+<!-- NOTE: This does not work with the relative pathed manifest.json above. -->
+<!--<base href="https://www.example.com/"> -->
```
rgomezp, line 9: Can we just remove the example.com base tag entirely? I always comment it out anyway since it prints a warning

rgomezp (approved):
Just one comment

express_webpack/index.html
jkasten2: It is commented out by default now in the PR, so think we should keep the comment so there is some awareness as it should be tested in some scenarios.

PR #1009, opened 2023-03-31
Disable confirmed delivery for Safari

jkasten2:
# Description
## 1 Line Summary
Disable Confirmed Deliveries for Safari as this causes iOS push endpoints to become invalid due Apple's restriction on background execution.

### Background
Safari supports two different push implements
* Safari Legacy - Safari's priority implementation, this does not use a service worker, pushes are displayed directly be Safari.
    - Code changes in this PR have no effect on this code path.
    - This implementation is available for macOS only
* Safari VAPID - This was introduced with macOS 13 and iOS 16.4 and follows the standard PushAPI which uses a service worker.
    - This code path is effected.
* Preference - We prefer Safari Legacy where available (all macOS version), and only use Safari VAPID where it is the only option (only iOS).
    - We currently continue to prefer Safari Legacy even through Safari VAPID is available on macOS 13 due to an unrelated bug.
### Implementation
Simply check if `bowser.safari` is `true` and return early to skip the Confirmed Delivery logic.

#### 1. Feature Detection
It is best practise for web development to use feature detection over browser detection however there isn't a native browser feature or limit API to detect here.
#### 2. Use navigator.platform
I originally attempted to use `navigator.platform` as I confirmed we get nice "iPad" and "iPhone" values from iOS devices, discovered by using this [example page](https://onesignal-without-sdk-test.glitch.me) which outputed these values:
* iPadOS 16.4 - navigator.platform == 'iPad"
* iPhone 6s w/ iOS 15.7.4- navigator.platform == 'iPhone"

And for macOS Safari we get "MacIntel", however we also get "MacIntel" in Chrome on macOS so this killed this idea.

#### 3. Push Endpoint Domain
Apple's push endpoints always use the format `*.push.apple.com` which seems like a good choice to know if the browser is Safari. We can get this endpoint in the browser context however it is not something you can get in the ServiceWorker context, the `PushManager` does not exist there.

We might be able to use this idea long term, if we put the browser type in the query parameter in the ServiceWorker URL when registering the ServiceWorker. However it probably isn't worth the complexly, would only be worth this was the only thing depending on the bowser library.

# Validation
### Confirm Issue
* iPadOS 16.4  - confirmed after sending 4 to 7 notifications or so they device would stop receiving them.
* macOS 13.2.1 - Was not able to reproduce issue.

### Testing After Fix
* iPadOS 16.4 - Send about 30 notifications and made sure each one was displayed. Also confirmed that Confirmed Deliveries did not show for a few notifications.
* Chrome 110 on macOS 13.21 - Confirmed that Confirmed Deliveries still works.

### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [x] No Typescript warnings

Other:

---

## Related Tickets

---

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Website-SDK/1009)

### DIFF
--- a/src/service-worker/ServiceWorker.ts
+++ b/src/service-worker/ServiceWorker.ts
@@ -345,6 +345,9 @@ export class ServiceWorker {
     if (!notification)
       return null;
 
+    if (!ServiceWorker.browserSupportsConfirmedDelivery())
+      return null;
+
     // Received receipts enabled?
     if (notification.rr !== "y")
       return null;
@@ -373,6 +376,19 @@ export class ServiceWorker {
 
     await awaitableTimeout(Math.floor(Math.random() * MAX_CONFIRMED_DELIVERY_DELAY * 1_000));
     return await OneSignalApiBase.put(`notifications/${notification.id}/report_received`, postData);
+  }
+
+  /**
+   * Confirmed Delivery isn't supported on Safari since they are very strict about the amount
+   * of time you have to finish the onpush event. Spending to much time in the onpush event
+   * will cause the push endpoint to become revoked!, causing the device to stop receiving pushes!
+   *
+   * iPadOS 16.4 it was observed to be only about 10 secounds.
+   * macOS 13.3 didn't seem to have this restriction when testing up to a 25 secound delay, however
+   * to be safe we are disabling it for all Safari browsers.
+  */
+  static browserSupportsConfirmedDelivery(): boolean {
+    return !bowser.safari
   }
 
   /**


src/service-worker/ServiceWorker.ts
```diff
@@ -345,6 +345,9 @@ export class ServiceWorker {
     if (!notification)
       return null;
 
+    if (!ServiceWorker.browserSupportsConfirmedDelivery())
```
rgomezp, line 348: Nit: use { }

src/service-worker/ServiceWorker.ts
```diff
@@ -375,6 +378,19 @@ export class ServiceWorker {
...
   }
 
+  /**
+   * Confirmed Delivery isn't supported on Safari since they are very strict about the amount
+   * of time you have to finish the onpush event. Spending to much time in the onpush event
+   * will cause the push endpoint to become revoked!, causing the device to stop receiving pushes!
+   *
+   * iPadOS 16.4 it was observed to be only about 10 secounds.
+   * macOS 13.3 didn't seem to have this restriction when testing up to a 25 secound delay, however
+   * to be safe we are disabling it for all Safari browsers.
```
rgomezp, line 388: This would essentially kill RR for all existing Safari subscriptions. I wonder if it's a better idea to only add the restriction to Safari on mobile.

You could use navigator.platform to detect this.

jkasten2: >This would essentially kill RR for all existing Safari subscriptions.

We have never had a release where we enabled macOS Safari VAPID push (device_type = 17) yet. For Safari Legacy push (device_type = 7) this code path would never happen, since pushes are not displayed by the service worker, but directly by Safari itself.

> I wonder if it's a better idea to only add the restriction to Safari on mobile.

I thought of that, but since Apple didn't state any different limits between Safari mobile vs desktop there is a chance Apple could start enforcing this for desktop later. It is much safer to disable it for all Safari types, there is some details released by Apple stating the differences, if it was even intended.

> You could use navigator.platform to detect this.

I did try this, I put my notes in the "Other Options Considered" section of this PR.

rgomezp (approved):

src/service-worker/ServiceWorker.ts
jkasten2: Left for now, since the `if` above and below does not have `{}`. Preferring consistency in the function over following the standard.

If this PR had more changes I would consider updating all the `if`s.

PR #1029, opened 2023-05-04
User model/fix notification types

rgomezp:
# 1-line Summary:
Fallback to `lastKnownPushToken` for push subscription, optimize `notification_types` setting, and add test coverage.

# Overview:
This PR improves the handling of push subscriptions by falling back to the `lastKnownPushToken` when getting the current push subscription model. It also optimizes the setting of `notification_types` using the `isOptedIn` property, removing the need to get the subscription state again. Additionally, test coverage has been added to ensure the push subscription models are updated correctly when the native permission changes.

# Details:
- Fallback to `lastKnownPushToken` when getting current push subscription model to handle cases when the function fires after revoking the native permission
- Use `isOptedIn` to set `notification_types`, removing the need to get the subscription state again
- Remove unused vars (nits)
- Remove deviceId (player id) usage under user model (nits)
- Update `TestEnvironmentHelper` with more OneSignal globals needed for testing

# Systems Affected
   - [x] WebSDK
   - [ ] Backend
   - [ ] Dashboard

# Validation
## Tests
- Add test coverage to ensure push subscription models are updated correctly when the native permission changes

### Info
New tests were tested without the changes and failed (good).

Had to also update the TestEnvironment since it was missing static properties like `EVENTS` and `NotificationsNamespace` needed for this test.

### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [x] No Typescript warnings

Other:

### Checklist

---

## Related Tickets

---

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Website-SDK/1029)

### DIFF
--- a/src/shared/managers/SubscriptionManager.ts
+++ b/src/shared/managers/SubscriptionManager.ts
@@ -34,8 +34,8 @@ import OneSignalError from "../errors/OneSignalError";
 import { SessionOrigin } from "../models/Session";
 import { executeCallback, logMethodCall } from "../utils/utils";
 import UserDirector from "../../onesignal/UserDirector";
-import { OSModel } from "src/core/modelRepo/OSModel";
-import { isCompleteSubscriptionObject } from "src/core/utils/typePredicates";
+import { OSModel } from "../../core/modelRepo/OSModel";
+import { isCompleteSubscriptionObject } from "../../core/utils/typePredicates";
 
 export interface SubscriptionManagerConfig {
   safariWebId?: string;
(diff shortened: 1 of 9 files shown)


src/shared/managers/SubscriptionManager.ts
```diff
@@ -834,7 +833,6 @@ export class SubscriptionManager {
      */
 
     const isPushEnabled = !!(
```
jkasten2, line 835: Removing the playerId check does change the definition of this, I won't say `PushSubscriptionState.subscribed ` is an accurate name for this then. `Eligible` would be a better wording, if we really don't want to change this playerId check to a subscriptionId check.

We should also check how this was used to see if this change will cause any side effects.

jkasten2 (changes requested):

jkasten2 (approved):

PR #1036, opened 2023-05-10
[user model] Bugfix: jwtToken argument to login function is optional, shouldn't throw

rgomezp:
Fix bug causing throw if `jwtToken` is omitted

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Website-SDK/1036)

### DIFF
--- a/src/onesignal/OneSignal.ts
+++ b/src/onesignal/OneSignal.ts
@@ -105,8 +105,8 @@ export default class OneSignal {
       throw new InvalidArgumentError('externalId', InvalidArgumentReason.WrongType);
     }
 
-    if (typeof jwtToken !== 'string') {
-      throw new InvalidArgumentError('token', InvalidArgumentReason.WrongType);
+    if (jwtToken !== undefined && typeof jwtToken !== 'string') {
+      throw new InvalidArgumentError('jwtToken', InvalidArgumentReason.WrongType);
     }
 
     LoginManager.login(externalId, jwtToken);


src/onesignal/OneSignal.ts
```diff
@@ -105,7 +105,7 @@ export default class OneSignal {
       throw new InvalidArgumentError('externalId', InvalidArgumentReason.WrongType);
     }
 
-    if (typeof jwtToken !== 'string') {
+    if (!!jwtToken && typeof jwtToken !== 'string') {
       throw new InvalidArgumentError('token', InvalidArgumentReason.WrongType);
```
jkasten2, line 109: This is just a bit out of scope for the PR but I just noticed this error field name isn't correct. It should be `'jwtToken'` instead of `'token'` so it is more clear.

jkasten2 (changes requested):

src/onesignal/OneSignal.ts
```diff
@@ -105,7 +105,7 @@ export default class OneSignal {
       throw new InvalidArgumentError('externalId', InvalidArgumentReason.WrongType);
     }
 
-    if (typeof jwtToken !== 'string') {
+    if (!!jwtToken && typeof jwtToken !== 'string') {
```
jkasten2, line 108: Thinking we should check `undefined` here instead of using `!!`, because `false` is not going to throw the error when it should.

jkasten2 (changes requested):

PR #1040, opened 2023-05-19
User model/upgrade bowser

rgomezp:
# Description
## 1 Line Summary
Upgrade `bowser` for improved testability.

## Overview
The following changes have been made:

1. Cleaned dead code.
2. Replaced the `bowser` dependency with `bowser-castle`, which uses a newer version of `bowser` while retaining the previous API.
3. Updated all instances of `bowser` with the `bowserCastle()` function.
4. Upgraded the `intl-tel-input` (nit) and `bowser` libraries.
5. Set the `userAgent` in `TestEnvironment` to configure the DOM correctly for tests.

# Systems Affected
   - [x] WebSDK
   - [ ] Backend
   - [ ] Dashboard

# Validation
### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [x] No Typescript warnings

Other:

### Checklist

---

## Related Tickets

---

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Website-SDK/1040)

### DIFF
--- a/__test__/support/environment/TestEnvironment.ts
+++ b/__test__/support/environment/TestEnvironment.ts
@@ -2,7 +2,7 @@ import SdkEnvironment from "../../../src/shared/managers/SdkEnvironment";
 import { AppUserConfig, ConfigIntegrationKind, ServerAppConfig } from "../../../src/shared/models/AppConfig";
 import { TestEnvironmentKind } from "../../../src/shared/models/TestEnvironmentKind";
 import BrowserUserAgent from "../models/BrowserUserAgent";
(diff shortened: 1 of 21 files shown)


jkasten2 (changes requested):
Few questions this PR doesn't answer that I have concerns on:
1. Does the new library work correctly in the ServiceWorker context?
2. What manual testing has been done?
3. Given this PR adds adds a new package and updates an existing one, how does this effect the size of the gziped SDK?

PR #1050, opened 2023-07-07
[Fix] iOS denied notificationPermissionChange

jkasten2:
# Description
## 1 Line Summary
Fixes the `OneSignal.on('notificationPermissionChange', function() {})` event not firing on iOS WebApps.

## Details
We were relying on the browser API `navigator.permissions.query({ name: 'notifications' }).onchange` to call our `triggerNotificationPermissionChanged` which fires the OneSignal `notificationPermissionChange` event, however on iOS WebApps this does not fire due to a bug with the browser.
To account for this we are calling our `PermissionUtils.triggerNotificationPermissionChanged` event every time the native notification prompt is answered or dismissed.

This however does create duplicate events so this commit adds logic to our `PermissionUtils.triggerNotificationPermissionChanged` to prevent this.

# Validation
## Tests
Tested event now fire exactly once on iOS 16.5.
* Tested both for accepting and declining.

Also confirmed Chrome on macOS still works
### Checklist
      - Tests no longer run on CI due to image no longer having Python 2.

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [X] No Typescript warnings

Other:

### Checklist
      - No visual changes

---

## Related Tickets

---

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Website-SDK/1050)

### DIFF
--- a/src/managers/SubscriptionManager.ts
+++ b/src/managers/SubscriptionManager.ts
@@ -334,15 +334,12 @@ export class SubscriptionManager {
       const permission = await SubscriptionManager.requestPresubscribeNotificationPermission();
 
       /*
-        Notification permission changes are already broadcast by the page's
-        notificationpermissionchange handler. This means that allowing or
-        denying the permission prompt will cause double events. However, the
-        native event handler does not broadcast an event for dismissing the
+        The native event handler does not broadcast an event for dismissing the
         prompt, because going from "default" permissions to "default"
         permissions isn't a change. We specifically broadcast "default" to "default" changes.
        */
-      if (permission === NotificationPermission.Default)
-        await PermissionUtils.triggerNotificationPermissionChanged(true);
+      const forcePermissionChangeEvent = permission === NotificationPermission.Default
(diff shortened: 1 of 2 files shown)


rgomezp:
In the details of this PR, your one line summary is different than the details section.

> Fixes the OneSignal.on('notificationPermissionChange', function() {}) event not firing on iOS WebApps.

> [calling triggerNotificationPermissionChanged] however does create duplicate events so this commit adds logic to our PermissionUtils.triggerNotificationPermissionChanged to prevent this.

Looking at the changes, I can see how this prevents duplicate events, but how does this fix the permissionChange not firing on iOS Web Apps?

If you could update the PR details to explain this?

Other than that LGTM

rgomezp (changes requested):
Clarify PR details

jkasten2:
@rgomezp I updated the PR details explaining the connection between the bug that was fixed and how it was fixed.

rgomezp (approved):

PR #1058, opened 2023-07-20
[User model] [Fix] User login before subscription create

jkasten2:
# Description
## 1 Line Summary
If `OneSignal.login` is called before notifications are accepted no longer error and cache value to create the User when possible.

## Details

# Validation
Tested on Chrome on macOS
## Tests
Added new test reproducing issue, confirming the test failed and passed afterwards.
### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [X] No Typescript warnings

Other:

### Checklist

---

## Related Tickets

---

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Website-SDK/1058)


emawby (approved):
LGTM!

PR #1061, opened 2023-07-26
[User Model] [Fix] Erroring on User property updates before push subscription create

jkasten2:
# Description
## 1 Line Summary
Fixes an error where any User property update before a push subscribed is crated causes the SDK to throw.

## Details

# Validation
## Tests
* Added new automated test
* Manually tested by calling `OneSignal.User.addTag("key", "value");` then accepting notifications, ensuring the User create call included the tags.
### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [X] No Typescript warnings

Other:

### Checklist

---

## Related Tickets

---

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Website-SDK/1061)

### DIFF
--- a/src/core/executors/ExecutorResult.ts
+++ b/src/core/executors/ExecutorResult.ts
@@ -1,28 +1,28 @@
-export default interface ExecutorResult<Model> {
+export interface ExecutorResult<Model> {
   readonly success: boolean,
   readonly retriable: boolean,
   readonly result?: Model
 }
 
 export class ExecutorResultSuccess<Model> implements ExecutorResult<Model> {
-  success = true;
-  retriable = false;
+  readonly success = true;
+  readonly retriable = false;
   constructor(readonly result?: Model) {
 
   }
 }
 
 export class ExecutorResultFailRetriable<Model> implements ExecutorResult<Model> {
-  success = false;
-  retriable = true;
+  readonly success = false;
+  readonly retriable = true;
   constructor(readonly result?: Model) {
 
   }
 }
 
 export class ExecutorResultFailNotRetriable<Model> implements ExecutorResult<Model> {
-  success = false;
-  retriable = false;
+  readonly success = false;
+  readonly retriable = false;
   constructor(readonly result?: Model) {
 
   }
--- a/src/core/executors/ExecutorResult.ts
+++ b/src/core/executors/ExecutorResult.ts
@@ -1,7 +1,29 @@
-export default class ExecutorResult<Model> {
-  constructor(
-    readonly success: boolean,
-    readonly retriable: boolean,
-    readonly result?: Model
-    ){}
+export default interface ExecutorResult<Model> {
+  readonly success: boolean,
+  readonly retriable: boolean,
+  readonly result?: Model
+}
+
+export class ExecutorResultSuccess<Model> implements ExecutorResult<Model> {
+  success = true;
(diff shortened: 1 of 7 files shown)


src/core/executors/ExecutorResult.ts
```diff
@@ -1,7 +1,29 @@
-export default class ExecutorResult<Model> {
-  constructor(
-    readonly success: boolean,
-    readonly retriable: boolean,
-    readonly result?: Model
-    ){}
+export default interface ExecutorResult<Model> {
```
rgomezp, line 1: no longer need to export

src/core/requestService/IdentityRequests.ts
```diff
@@ -1,7 +1,11 @@
 import OneSignalApiBaseResponse from "../../shared/api/OneSignalApiBaseResponse";
 import { logMethodCall } from "../../shared/utils/utils";
 import OneSignalError from "../../shared/errors/OneSignalError";
-import ExecutorResult from "../executors/ExecutorResult";
+import ExecutorResult, {
```
rgomezp, line 4: unneeded import

src/core/requestService/SubscriptionRequests.ts
```diff
@@ -1,7 +1,11 @@
 import MainHelper from "../../shared/helpers/MainHelper";
 import OneSignalError from "../../shared/errors/OneSignalError";
 import { logMethodCall } from "../../shared/utils/utils";
-import ExecutorResult from "../executors/ExecutorResult";
+import ExecutorResult, {
```
rgomezp, line 4: unneeded import

src/core/requestService/UserPropertyRequests.ts
```diff
@@ -1,6 +1,10 @@
 import { logMethodCall } from "../../shared/utils/utils";
 import OneSignalError from "../../shared/errors/OneSignalError";
-import ExecutorResult from "../executors/ExecutorResult";
+import ExecutorResult, {
```
rgomezp, line 3: unneeded import

rgomezp (changes requested):

src/core/requestService/IdentityRequests.ts
jkasten2: It's still being used as below as as type see this line:
```typescript
static async addIdentity<Model>(operation: Operation<Model>): Promise<ExecutorResult<IdentityModel>> {
```

src/core/executors/ExecutorResult.ts
jkasten2: This still needs to be exported as it is still a result type used outside of this file.
But it should probably don't be `default`, so I can fix that.

jkasten2: Removed `default` here

src/core/requestService/SubscriptionRequests.ts
jkasten2: Import is need in this file, but just removed the 'default' import here.

src/core/requestService/UserPropertyRequests.ts
jkasten2: Import is need in this file, but just removed the 'default' import here.

rgomezp (approved):

PR #1062, opened 2023-07-26
[User model] [Fix] macOS Safari prompting and subscribing

jkasten2:
# Description
## 1 Line Summary
Fix macOS Safari error "Push notification prompting can only be done from a user gesture." when prompting for notification permission and registering for push.

## Details
* Safari's legacy prompting API (`window.safari.pushNotification.requestPermission()`) doesn't allow awaiting for any I/O which was causing the prompting error. Commits:
    - b724a625d0ee764b12fbe75787614fedb899208d
    - 04c97aed1731552a7e7a12dc73f97c73179bc1f8
* After prompting was fixed Safari wasn't registering for push due to a few changes that didn't come over correctly from PR #1002. Commits:
   - 03d59892ee6c098052eef2252ddd0a6dc489af5b

# Validation
## Tests
* Tested with Safari 16.4 on macOS 13.3.1, ensured prompting worked and a subscription record was created on OneSignal's server.
* Tested end-to-end with iOS 16.5, ensuring notifications display as well.
   - Unsure if this was working before this PR, since it works different from macOS Safari
* Tested refreshing the page to ensure no behavior changes with session counts. (referring to changes in commit b724a625d0ee764b12fbe75787614fedb899208d)
* Automated tests were not added since it isn't feasible to cover all I/O Promises that could happen in the chain. Additionally Legacy Safari will be deprecated in the future.
### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [X] No Typescript warnings

Other:

### Checklist

---

## Related Tickets

---

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Website-SDK/1062)

### DIFF
--- a/src/shared/helpers/SubscriptionHelper.ts
+++ b/src/shared/helpers/SubscriptionHelper.ts
@@ -11,40 +11,18 @@ import Log from '../libraries/Log';
 import { ContextSWInterface } from '../models/ContextSW';
 import SdkEnvironment from '../managers/SdkEnvironment';
 import { EnvironmentInfo } from '../../page/models/EnvironmentInfo';
-import { SessionOrigin } from '../models/Session';
 import { Browser } from '../models/Browser';
-import { PushDeviceRecord } from '../models/PushDeviceRecord';
 import { PermissionUtils } from '../utils/PermissionUtils';
-import MainHelper from './MainHelper';
 
 export default class SubscriptionHelper {
   public static async registerForPush(): Promise<Subscription | null> {
-    const isPushEnabled = await OneSignal.context.subscriptionManager.isPushNotificationsEnabled();
-    return await SubscriptionHelper.internalRegisterForPush(isPushEnabled);
+    return await SubscriptionHelper.internalRegisterForPush();
   }
 
-  public static async internalRegisterForPush(isPushEnabled: boolean): Promise<Subscription | null> {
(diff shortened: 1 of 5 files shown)


src/shared/helpers/SubscriptionHelper.ts
```diff
@@ -11,40 +11,18 @@ import Log from '../libraries/Log';
...
 
-  public static async internalRegisterForPush(isPushEnabled: boolean): Promise<Subscription | null> {
+  public static async internalRegisterForPush(): Promise<Subscription | null> {
     const context: ContextSWInterface = OneSignal.context;
     let subscription: Subscription | null = null;
 
-    /*
-      Within the same page navigation (the same session), do not register for
-      push if the user is already subscribed, otherwise the user will have its
-      session count incremented on each page refresh. However, if the user is
```
rgomezp, line 33: Did you confirm the session count is not incrementing even after the refresh? What is the impact on other browsers? From first glance it doesn't appear this code was Safari-specific.

rgomezp (approved):
Conditionally approved

src/shared/helpers/SubscriptionHelper.ts
jkasten2: I updated the "Tests" section in this PR to provide more details on this.

PR #1068, opened 2023-07-27
User model/ios badges

rgomezp:
# Description
## Details

# Systems Affected
   - [ ] WebSDK
   - [ ] Backend
   - [ ] Dashboard

# Validation
### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [ ] No Typescript warnings

Other:

### Checklist

---

## Related Tickets

---

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Website-SDK/1068)

### DIFF
--- a/src/shared/models/OSNotificationDataPayload.ts
+++ b/src/shared/models/OSNotificationDataPayload.ts
@@ -24,9 +24,13 @@
   icon?: string;
   image?: string;
   tag?: string;
+  // URL of the image used to represent the notification when there is
+  // not enough space to display the notification itself
+  // (https://developer.mozilla.org/en-US/docs/Web/API/Notification/badge)
   badge?: string;
   vibrate?: VibratePattern;
   buttons?: NotificationButtonData[];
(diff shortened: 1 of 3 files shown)


src/shared/helpers/EventHelper.ts
```diff
@@ -14,6 +14,13 @@ import { OSModel } from '../../core/modelRepo/OSModel';
 import { SubscriptionModel } from '../../core/models/SubscriptionModels';
 
 export default class EventHelper {
+  // Sets the app badge count: https://developer.mozilla.org/en-US/docs/Web/API/Badging_API
+  static setAppBadgeCount(notification: OSNotificationDataPayload): void {
```
jkasten2, line 18: This runs on the page, which means the badge would only get set if the end-user has a page open to the site. Shouldn't we read the value from the notification payload in the ServiceWorker and update it there? `navigator` is available in the Service Worker, is `setAppBadge` not?

PR #1073, opened 2023-08-01
[User Model] [Fix] Bell not updating to subscribed state

jkasten2:
# Description
## 1 Line Summary
Fix bug where the bell would not update it's visual state when becomes subscribed.

## Details

# Validation
## Tests
Test with Chrome on macOS, ensured bell grayed out and decreased in size after the push subscribed we was created.
### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [X] No Typescript warnings

Other:

### Checklist

---

## Related Tickets

---

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Website-SDK/1073)


rgomezp (approved):
Nice fix!

PR #1075, opened 2023-08-01
[User Model] [Fix] Notification permission change event type to boolean

jkasten2:
# Description
## 1 Line Summary
Correct `OneSignal.Notification.addEventListener("permissionChange", function....)` to provide a `boolean` instead of a `string`.

## Details
# Validation
## Tests
* Tested on Chrome on macOS, ensuring the event fires with a boolean
* Added new automated tests
### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [X] No Typescript warnings

Other:

### Checklist

---

## Related Tickets

---

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Website-SDK/1075)


__test__/support/managers/PermissionManager.ts
```diff
@@ -0,0 +1,13 @@
...
+export class PermissionManager {
+    public static async mockNotificationPermissionChange(
+        test: jest.It,
+        nativePermission: NotificationPermission,
+      ): Promise<void> {
+        test.stub(RealPermissionManager.prototype, 'getPermissionStatus', nativePermission);
+        // mimick native permission change
+        await PermissionUtils.triggerNotificationPermissionChanged();
+      }
+}
```
rgomezp, line 13: newline

src/shared/utils/PermissionUtils.ts
```diff
@@ -35,5 +35,14 @@ export class PermissionUtils {
 
     await Database.put('Options', { key: 'notificationPermission', value: newPermission });
     OneSignalEvent.trigger(OneSignal.EVENTS.NOTIFICATION_PERMISSION_CHANGED_AS_STRING, newPermission);
+    
+    const newPermissionBoolean = newPermission === 'granted';
+    const previousPermissionBoolean = previousPermission === 'granted';
+
+    const shouldBeUpdatedBoolean = newPermissionBoolean !== previousPermissionBoolean || updateIfIdentical;
```
rgomezp, line 42: can shorten to `shouldBeUpdated`. Also maybe add a comment to the function explaining what the argument `updateIfIdentical` does for quick reference.

rgomezp (changes requested):
Overall looks great. A couple nits

__test__/support/managers/PermissionManager.ts
jkasten2: Fixed

src/shared/utils/PermissionUtils.ts
jkasten2: Fixed the naming on all these and split out the boolean even into it' own function to make things more clear.

rgomezp (approved):

PR #1076, opened 2023-08-02
[User Model] [Fix] Wrong notifications type when disabling browser permission

jkasten2:
# Description
## 1 Line Summary
Fix reported notifications type when disabling browser permission, from -2 (api disabled) to 0 (no notification permissions).

## Details

# Validation
Tested on Chrome on macOS, disabling permission after accepting and observing the network call.
## Tests
Updated unit tests looking for the specific notification type.
### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [X] No Typescript warnings

Other:

### Checklist

---

## Related Tickets

---

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Website-SDK/1076)

### DIFF
--- a/__test__/unit/pushSubscription/nativePermissionChange.test.ts
+++ b/__test__/unit/pushSubscription/nativePermissionChange.test.ts
@@ -29,7 +29,7 @@ describe('Notification Types are set correctly on subscription change', () => {
     await EventHelper.checkAndTriggerSubscriptionChanged();
 
     // check that the set function was called at least once with the correct value
-    expect(osModelSetSpy).toHaveBeenCalledWith('notification_types', SubscriptionStateKind.Default);
+    expect(osModelSetSpy).toHaveBeenCalledWith('notification_types', SubscriptionStateKind.NoNativePermission);
     expect(osModelSetSpy).toHaveBeenCalledWith('enabled', false);
   });
 
--- a/__test__/unit/pushSubscription/nativePermissionChange.test.ts
+++ b/__test__/unit/pushSubscription/nativePermissionChange.test.ts
@@ -6,6 +6,7 @@ import { DUMMY_PUSH_TOKEN } from "../../support/constants";
 import { initializeWithPermission } from "../../support/helpers/pushSubscription";
 import { PermissionManager } from "../../support/managers/PermissionManager";
 import { NotificationPermission } from "../../../src/shared/models/NotificationPermission";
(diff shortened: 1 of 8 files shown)


emawby (approved):

__test__/unit/pushSubscription/nativePermissionChange.test.ts
```diff
@@ -18,7 +19,7 @@ describe('Notification Types are set correctly on subscription change', () => {
     jest.resetAllMocks();
   });
 
-  test('When native permission is rejected, we update notification_types to -2 & enabled to false', async () => {
+  test('When native permission is rejected, we update notification_types to 0 & enabled to false', async () => {
```
rgomezp, line 22: when the native permission is rejected, the notification types being 0 is equivalent to there not being a permission yet. I still think this should be -2

src/shared/managers/SubscriptionManager.ts
```diff
@@ -188,20 +186,35 @@ export class SubscriptionManager {
...
-  async updatePushSubscriptionNotificationTypes(notificationTypes: number): Promise<void> {
+  async updateNotificationTypes(): Promise<void> {
+    const notificationTypes = await this.getNotificationTypes()
+    await this.updatePushSubscriptionNotificationTypes(notificationTypes);
+  }
+
+  async getNotificationTypes(): Promise<SubscriptionStateKind> {
+    const { optedOut } = await Database.getSubscription();
+    if (optedOut) {
+      return SubscriptionStateKind.MutedByApi;
```
rgomezp, line 197: Maybe we should think about renaming this since the notification type could be set from the SDK. Could we change it to 'Granted', 'Denied', and 'Default'?

jkasten2: Renamed this as as well as the `SubscriptionStateKind`, `Default`, so this isn't misleading. Also added some comments in SubscriptionStateKind to further clarify in commit eb4f784912fedf3bab80e5c2a6547ec4646df204.

__test__/unit/pushSubscription/nativePermissionChange.test.ts
jkasten2: Talked about this outside of this thread, concussion was the names of the enums were misleading. A follow up commit was made to rename them to address this.

rgomezp (approved):

PR #1077, opened 2023-08-03
[User Model] [Fix] Update existing subscription on token changes instead of creating new

jkasten2:
# Description
## 1 Line Summary
Fix bug where new push subscriptions were being created instead of updating the existing one.

## Details
There are two key changes that were needed to fix this bug:
1. `getCurrentPushSubscriptionModel()`  - wasn't correctly using `lastKnownPushToken`, fixed in commit e9134eec3fb7c9f12f7520b6718e95740738bf53
2. `checkAndTriggerSubscriptionChanged()` - wasn't checking for token changes, so the backend was never updated, fixed in commit https://github.com/OneSignal/OneSignal-Website-SDK/commit/8e4235abd1abe5358dea4636f89c617e2e73074a
3. Other changes are small refactors and commit a4f312908741777638663096323b5219e4bde8ea is clearing up some related dead code.

# Validation
Tested on macOS with Chrome, with a clean browser. Then toggled notification permission off and on, refreshed page and repeated a few times in different combination. Observed notification_types and token network calls were made as expected without errors.
### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [X] No Typescript warnings

Other:

### Checklist

---

## Related Tickets

---

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Website-SDK/1077)

### DIFF
--- a/src/core/CoreModuleDirector.ts
+++ b/src/core/CoreModuleDirector.ts
@@ -160,25 +160,39 @@ export class CoreModuleDirector {
     return modelStores.pushSubscriptions.models as { [key: string]: OSModel<SupportedSubscription> };
   }
 
+  private async getPushSubscriptionModelByCurrentToken()
+    :Promise<OSModel<SupportedSubscription> | undefined> {
+      logMethodCall("CoreModuleDirector.getPushSubscriptionModelByCurrentToken");
+      const pushToken = await MainHelper.getCurrentPushToken();
+      if (pushToken) {
+        return this.getSubscriptionOfTypeWithToken(ModelName.PushSubscriptions, pushToken);
+      }
+      return undefined;
+  }
+
+  // Browser may return a different PushToken value, use the last-known value as a fallback.
+  //   - This happens if you disable notification permissions then re-enable them.
+  private async getPushSubscriptionModelByLastKnownToken()
+    :Promise<OSModel<SupportedSubscription> | undefined> {
+    logMethodCall("CoreModuleDirector.getPushSubscriptionModelByLastKnownToken");
+    const { lastKnownPushToken } = await Database.getAppState();
+    if (lastKnownPushToken) {
+      return this.getSubscriptionOfTypeWithToken(ModelName.PushSubscriptions, lastKnownPushToken);
+    }
+    return undefined;
+  }
+
   /**
    * Gets the current push subscription model for the current browser.
    * @returns The push subscription model for the current browser, or undefined if no push subscription exists.
    */
-  // TO DO: make synchronous by making getting the current push token synchronous
-  public async getCurrentPushSubscriptionModel(): Promise<OSModel<SupportedSubscription> | undefined> {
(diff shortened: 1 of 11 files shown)


src/core/CoreModuleDirector.ts
```diff
@@ -160,25 +160,39 @@ export class CoreModuleDirector {
...
+    return undefined;
+  }
+
   /**
    * Gets the current push subscription model for the current browser.
    * @returns The push subscription model for the current browser, or undefined if no push subscription exists.
    */
-  // TO DO: make synchronous by making getting the current push token synchronous
-  public async getCurrentPushSubscriptionModel(): Promise<OSModel<SupportedSubscription> | undefined> {
+  public async getPushSubscriptionModel()
```
rgomezp, line 189: Did you think about ways to make getting the push token synchronous? This has the potential to make tons of dependent functions synchronous.

src/shared/managers/SubscriptionManager.ts
```diff
@@ -163,7 +163,8 @@ export class SubscriptionManager {
   }
 
   private async _updatePushSubscriptionModelWithRawSubscription(rawPushSubscription: RawPushSubscription) {
-    const pushModel = await OneSignal.coreDirector.getCurrentPushSubscriptionModel();
+    // This undefined when it shouldn't be
```
rgomezp, line 166: Can you rephrase the comment?

src/core/CoreModuleDirector.ts
```diff
@@ -160,25 +160,39 @@ export class CoreModuleDirector {
...
-
-      if (!pushToken) {
-        Log.debug("No push token found, returning undefined");
-        return undefined;
-      }
-    }
-    return this.getSubscriptionOfTypeWithToken(ModelName.PushSubscriptions, pushToken);
+    return (
+      await this.getPushSubscriptionModelByCurrentToken() ||
+      await this.getPushSubscriptionModelByLastKnownToken()
```
rgomezp, line 194: I think we should add some tests for these changes

rgomezp (changes requested):

src/core/CoreModuleDirector.ts
jkasten2: The native browse API to get the current token `registration.pushManager.getSubscription()` returns a `Promise` so it isn't possible.
* If cached it and use a cache value that could work around that, but then it won't be current and it has a risk of being out of date then.

src/shared/managers/SubscriptionManager.ts
jkasten2: ah ops, ya that problem is solved now, as this is the reason for this PR 😄

src/core/CoreModuleDirector.ts
jkasten2: Added tests

emawby (approved):

PR #1078, opened 2023-08-08
[User Model] Fix notification event types; click, willDisplay, and dismiss 

jkasten2:
# Description
## 1 Line Summary
Fix public SDK API field names and types for the notification event types; click, willDisplay, and dismiss

### Scope of changes summary
The changes are related to OSNotification and the places where it is used, in changing the fields it was discovered there is a lot of tight coupling with other things such as indexDb and webhooks. While they may seem unrelated changes to those were needed and refactored to a point to decouple them for the future.

### OSNotification
This wasn't using the same field names as [Android](https://github.com/OneSignal/OneSignal-Android-SDK/blob/5.0.0-beta4/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/notifications/INotification.kt) and [iOS](https://github.com/OneSignal/OneSignal-iOS-SDK/blob/5.0.0-beta-06/iOS_SDK/OneSignalSDK/OneSignalCore/Source/OSNotification.h) so corrected it to match and renamed it to an interface `IOSNotification`.

### Notification Events:
* click - This was using `NotificationClickResult` instead of `NotificationClickEvent`
* dismiss - This didn't' have it's own event, introduced `NotificationDismissEvent`
* foregroundWillDisplay - This had its own type already, but wasn't formatted correctly which was missed due to missing types

### indexDb
Since `IOSNotification` changed as well as the format for events we introduced [`OSNotificationDatabaseSerializer.ts`](https://github.com/OneSignal/OneSignal-Website-SDK/compare/user-model/v1...user-model/fix-notification-event-types?expand=1#diff-bb8b80d58d68d73716a92fd710e876ee60dc2e786c2735625c0db72ce0c74b56) to convert the format to and from the database. This was done so no migration is needed, and also to decouple and future proof any changes to these classes.
* This includes the the click handling, as well as clicks and receives for outcomes.

### Webhooks
Since we added type where they were missing, `executeWebhooks` needed to be refactored as it was account for 3 different types of payloads; click, display, and dismiss. This was refactored into 3 different payload classes, again to decouple and future  proof field changes.

# Validation
## Tests
Test on Chrome on macOS. Ensuring notifications are displayed and click handling directs to the correct URL.
### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [X] No Typescript warnings

Other:

### Checklist

---

## Related Tickets

---

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Website-SDK/1078)


src/shared/helpers/OSNotificationDatabaseSerializer.ts
```diff
@@ -0,0 +1,117 @@
...
+}
+
+interface NotificationButtonsClickForOpenHandlingSchema {
+  readonly action: string;
+  readonly title: string;
+  readonly icon?: string;
+  readonly url?: string;
+}
+
+export class NotificationClickForOpenHandlingSerializer {
```
jkasten2, line 43: add comment here to check the top of the file for more details.
* This way you see hint in VS code / yoru IDE.

jkasten2, line 43: Look into adding a test that checks to ensure this is being used to serialize data in and out of indexDb, checking that nothing is doing it directly.

src/sw/webhooks/OSWebhookSender.ts
```diff
@@ -0,0 +1,33 @@
+import Database from "../../shared/services/Database";
+import Log from "../libraries/Log";
+import { IOSWebhookEventPayload } from "./IOSWebhookEventPayload";
+
+export class OSWebhookSender {
+  async send(payload: IOSWebhookEventPayload): Promise<void> {
+    const webhookTargetUrl = await Database.get<string>('Options', `webhooks.${payload.event}`);
+    if (!webhookTargetUrl)
```
rgomezp, line 8: brackets

rgomezp (approved):

PR #1079, opened 2023-08-10
[User Model] [Fix] Notification Open & Confirm Delivery REST API calls failing

jkasten2:
# Description
## 1 Line Summary
Notification Open & Confirm Delivery REST API calls were failing due to placeholder player_id being used.

## Details
`CoreModuleDirector` was introduced in v16, and is they place to access the `pushSubscriptionId` we needed. However `CoreModuleDirector` does not work in the ServiceWorker context since it uses JS API that are not available. This includes `localStorage` (possibly switch to `indexDB` in the future) as well as access to `window` and possibly others. Give the scope of such changes would large and has unknowns to address `CoreModuleDirector` we added a `ModelCacheDirectAccess` class to directly access what we needed, on the trade off it duplicates some code.

# Validation
## Tests
Tested on Chrome on Windows 11. Ensure both notification open and confirm deliveries show on the dashboard.
### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [X] No Typescript warnings

Other:

### Checklist

---

## Related Tickets

---

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Website-SDK/1079)

### DIFF
--- a/src/sw/helpers/ModelCacheDirectAccess.ts
+++ b/src/sw/helpers/ModelCacheDirectAccess.ts
@@ -1,0 +1,20 @@
+import EncodedModel from "../../core/caching/EncodedModel";
+import { ModelName } from "../../core/models/SupportedModels";
+import Database from "../../shared/services/Database";
+
+/**
+ * WARNING: This is a temp workaround for the ServiceWorker context only!
+ * PURPOSE: CoreModuleDirector doesn't work in the SW context.
+ * TODO: This is duplicated logic tech debt to address later
+ */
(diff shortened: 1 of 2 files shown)


emawby:
Do we know why CoreModuleDirector doesn't work in the SW context?

jkasten2:
> Do we know why CoreModuleDirector doesn't work in the SW context?

That is a good question, I updated my PR details with an answer to this.

PR #1103, opened 2023-08-25
[CI] Enforce linting

jkasten2:
# Description
## One Line Summary
Enable enforcement of eslint and prettier rules, as well as made the code changes required them to pass.

## Details
Recommend reviewing commit-by-commit.
The following two commands were run on the source.
* `yarn prettier --write src test __test__`
* `yarn eslint src --ext .js,.jsx,.ts,.tsx --fix`

Since not all rules had auto fixes, some changes were done manually to meet the rules. Also a few in-line suppresses were added to since it was to risky to change the code at this time, but preferred over omitting the rule as the rule should still be applied to new code.

# Validation
Tested on Windows 11 22H2 with Chrome 115.0.5790.171 ensuring it can subscribe, receive a push, register an email, and open a new tab on notification click.
### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [X] No Typescript warnings

Other:

### Checklist

---

## Related Tickets

---

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Website-SDK/1103)


emawby (approved):
Looks good the tests are passing now in the PR!

src/shared/libraries/Emitter.ts
```diff
@@ -41,10 +41,12 @@ export default class Emitter {
    * execute only once.
    */
   public once(event: string, listener: EventHandler): Emitter {
+    // eslint-disable-next-line @typescript-eslint/no-this-alias
     const that = this;
 
-    function fn(this: Function) {
+    function fn(this: unknown) {
```
rgomezp, line 45: There are 18 new `unknown` types in this commit. I think we can fix at least some of these such as here

rgomezp (changes requested):

src/shared/libraries/Emitter.ts
jkasten2: This one is fixable, it should be `EventHandler`, fixed it in commit 543f21a59cf5b472a9f9535cc2f3cbeef23beea4.

Most of the other ones are an improvement though still, example `Function` was turninged into `(param: unknown) => void` which is more accurate.
* `Function` is a banned type with the [default eslint rules](https://typescript-eslint.io/rules/ban-types/)

A few of the `unknowns` could probably be replaced with generics, however that will require adding new types which seems out of scope for this PR.

src/shared/models/UpdatePlayerOptions.ts
```diff
@@ -11,7 +11,7 @@ export interface UpdatePlayerOptions {
     sdk                         ?: string;
     device_model                ?: string;
     app_id                      ?: string;
-    tags                        ?: Object | null;
+    tags                        ?: Record<string, unknown> | null;
```
rgomezp: what about this record, we know this should be a string, correct?

src/shared/api/OneSignalApi.ts
```diff
@@ -21,15 +21,15 @@ export default class OneSignalApi {
     return OneSignalApiShared.sendNotification(appId, playerIds, titles, contents, url, icon, data, buttons);
   }
 
-  static jsonpLib(url: string, fn: Function) {
+  static jsonpLib(url: string, fn: (err: unknown, data: ServerAppConfig) => void) {
```
rgomezp: could this `err` be of type `Error`?

src/shared/api/OneSignalApi.ts
```diff
@@ -21,15 +21,15 @@ export default class OneSignalApi {
...
     JSONP(url, null, fn);
   }
 
   static async downloadServerAppConfig(appId: string): Promise<ServerAppConfig> {
     if (SdkEnvironment.getWindowEnv() !== WindowEnvironmentKind.ServiceWorker) {
       return await new Promise<ServerAppConfig>((resolve, reject) => {
         // Due to CloudFlare's algorithms, the .js extension is required for proper caching. Don't remove it!
-        OneSignalApi.jsonpLib(`${SdkEnvironment.getOneSignalApiUrl().toString()}/sync/${appId}/web`,(err, data) => {
+        OneSignalApi.jsonpLib(`${SdkEnvironment.getOneSignalApiUrl().toString()}/sync/${appId}/web`,(err: unknown, data: ServerAppConfig) => {
```
rgomezp: see above

rgomezp (changes requested):

src/shared/models/UpdatePlayerOptions.ts
jkasten2: ah yes, good catch! Fixed now

src/shared/api/OneSignalApi.ts
jkasten2: ah yes, good catch! Fixed now

src/shared/api/OneSignalApi.ts
jkasten2: ah yes, good catch! Fixed now

rgomezp (approved):

PR #1132, opened 2023-11-14
[Fix] notification click event not firing if it opens a new tab

jkasten2:
# Description
## 1 Line Summary
Fix notification click event not firing if it opens a new tab.

## Details
The internal `fireStoredNotificationClicks()` function was not called to fire any pending click events when `OneSignal.Notifications.addEventListener('click', function() {})` is called.

# Validation
## Tests
Tested on Chrome 119 on Window 11, testing both "focus" and "navigate" options for notification open behavior.
* Click behavior set on dashboard "Origin: Take actions on a previous tab open to same domain"
* Click behavior set on dashboard "Exact: (default) Opens a new window"

### Checklist

**Programming Checklist**
Interfaces:

Functions:
   - [X} Helpers should not access any data but rather be given the data to operate on.

Typescript:
   - [X] No Typescript warnings

Other:

## Screenshots
Tested click Behavior Exact:
image

Tested click behavior Focus:
image

### Checklist

---

## Related Tickets

---

### DIFF
--- a/express_webpack/index.html
+++ b/express_webpack/index.html
@@ -124,8 +124,8 @@
     });
 
     OneSignalDeferred.push(function(onesignal) {
-      onesignal.Notifications.addEventListener('willDisplay', function (event) {
-        showEventAlert('willDisplay', { event });
+      onesignal.Notifications.addEventListener('foregroundWillDisplay', function (event) {
+        showEventAlert('foregroundWillDisplay', event);
       });
     });
 
--- a/express_webpack/index.html
+++ b/express_webpack/index.html
@@ -149,6 +149,7 @@
     }
 
     function showEventAlert(eventName, payload) {
+      console.log(`OneSignal event ${eventName} fired!`, payload);
       if (!showEventAlertToggleSetting) {
         return;
       }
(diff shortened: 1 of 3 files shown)


emawby (approved):

__test__/unit/notifications/eventListeners.test.ts
```diff
@@ -0,0 +1,22 @@
...
+
+  afterEach(() => {
+    jest.resetAllMocks();
+  });
+
+  test('Adding click listener fires internal EventHelper', async () => {
+    const stub = test.stub(
+      EventHelper,
+      'fireStoredNotificationClicks',
+      null,
```
rgomezp, line 17: This is an optional argument. And I don't believe this function returns anything

rgomezp (approved):
Just 1 nit and fix linting errors. Other than that, looks good.

__test__/unit/notifications/eventListeners.test.ts
jkasten2: ah good catch, removed the `null`.

PR #1136, opened 2023-11-15
[Fix] Service Worker race condition, when a 2nd sw is in play, causing User not to register

jkasten2:
# Description
## 1 Line Summary
Fix Service Worker race condition, when a 2nd sw is in play, causing User not to register.

## Details
If you attempt to get the service worker registration immediately after registering it the browser can give you a different registration than expected. This is due to the fact the new registration may not be ready yet and there might be sw in scope at a higher level.

The problem noted above cause the OneSignal SDK to register for a push token on the wrong service worker. This in turn causes the SDK to attempt to create a user with a push subscription but without a push token, which results in a `400` error on `POST /users`.

To fix, instead of calling `navigator.serviceWorker.getRegistration` we are now simply using the `ServiceWorkerRegistration` give to us by `navigator.serviceWorker.register` to avoid this race condition completely

# Validation
## Tests
Test on Chrome 119 on Windows 11
* Ensured after accepting push notifications I received a welcome notification. Clear the browser data and push permission, opened a new tab, and tested again 3 more times.
* Refreshed page after accepting notifications to ensure existing service worker is found without an error.
### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [X] No Typescript warnings

Other:

### Checklist

---

## Related Tickets

---

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Website-SDK/1136)

### DIFF
--- a/src/shared/managers/ServiceWorkerManager.ts
+++ b/src/shared/managers/ServiceWorkerManager.ts
@@ -174,6 +174,8 @@ export class ServiceWorkerManager {
     });
   }
 
+  // Returns false if the OneSignal service worker can't be installed
+  // or is already installed and doesn't need updating.
   private async shouldInstallWorker(): Promise<boolean> {
     // 1. Does the browser support ServiceWorkers?
     if (!Environment.supportsServiceWorkers()) return false;
@@ -486,9 +488,11 @@ export class ServiceWorkerManager {
    * additional query parameters, but this must then stay consistent.
    */
 
-  public async installWorker() {
+  public async installWorker(): Promise<
+    ServiceWorkerRegistration | undefined | null
+  > {
     if (!(await this.shouldInstallWorker())) {
-      return;
+      return this.getRegistration();
     }
 
     Log.info('Installing worker...');
@@ -512,8 +516,12 @@ export class ServiceWorkerManager {
     Log.info(
       `[Service Worker Installation] Installing service worker ${workerHref} ${scope}.`,
     );
+
+    let registration: ServiceWorkerRegistration;
     try {
-      await navigator.serviceWorker.register(workerHref, { scope });
+      registration = await navigator.serviceWorker.register(workerHref, {
+        scope,
+      });
     } catch (error) {
       Log.error(
         `[Service Worker Installation] Installing service worker failed ${error}`,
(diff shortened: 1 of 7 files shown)


src/shared/managers/ServiceWorkerManager.ts
```diff
@@ -174,6 +174,8 @@ export class ServiceWorkerManager {
     });
   }
 
+  // Returns true if a the OneSignal service worker can't be installed
```
rgomezp, line 177: "Returns true if a" - grammar

Also, this logic doesn't match the function. `shouldInstallWorker` should return `true` if it can and should be installed.

src/sw/helpers/ServiceWorkerUtilHelper.ts
```diff
@@ -33,4 +33,29 @@ export default class ServiceWorkerUtilHelper {
     }
     return availableWorker;
   }
+
+  // Allows waiting for the service worker registration to become active.
+  // Some APIs, like registration.pushManager.subscribe, required it be active
+  // otherwise it throws.
+  static waitForServiceWorkerToActive(
```
rgomezp, line 40: Optional nit: `waitForActiveServiceWorker` or `waitForActiveSW`

src/sw/helpers/ServiceWorkerUtilHelper.ts
```diff
@@ -33,4 +33,29 @@ export default class ServiceWorkerUtilHelper {
...
+    registration: ServiceWorkerRegistration,
+  ): Promise<void> {
+    return new Promise((resolver) => {
```
rgomezp, line 43: Nit: typically this is just `resolve`

rgomezp (approved):
Great fixes. Logic looks good. Just some nits & comment changes

src/sw/helpers/ServiceWorkerUtilHelper.ts
jkasten2: ah ya, name is confusing, i think `waitUntilActive` is a better name.
* This function is a static function so the caller code will be `ServiceWorkerUtilHelper.waitUntilActive(registration)`

src/sw/helpers/ServiceWorkerUtilHelper.ts
jkasten2: hmm, prettier wants `(resolver)` instead of `resolver` here. I get this error when I try to change it to your suggestion:
```
 src/sw/helpers/ServiceWorkerUtilHelper.ts:43:24
  ⚠  43:24  Replace resolver with (resolver)  prettier/prettier
```

src/shared/managers/ServiceWorkerManager.ts
jkasten2: Good catch on both, fixed

src/sw/helpers/ServiceWorkerUtilHelper.ts
rgomezp: I meant `resolve` over `resolver`

jkasten2: ah yes, my bad🤦. fixed

PR #1138, opened 2023-11-16
use new URL in `subscribeSafariPromptPermission`

fedterzi:
# Description
## 1 Line Summary
use new URL with `appId` in `subscribeSafariPromptPermission`

## Details
Due to changes in our API, the `appId` is now required in the URL used by `subscribeSafariPromptPermission`. In instances where the new URL is not usable, we fallback to the legacy URL with an additional request. This ensures compatibility with both the updated and legacy API endpoints.

# Systems Affected
   - [x] WebSDK
   - [x] Backend
   - [ ] Dashboard

# Validation
### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [ ] No Typescript warnings

Other:

### Checklist

---

## Related Tickets

---

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Website-SDK/1138)

### DIFF
--- a/src/shared/managers/SubscriptionManager.ts
+++ b/src/shared/managers/SubscriptionManager.ts
@@ -66,6 +66,7 @@ export type SubscriptionStateServiceWorkerNotIntalled =
 export class SubscriptionManager {
   private context: ContextSWInterface;
   private config: SubscriptionManagerConfig;
+  private safariPermissionPromptFailed = false;
 
   constructor(context: ContextSWInterface, config: SubscriptionManagerConfig) {
     this.context = context;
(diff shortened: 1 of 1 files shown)


src/shared/managers/SubscriptionManager.ts
```diff
@@ -372,20 +372,34 @@ export class SubscriptionManager {
...
+          this.config.safariWebId,
+          { app_id: this.config.appId },
+          (response) => {
+            if ((response as any).deviceToken) {
+              resolve((response as any).deviceToken.toLowerCase());
+            } else {
+              // If there's no token and the new URL was used,
+              // retry with the legacy URL
+              if (url.includes('/safari/apps/')) {
+                requestPermission(
```
jkasten2, line 388: This recursive call makes the code really hard to follow. Could we simply call `await requestPermission(...)`, check if we get a value, then call `await requestPermission(...)` again with the fallback URL?

jkasten2 (changes requested):

src/shared/managers/SubscriptionManager.ts
fedterzi: Done.

jkasten2 (approved):

jkasten2 (changes requested):
I tested this and unfortunately the fallback call to `window.safari.pushNotification.requestPermission` doesn't work with Safari. The browser throws the following error "Push notification prompting can only be done from a user gesture.". It seems even though there is an `await` the browser doesn't consider it the same event.

jkasten2 (approved):
Confirmed this works as expected in Safari 17.1 on macOS 13.6.1. First click does nothing, then the 2nd click results in the native Safari notification showing correctly. Once the Safari zips are regenerated server side then the first click will start working again.

PR #1140, opened 2023-11-17
[Feat] Add getTags to public API

shepherd-l:
# Description
## 1 Line Summary
Adds the `getTags` method to the public API

## Details
Calling the public API `getTags` method returns an empty dictionary `{}` when the user property model tags field is undefined

# Systems Affected
   - [x] WebSDK
   - [ ] Backend
   - [ ] Dashboard

# Validation
### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [x] No Typescript warnings

Other:

### Checklist

---

## Related Tickets

---

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Website-SDK/1140)

### DIFF
--- a/src/onesignal/UserNamespace.ts
+++ b/src/onesignal/UserNamespace.ts
@@ -72,6 +72,6 @@ export default class UserNamespace {
   }
 
   public getTags(): { [key: string]: string } {
-    return this._currentUser?.getTags() ? this._currentUser.getTags() : {};
+    return this._currentUser?.getTags() || {};
   }
 }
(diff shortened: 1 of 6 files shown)


src/onesignal/UserNamespace.ts
```diff
@@ -70,4 +70,8 @@ export default class UserNamespace {
   public removeTags(keys: string[]): void {
     this._currentUser?.removeTags(keys);
   }
+
+  public getTags(): { [key: string]: string } {
+    return this._currentUser?.getTags() ? this._currentUser.getTags() : {};
```
jkasten2, line 75: We can shorten this by using `||` instead of the ternary operator here.

jkasten2 (approved):

PR #1170, opened 2024-04-17
[Fix] Replace deprecated CSS 'push-button'. Adds type='button' to Custom Link Prompt.

marclucraft:
# Description
- Fixes warning in console: `[Deprecation] The keyword 'push-button' specified to an 'appearance' property is not standardized. It will be removed in the future.`
- Adds `type="button"` to the Custom Link Button. This prevents this button submitting if included in a form.

## Details
Both long standing bugs.

# Systems Affected
   - [x] WebSDK
   - [ ] Backend
   - [ ] Dashboard

# Validation
### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [x] No Typescript warnings

Other:

## Screenshots
image

image

### Checklist

---

## Related Tickets

---

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Website-SDK/1170)

### DIFF
--- a/src/page/stylesheets/reset.scss
+++ b/src/page/stylesheets/reset.scss
@@ -490,7 +490,7 @@
   font-family: arial, helvetica, sans-serif;
   font-size: small;
   background: white;
-  appearance: button; /* [Deprecation] The keyword 'push-button' specified to an 'appearance' property is not standardized. It will be removed in the future. */
+  appearance: button;
   color: buttontext;
   border: 1px #a6a6a6 solid;
   background: lightgrey; /* Old browsers */
--- a/src/page/stylesheets/reset.scss
+++ b/src/page/stylesheets/reset.scss
@@ -490,7 +490,7 @@
   font-family: arial, helvetica, sans-serif;
   font-size: small;
   background: white;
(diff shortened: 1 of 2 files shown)


src/page/stylesheets/reset.scss
```diff
@@ -490,7 +490,7 @@
   font-family: arial, helvetica, sans-serif;
   font-size: small;
   background: white;
-  -webkit-appearance: push-button;
+  appearance: button; /* [Deprecation] The keyword 'push-button' specified to an 'appearance' property is not standardized. It will be removed in the future. */
```
jkasten2, line 493: We don't need this comment if we are no longer using `push-button` now.

jkasten2 (changes requested):
* Let's remove the comment about `push-button` since we are not using it anymore.
* Change PR title to what is being addressed
   - Details in the PR can include that they are long standing bugs.

src/page/stylesheets/reset.scss
marclucraft: Done and Done!

jkasten2 (approved):

PR #1179, opened 2024-06-06
Operation repo refactor

shepherd-l:
# Description
## 1 Line Summary
An operation repo refactor that will help with the implementation of identity verification

## Details
- Adds a delta queue to the operation repo to flush all executor deltas at the same time. Now the operation repo will be in charge of enqueuing deltas and passing the delta to its correct executor. Previously, each executor was in charge of enqueuing and flushing their own delta queue.
  - Makes it easier to implement pausing on the operation repo for Jwt when receiving an invalid token.
  - This change also aligns the Web SDK more with the iOS SDK's operation repo implementation.
- Groups push, sms, and email subscriptions under 1 executor. Previously push, sms, and email each had their own subscription executor.
  - Omits creating multiple subscription executors. Unnecessary now that there is only 1 flush from the operation repo that flushes all executors
  - Deprecates the `pushSubscription`, `smsSubscription`, and `emailSubscription` model names for 1 new name: `subscriptions`. This change requires a database upgrade.
- Adds external id property to operations
  - Will be used to retry operations for users with a valid Jwt token. Accessed through `Operation.model.externalId`
- Added to following tests below:

Unit Tests:
- v5 to v6 database migration for subscription records. If a user was already logged in v5, the external id is added to each subscription record in v6.
- Login test, where the external id is updated correctly for each model which will be used in deltas/operations.
  - Future work: add logout tests

# Systems Affected
   - [x] WebSDK
   - [ ] Backend
   - [ ] Dashboard

# Validation
### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [x] No Typescript warnings

Other:

### Checklist

---

## Related Tickets

---

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Website-SDK/1179)

### DIFF
--- a/__test__/support/constants.ts
+++ b/__test__/support/constants.ts
@@ -14,7 +14,7 @@
  * @constant {number} DELTA_QUEUE_TIME_ADVANCE - The time advance for the delta queue.
  */
 export const OPERATION_QUEUE_TIME_ADVANCE = 5001;
-export const DELTA_QUEUE_TIME_ADVANCE = 1001;
+export const DELTA_QUEUE_TIME_ADVANCE = 5001;
 
 /* S T R I N G   C O N S T A N T S */
 export const APP_ID = '34fcbe85-278d-4fd2-a4ec-0f80e95072c5';
--- a/__test__/support/constants.ts
+++ b/__test__/support/constants.ts
@@ -14,7 +14,7 @@
  * @constant {number} DELTA_QUEUE_TIME_ADVANCE - The time advance for the delta queue.
  */
 export const OPERATION_QUEUE_TIME_ADVANCE = 5001;
-export const DELTA_QUEUE_TIME_ADVANCE = 5001;
+export const DELTA_QUEUE_TIME_ADVANCE = 1001;
 
 /* S T R I N G   C O N S T A N T S */
 export const APP_ID = '34fcbe85-278d-4fd2-a4ec-0f80e95072c5';
--- a/__test__/support/constants.ts
+++ b/__test__/support/constants.ts
@@ -22,6 +22,7 @@ export const APP_ID = '34fcbe85-278d-4fd2-a4ec-0f80e95072c5';
 export const DUMMY_PUSH_TOKEN =
   'https://fcm.googleapis.com/fcm/send/01010101010101';
 export const DUMMY_ONESIGNAL_ID = '1111111111-2222222222-3333333333';
+export const DUMMY_ONESIGNAL_ID_2 = '2222222222-3333333333-4444444444';
 export const DUMMY_EXTERNAL_ID = 'rodrigo';
 export const DUMMY_EXTERNAL_ID_2 = 'iryna';
 export const DUMMY_SUBSCRIPTION_ID = '4444444444-5555555555-6666666666';
--- a/src/core/CoreModuleDirector.ts
+++ b/src/core/CoreModuleDirector.ts
@@ -6,6 +6,7 @@ import { SupportedIdentity } from './models/IdentityModel';
 import { ModelStoresMap } from './models/ModelStoresMap';
 import {
   SubscriptionModel,
+  SubscriptionType,
   SupportedSubscription,
 } from './models/SubscriptionModels';
 import { ModelName, SupportedModel } from './models/SupportedModels';
@@ -33,7 +34,7 @@ export class CoreModuleDirector {
     });
     // new subscription
     const pushModel = new OSModel<SupportedSubscription>(
-      ModelName.PushSubscriptions,
+      ModelName.Subscriptions,
       new FuturePushSubscriptionRecord(rawPushSubscription).serialize(),
     );
 
@@ -50,7 +51,7 @@ export class CoreModuleDirector {
 
     // don't propagate since we will be including the subscription in the user create call
     OneSignal.coreDirector.add(
-      ModelName.PushSubscriptions,
+      ModelName.Subscriptions,
       pushModel as OSModel<SupportedModel>,
       false,
     );
@@ -114,24 +115,16 @@ export class CoreModuleDirector {
 
     const modelStores = this.getModelStores();
 
-    const getModelName = (subscription: SupportedSubscription) => {
-      if (subscription.type === 'Email') {
-        return ModelName.EmailSubscriptions;
-      } else if (subscription.type === 'SMS') {
-        return ModelName.SmsSubscriptions;
-      } else {
-        return ModelName.PushSubscriptions;
-      }
-    };
-
     subscriptions.forEach(async (subscription) => {
-      const modelName = getModelName(subscription);
       /* We use the token to identify the model because the subscription ID is not set until the server responds.
        * So when we initially hydrate after init, we may already have a push model with a token, but no ID.
        * We don't want to create a new model in this case, so we use the token to identify the model.
        */
       const existingSubscription = !!subscription.token
-        ? this.getSubscriptionOfTypeWithToken(modelName, subscription.token)
+        ? this.getSubscriptionOfTypeWithToken(
+            subscription.type,
+            subscription.token,
+          )
         : undefined;
 
       if (existingSubscription) {
@@ -142,12 +135,15 @@ export class CoreModuleDirector {
         }
         existingSubscription.hydrate(subscription);
       } else {
-        const model = new OSModel<SupportedModel>(modelName, subscription);
+        const model = new OSModel<SupportedModel>(
+          ModelName.Subscriptions,
+          subscription,
+        );
         model.setOneSignalId(onesignalId);
         if (externalId) {
           model?.setExternalId(externalId);
         }
-        modelStores[modelName].add(model, false); // don't propagate to server
+        modelStores[ModelName.Subscriptions].add(model, false); // don't propagate to server
       }
     });
   }
@@ -195,13 +191,21 @@ export class CoreModuleDirector {
   } {
     logMethodCall('CoreModuleDirector.getEmailSubscriptionModels');
     const modelStores = this.getModelStores();
-    return modelStores.emailSubscriptions.models as {
+    const subscriptionModels = modelStores.subscriptions.models as {
       [key: string]: OSModel<SupportedSubscription>;
     };
+
+    const emailSubscriptions = Object.fromEntries(
+      Object.entries(subscriptionModels).filter(
+        ([, model]) => model.data?.type === SubscriptionType.Email,
+      ),
+    );
+
+    return emailSubscriptions;
   }
 
   public async hasEmail(): Promise<boolean> {
-    const emails = await this.getEmailSubscriptionModels();
+    const emails = this.getEmailSubscriptionModels();
     return Object.keys(emails).length > 0;
   }
 
@@ -210,13 +214,21 @@ export class CoreModuleDirector {
   } {
     logMethodCall('CoreModuleDirector.getSmsSubscriptionModels');
     const modelStores = this.getModelStores();
-    return modelStores.smsSubscriptions.models as {
+    const subscriptionModels = modelStores.subscriptions.models as {
       [key: string]: OSModel<SupportedSubscription>;
     };
+
+    const smsSubscriptions = Object.fromEntries(
+      Object.entries(subscriptionModels).filter(
+        ([, model]) => model.data?.type === SubscriptionType.SMS,
+      ),
+    );
+
+    return smsSubscriptions;
   }
 
   public async hasSms(): Promise<boolean> {
-    const smsModels = await this.getSmsSubscriptionModels();
+    const smsModels = this.getSmsSubscriptionModels();
     return Object.keys(smsModels).length > 0;
   }
 
@@ -228,9 +240,17 @@ export class CoreModuleDirector {
   } {
     logMethodCall('CoreModuleDirector.getAllPushSubscriptionModels');
     const modelStores = this.getModelStores();
-    return modelStores.pushSubscriptions.models as {
+    const subscriptionModels = modelStores.subscriptions.models as {
       [key: string]: OSModel<SupportedSubscription>;
     };
+
+    const pushSubscriptions = Object.fromEntries(
+      Object.entries(subscriptionModels).filter(([, model]) =>
+        this.isPushSubscriptionType(model.data?.type),
+      ),
+    );
+
+    return pushSubscriptions;
   }
 
   private async getPushSubscriptionModelByCurrentToken(): Promise<
@@ -239,9 +259,9 @@ export class CoreModuleDirector {
     logMethodCall('CoreModuleDirector.getPushSubscriptionModelByCurrentToken');
     const pushToken = await MainHelper.getCurrentPushToken();
     if (pushToken) {
-      return this.getSubscriptionOfTypeWithToken(
-        ModelName.PushSubscriptions,
-        pushToken,
+      const pushSubscriptions = this.getAllPushSubscriptionModels();
+      return Object.values(pushSubscriptions).find(
+        (subscription) => subscription.data.token === pushToken,
       );
     }
     return undefined;
@@ -257,9 +277,9 @@ export class CoreModuleDirector {
     );
     const { lastKnownPushToken } = await Database.getAppState();
     if (lastKnownPushToken) {
-      return this.getSubscriptionOfTypeWithToken(
-        ModelName.PushSubscriptions,
-        lastKnownPushToken,
+      const pushSubscriptions = this.getAllPushSubscriptionModels();
+      return Object.values(pushSubscriptions).find(
+        (subscription) => subscription.data.token === lastKnownPushToken,
       );
     }
     return undefined;
@@ -312,7 +332,7 @@ export class CoreModuleDirector {
   }
 
   public getSubscriptionOfTypeWithToken(
-    type: ModelName,
+    type: SubscriptionType,
     token: string,
   ): OSModel<SupportedSubscription> | undefined {
     logMethodCall('CoreModuleDirector.getSubscriptionOfTypeWithToken', {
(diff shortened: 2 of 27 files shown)


src/core/operationRepo/OperationRepo.ts
```diff
@@ -16,6 +18,12 @@ export class OperationRepo {
         this._processDelta(delta);
       },
     );
+
+    setInterval(() => {
+      if (this._deltaQueue.length > 0) {
```
jkasten2, line 23: Length check seems unnecessary, can we remove it?

src/core/executors/SubscriptionExecutor.ts
```diff
@@ -12,6 +12,10 @@ export class SubscriptionExecutor extends ExecutorBase {
   }
 
   processDeltaQueue(): void {
+    if (this._deltaQueue.length === 0) {
```
jkasten2, line 15: Length check seems unnecessary, can we remove it?

src/core/executors/ExecutorStore.ts
```diff
@@ -19,8 +20,21 @@ export class ExecutorStore {
 
   // call processDeltaQueue on all executors immediately
   public forceDeltaQueueProcessingOnAllExecutors(): void {
+    let didSubscriptionExecutorProcessDeltaQueue = false;
+
     Object.values(this.store).forEach((executor) => {
+      if (
+        executor instanceof SubscriptionExecutor &&
```
jkasten2, line 27: Does the `ExecutorFactory` already ensure we don't have duplicate executors? Or if it doesn't take care of it shouldn't we just omit adding it instead?

jkasten2 (changes requested):

src/core/executors/SubscriptionExecutor.ts
shepherd-l: I think we can remove it. I was copying the pattern in the other executors: [IdentityExecutor](https://github.com/OneSignal/OneSignal-Website-SDK/blob/main/src/core/executors/IdentityExecutor.ts#L16), [PropertiesExecutor](https://github.com/OneSignal/OneSignal-Website-SDK/blob/ff52c3cd26b2c90caf9c7af80883f4f5503acafe/src/core/executors/PropertiesExecutor.ts#L14). Do you think those could be removed as well?

src/core/executors/ExecutorFactory.ts
```diff
@@ -6,6 +6,8 @@ import { PropertiesExecutor } from './PropertiesExecutor';
 import { SubscriptionExecutor } from './SubscriptionExecutor';
 
 export class ExecutorFactory {
+  static subscriptionExecutor?: SubscriptionExecutor = undefined;
```
rgomezp, line 9: Nit: `= undefined` unnecessary

src/core/executors/ExecutorStore.ts
rgomezp: Under this PR, the factory will return the reference to the single static subscription executor three times and be mapped to via the three `ModelName` keys that correspond to it: `pushSubscriptions, emailSubscriptions, and smsSubscriptions`

There's probably a cleaner way to do this without hard-coding a check here

src/core/executors/SubscriptionExecutor.ts
rgomezp: Probably good to avoid doing unnecessary work

src/core/operationRepo/OperationRepo.ts
shepherd-l: The check seems necessary:
Removing the check fails the following in [`executor.test.ts`](https://github.com/OneSignal/OneSignal-Website-SDK/blob/main/__test__/unit/core/executors.test.ts):
```
✕ Subscriptions executor flushes deltas at end of `processDeltaQueue` (17 ms)
✕ Identity executor flushes deltas at end of `processDeltaQueue` (10 ms)
✕ User Properties executor flushes deltas at end of `processDeltaQueue` (9 ms)
```
Where in each `processDeltaQueueSpy` is called 2 times instead of 1

src/core/executors/ExecutorStore.ts
shepherd-l: Adding on to what rgomezp said:

The [ExecutorStore](https://github.com/OneSignal/OneSignal-Website-SDK/blob/main/src/core/executors/ExecutorStore.ts) creates an executor for each value in [ModelName](https://github.com/OneSignal/OneSignal-Website-SDK/blob/main/src/core/models/SupportedModels.ts)
And there are 3 names for subscriptions:
1.  Identity
2. Properties
3. PushSubscriptions
4. EmailSubscriptions
5. SmsSubscriptions

This Pr should ensure that there are no duplicate executors from the `ExecutorFactory`. But we could omit adding it in the first place having 1 `ModelName` for the subscription executor instead. I don't think this would heavy lift but I'm not sure.

__test__/support/constants.ts
```diff
@@ -14,7 +14,7 @@
  * @constant {number} DELTA_QUEUE_TIME_ADVANCE - The time advance for the delta queue.
  */
 export const OPERATION_QUEUE_TIME_ADVANCE = 5001;
-export const DELTA_QUEUE_TIME_ADVANCE = 1001;
+export const DELTA_QUEUE_TIME_ADVANCE = 5001;
```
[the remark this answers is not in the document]
shepherd-l: We want to fire the delta queue right after the operation queue. This is to match more closely with iOS SDK operation repo behaviour, where their delta queue is 5 seconds and then immediately turned into requests. This is also why the delta queue was moved to the operation repo in this PR.

Is there a better way to do that?/Will slowing it down cause any problems?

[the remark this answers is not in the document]
rgomezp: wouldn't we want the delta queue to fire before the operation queue? will this possibly lead to a +-10 second delay?

src/core/executors/ExecutorStore.ts
rgomezp: The `ModelName` enum is widely used, not just for the executor factory.

You would have to change the way the `ExecutorStore` indexes the executors to only have 3 keys. But this adds a point of maintenance in the future should we add new supported models.

Is there an issue arising from the differences in behavior between the WebSDK & iOS? Or is this PR aimed at bringing the SDKs to parity to address an outstanding issue?

shepherd-l: Having one flush that is controlled by the operation repo would make it easier to implement a JWT feature where we would like to pause the operation repo when we receive an invalid token. And with changing to 1 flush, I didn't think there would be a need to have 3 separate subscription executors for push, sms, email.

__test__/support/constants.ts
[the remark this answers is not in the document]
shepherd-l: Good point, I'll try testing to see if it's possible that the delta queue gets processed too late/after the operation queue.

src/core/executors/ExecutorStore.ts
shepherd-l: > But this adds a point of maintenance in the future should we add new supported models.

Could you elaborate on what maintenance would be needed?

> You would have to change the way the ExecutorStore indexes the executors to only have 3 keys. 

I thought this change would consist of changing the enum and the previous references to use 1 subscription enum. What other things would need to be done?

__test__/support/constants.ts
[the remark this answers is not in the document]
shepherd-l: I found some bugs when testing. I'll change the time back to 1 second

src/core/executors/ExecutorStore.ts
rgomezp: > Could you elaborate on what maintenance would be needed?
The executor store store interface looks like

```ts
type ExecutorStoreInterface = {
  [key in ModelName]?: OSExecutor;
};
```

If you change it to say
```ts
type ExecutorStoreInterface = {
  identity: OSExecutor;
  properties: OSExecutor;
  subscriptions: OSExecutor;
};
```

or similar you would have to update in two places should you add new model types. Not a huge deal though.

> I thought this change would consist of changing the enum and the previous references to use 1 subscription enum. What other things would need to be done?

The enum isn't just used for creating executors. It's used throughout the user-model SDK.

jkasten2: I believe it would instead be changing this code:
```ts
export enum ModelName {
  Identity = 'identity',
  Properties = 'properties',
  // TO DO: make singular
  PushSubscriptions = 'pushSubscriptions',
  EmailSubscriptions = 'emailSubscriptions',
  SmsSubscriptions = 'smsSubscriptions',
}
```
to this:
```ts
export enum ModelName {
  Identity = 'identity',
  Properties = 'properties',
  Subscriptions = 'subscriptions',
}
```

I think this will simply things and means if we add another subscription type we won't have to change this again. We will have to update code that depends on `ModelName` however, and migrate any existing cached operations to the new name though.

rgomezp: > I think this will simply things and means if we add another subscription type we won't have to change this again. We will have to update code that depends on ModelName however, and migrate any existing cached operations to the new name though.

That's a big lift. We also rely on the model name to differentiate between subscription models, like in the `CoreModuleDirector`, `OSModel`, etc...

src/core/executors/SubscriptionExecutor.ts
shepherd-l: I think we can remove it since the the next line will run `this._deltaQueue.forEach(...)` and since the delta queue is empty, it wouldn't do any work anyway. Technically I think the only extra work it would do is to create a variable but I think that is negligible.

src/core/operationRepo/OperationRepo.ts
shepherd-l: After investigating, it looks like the tests were failing due to using to using the incorrect jest timer. More details in my commit here: https://github.com/OneSignal/OneSignal-Website-SDK/pull/1179/commits/f680239c6719e06c49f9dd5b83b7c63206851174

I think the length check is unnecessary and that we should remove it

src/core/CoreModuleDirector.ts
```diff
@@ -291,27 +330,30 @@ export class CoreModuleDirector {
...
         const smsSubscriptions = this.getSmsSubscriptionModels();
         return Object.values(smsSubscriptions).find(
           (subscription) => subscription.data.token === token,
         );
       }
-      case ModelName.PushSubscriptions: {
+      case SubscriptionType.ChromePush:
+      case SubscriptionType.SafariPush:
+      case SubscriptionType.SafariLegacyPush:
+      case SubscriptionType.FirefoxPush: {
```
jkasten2, line 356: If we are making this change we should use the new `isPushSubscriptionType()` helper method you created.

src/core/CoreModuleDirector.ts
```diff
@@ -207,9 +238,17 @@ export class CoreModuleDirector {
   } {
     logMethodCall('CoreModuleDirector.getAllPushSubscriptionModels');
     const modelStores = this.getModelStores();
-    return modelStores.pushSubscriptions.models as {
+    const subscriptionModels = modelStores.subscriptions.models as {
       [key: string]: OSModel<SupportedSubscription>;
     };
+
+    let pushSubscriptions = Object.fromEntries(
```
jkasten2, line 245: This can be a `const`

src/core/CoreModuleDirector.ts
```diff
@@ -218,9 +257,9 @@ export class CoreModuleDirector {
     logMethodCall('CoreModuleDirector.getPushSubscriptionModelByCurrentToken');
     const pushToken = await MainHelper.getCurrentPushToken();
     if (pushToken) {
-      return this.getSubscriptionOfTypeWithToken(
-        ModelName.PushSubscriptions,
-        pushToken,
+      const pushSubscriptions = this.getAllPushSubscriptionModels();
+      return Object.values(pushSubscriptions).find(
+        (subscription) => subscription.data.token === pushToken,
```
jkasten2, line 262: Why not continue using `getSubscriptionOfTypeWithToken()`?

jkasten2 (changes requested):

src/core/CoreModuleDirector.ts
shepherd-l: I did this because I don't know what `SubscriptionType` the pushSubscription is. I could put a random push `SubscriptionType` to pass into `getSubscriptionOfTypeWithToken()` or make the parameter optional instead. Do you have any thoughts?

jkasten2: There are other spots in this file where we can make this same change to simplify the code. Can we update those as well?

src/core/CoreModuleDirector.ts
```diff
@@ -99,33 +115,32 @@ export class CoreModuleDirector {
...
-      } else if (subscription.type === 'SMS') {
-        return ModelName.SmsSubscriptions;
-      } else {
-        return ModelName.PushSubscriptions;
-      }
-    };
-
     subscriptions.forEach(async (subscription) => {
-      const modelName = getModelName(subscription);
+      const modelName = ModelName.Subscriptions;
```
jkasten2, line 119: Since this is always `ModelName.Subscriptions` we don't need this `const` here. I see it being used in only two spots below, would could probably just refer to `ModelName.Subscriptions` directly instead of using a `const`, or at least move it down there.

src/core/CoreModuleDirector.ts
```diff
@@ -291,40 +330,58 @@ export class CoreModuleDirector {
...
-      }
+      case SubscriptionType.Email:
+        subscriptions = this.getEmailSubscriptionModels();
+        break;
+      case SubscriptionType.SMS:
+        subscriptions = this.getSmsSubscriptionModels();
+        break;
+      case ModelName.Subscriptions:
+        subscriptions = this.getAllPushSubscriptionModels();
+        break;
```
jkasten2, line 352: This seems like a bad assumption (or atleast a bad name), `ModelName.Subscriptions` sounds like all subscriptions (including SMS and Email), but it getting a list of push subceptions only.

src/core/executors/ExecutorBase.ts
```diff
@@ -85,7 +78,6 @@ export default abstract class ExecutorBase {
   }
 
   protected _flushDeltas(): void {
-    logMethodCall('ExecutorBase._flushDeltas');
```
jkasten2, line 88: Why remove this log?

jkasten2 (changes requested):

src/core/executors/ExecutorBase.ts
shepherd-l: This log gets called every second and I thought it cluttered the console, but I guess you can filter the logs. I'll add it back

jkasten2: ah I see, ya probably better to remove this line then

src/core/CoreModuleDirector.ts
```diff
@@ -291,40 +333,68 @@ export class CoreModuleDirector {
...
+      case SubscriptionType.SMS:
+        return SubscriptionChannel.SMS;
+      default:
+        if (this.isPushSubscriptionType(type)) {
+          return SubscriptionChannel.Push;
+        }
+
+        return undefined;
+    }
+  }
```
rgomezp, line 399: Move above PRIVATE comment

src/shared/services/IndexedDb.ts
```diff
@@ -251,6 +258,57 @@ export default class IndexedDb {
       );
     };
   }
+  private migrateModelNameSubscriptionsTableForV6(
```
rgomezp, line 261: Might be a good idea to add test coverage

src/core/operationRepo/OperationRepo.ts
```diff
@@ -7,6 +7,8 @@ import { logMethodCall } from '../../shared/utils/utils';
 export class OperationRepo {
   public executorStore: ExecutorStore;
   private _unsubscribeFromModelRepo: () => void;
+  private _deltaQueue: CoreDelta<SupportedModel>[] = [];
+  static DELTAS_BATCH_PROCESSING_TIME = 1;
```
rgomezp, line 11: Consider re-using
```
import { DELTA_QUEUE_TIME_ADVANCE } from '../../support/constants';
```

rgomezp (changes requested):

src/core/operationRepo/OperationRepo.ts
shepherd-l: I don't think we should be using it here because I think there should be a one way dependency between files in `__test__` to `src` - and `DELTA_QUEUE_TIME_ADVANCE` is in `__test__/support/constants.ts `

src/shared/services/IndexedDb.ts
shepherd-l: I've added test coverage for this here: https://github.com/OneSignal/OneSignal-Website-SDK/pull/1179/commits/fcf13ee257092301c9f53a1ba7a9b6ae0d24775f, https://github.com/OneSignal/OneSignal-Website-SDK/pull/1179/commits/0633e7caabffc60e88c4e236f3ccd7dd614b238e

Code coverage report:
| File              | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s                              |
|-------------------|---------|----------|---------|---------|------------------------------------------------|
| IndexedDb.ts      | 79.16   | 80       | 64.28   | 79.16   | 67-93,113,129,222,255,278,305,333-355,379,401-407,431-436 |

src/core/CoreModuleDirector.ts
shepherd-l: I've created a new `SubscriptionChannel` enum: https://github.com/OneSignal/OneSignal-Website-SDK/pull/1179/commits/f260cab3ef4ee4e815e3eb569aa3101162be5cf3

src/core/CoreModuleDirector.ts
shepherd-l: I found a spot here: https://github.com/OneSignal/OneSignal-Website-SDK/pull/1179/commits/a2a8ef1a48ff7260d52364786b25ab8576d347ec

Did I miss anything else?

src/shared/services/IndexedDb.ts
rgomezp: ah, missed that. thanks

jkasten2 (approved):

src/core/CoreModuleDirector.ts
```diff
@@ -291,35 +333,63 @@ export class CoreModuleDirector {
...
+    switch (type) {
+      case SubscriptionType.ChromePush:
+      case SubscriptionType.SafariPush:
+      case SubscriptionType.SafariLegacyPush:
+      case SubscriptionType.FirefoxPush:
+        return true;
+      default:
+        return false;
+    }
+  }
```
rgomezp, line 378: This method should probably be moved elsewhere like `SubscriptionModels.ts`

src/core/CoreModuleDirector.ts
```diff
@@ -291,35 +333,63 @@ export class CoreModuleDirector {
...
+        return SubscriptionChannel.Email;
+      case SubscriptionType.SMS:
+        return SubscriptionChannel.SMS;
+      default:
+        if (this.isPushSubscriptionType(type)) {
+          return SubscriptionChannel.Push;
+        }
+
+        return undefined;
+    }
```
rgomezp, line 392: This method should probably be moved elsewhere like `SubscriptionModels.ts`

rgomezp (changes requested):
Just a couple comments. After this I think LGTM

src/core/CoreModuleDirector.ts
shepherd-l: I moved both methods into `SubscriptionHelper.ts` https://github.com/OneSignal/OneSignal-Website-SDK/pull/1179/commits/d4bbe89a7513745d8bf7bb0c9efb1464bd31c361

rgomezp (approved):

PR #1180, opened 2024-07-09
Change `sendSelfNotification` to use local push

rgomezp:
# Description
## 1 Line Summary
Change the `sendSelfNotification` method to use local push functionality.

# Systems Affected
   - [x] WebSDK
   - [ ] Backend
   - [ ] Dashboard

# Validation
- Call `OneSignal.sendSelfNotification()`
- See that the notification is shown

## Testing
```js
OneSignal.sendSelfNotification('Title', 'message', 'https://google.com', 'https://t3.ftcdn.net/jpg/03/08/73/34/360_F_308733458_QBzwMVu8ZzdGEp9Wwq1fAYaDgtP3UVwl.jpg', { test: 'foo' }, [{
    action: 'like-button',
    title: 'Like',
    icon: 'https://image.similarpng.com/very-thumbnail/2020/06/Icon-like-button-transparent-PNG.png'
}]);
```

| Feature           | Chrome | Firefox | Safari |
|-------------------|--------|---------|--------|
| Title             | ✅      | ❌       | ❌      |
| Message           | ✅      | ❌       | ❌      |
| Launch Url Click  | ✅      | ❌       | ❌      |
| Data              | ✅      | ❌       | ❌      |
| Button Displayed  | ✅      | ❌       | ❌      |
| Buttons Displayed | ✅      | ❌       | ❌      |
| Button Click      | ✅      | ❌       | ❌      |

### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [x] No Typescript warnings

Other:

### Checklist

---

## Related Tickets

---

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Website-SDK/1180)

### DIFF
--- a/src/OneSignal.ts
+++ b/src/OneSignal.ts
@@ -36,6 +36,7 @@ import {
   awaitOneSignalInitAndSupported,
   executeCallback,
   getConsoleStyle,
+  getPlatformNotificationIcon,
   isValidEmail,
   logMethodCall,
 } from './utils';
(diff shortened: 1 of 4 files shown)


src/OneSignal.ts
```diff
@@ -731,11 +732,25 @@ export default class OneSignal {
...
+        Log.error("Service worker registration not available.");
+        return;
+      }
+
+      const options: NotificationOptions = {
+        body: message,
+        data: data,
+        icon: icon,
+      };
+      registration.showNotification(title, options);
```
jkasten2, line 752: The OneSignal REST API and the brower's `showNotification()` data format is has a ton of differences. While the SDK's internal welcome notification doesn't use many of the features / parameters customer may be doing so, since this is a public API on the OneSignal class.

jkasten2 (changes requested):

jkasten2 (changes requested):
Need to include what was tested.
* Does setting a title work?
* Does clicking on the notification work? (fire the click event, go to the URL, etc)
* Do action buttons work?
* Etc

PR #1181, opened 2024-07-10
Upgrade deps, build scripts, config, & lint for player-model branch

rgomezp:
# Description
## One line summary
Upgrade dependencies, fix type errors, and update build configurations to align with the main branch.

## Description
**Dependency Updates**
- Upgraded outdated dependencies to align the player-model version with the user-model version.
- Fixed type errors that appeared after updating dependencies.

**Build Script Updates**
- Updated build scripts to match those on the main branch.
- Created `stylesheet.js` to use new webpack loaders for importing SCSS files.

**Configuration Updates**
- Adjusted `tsconfig.json` to correctly handle ES5 as a valid target but not as a module type.
- Updated `sdk.config.js`, `page-sdk-es6.config.js`, `serviceworker.config.js`, and `webpack.config.js` to incorporate new webpack plugins and align with the main branch.
- Revised yarn scripts for better alignment with the main branch.

**Other**
- Applied SCSS lint fixes as auto-run by VSCode.

# Systems Affected
   - [x] WebSDK
   - [ ] Backend
   - [ ] Dashboard

### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [x] No Typescript warnings

Other:

## Related Tickets
https://app.asana.com/0/1207499441939378/1207762224893595
---

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Website-SDK/1181)


jkasten2 (changes requested):
CI tests are failing to start

PR #1183, opened 2024-07-11
Change the sendSelfNotification method to use local push functionality

rgomezp:
# Description
## 1 Line Summary
Change the `sendSelfNotification` method to use local push functionality.

# Systems Affected
   - [x] WebSDK
   - [ ] Backend
   - [ ] Dashboard

# Validation
- Call `OneSignal.sendSelfNotification()`
- See that the notification is shown

## Testing
```js
OneSignal.sendSelfNotification('Title', 'message', 'https://google.com', 'https://t3.ftcdn.net/jpg/03/08/73/34/360_F_308733458_QBzwMVu8ZzdGEp9Wwq1fAYaDgtP3UVwl.jpg', { test: 'foo' }, [{
    action: 'like-button',
    title: 'Like',
    icon: 'https://image.similarpng.com/very-thumbnail/2020/06/Icon-like-button-transparent-PNG.png'
}]);
```

| Feature           | Chrome | Firefox Windows | Firefox Mac | Safari |
|-------------------|--------|---------|--------|--------|
| Welcome Notification             | ✅      | ✅      | ✅       | ❌      |
| Title             | ✅      | ✅      | ✅       | ❌      |
| Message           | ✅      | ✅      | ✅       | ❌      |
| Launch Url Click  | ✅      | ✅      | ✅       | ❌      |
| Data              | ✅      | ✅      | ✅       | ❌      |
| Button Displayed  | ✅      | ✅      | ✅       | ❌      |
| Buttons Displayed | ✅      | ✅      | ✅       | ❌      |
| Button Click      | ✅      | ✅      | ✅       | ❌      |

### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [x] No Typescript warnings

Other:

### Checklist

---

## Related Tickets

---

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Website-SDK/1183)

### DIFF
--- a/express_webpack/index.html
+++ b/express_webpack/index.html
@@ -154,6 +154,17 @@
         })
     }
 
+    function sendSelfNotification() {
+        OneSignal.push(function() {
+          OneSignal.sendSelfNotification('Title', 'message', `https://localhost:4001/?app_id=${appId}`, 'https://t3.ftcdn.net/jpg/03/08/73/34/360_F_308733458_QBzwMVu8ZzdGEp9Wwq1fAYaDgtP3UVwl.jpg', { test: 'foo' }, [{
+            id: 'like-button',
+            text: 'Like',
+            icon: 'https://image.similarpng.com/very-thumbnail/2020/06/Icon-like-button-transparent-PNG.png',
+            url: 'https://onesignal.com'
+        }]);
+        });
+    }
+
   </script>
 <head>
     <meta charset="utf-8">
@@ -189,6 +200,7 @@
     <button onclick="javascript:showSmsSlidedown();">Show Sms Slidedown</button>
     <button onclick="javascript:showEmailSlidedown();">Show Email Slidedown</button>
     <button onclick="javascript:showSmsAndEmailSlidedown();">Show Sms & Email Slidedown</button>
+    <button onclick="javascript:sendSelfNotification();">Send Self Notification</button>
     <br />
     <br />
     <div class='onesignal-customlink-container'></div>
--- a/src/OneSignal.ts
+++ b/src/OneSignal.ts
@@ -36,6 +36,7 @@ import {
   awaitOneSignalInitAndSupported,
   executeCallback,
   getConsoleStyle,
+  getPlatformNotificationIcon,
   isValidEmail,
   logMethodCall,
 } from './utils';
@@ -713,16 +714,16 @@ export default class OneSignal {
   /**
    * @PublicApi
    */
-  static async sendSelfNotification(title: string = 'OneSignal Test Message',
-                              message: string = 'This is an example notification.',
-                              url: string = `${new URL(location.href).origin}?_osp=do_not_open`,
-                              icon: URL,
-                              data: Map<String, any>,
-                              buttons: Array<NotificationActionButton>): Promise<void> {
+  static async sendSelfNotification(title = 'OneSignal Test Message',
+                              message = 'This is an example notification.',
+                              url = `${new URL(location.href).origin}?_osp=do_not_open`,
+                              icon?: string,
+                              data?: Record<string, any>,
+                              buttons?: Array<any>): Promise<void> {
     await awaitOneSignalInitAndSupported();
     logMethodCall('sendSelfNotification', title, message, url, icon, data, buttons);
     const appConfig = await Database.getAppConfig();
-    const subscription = await Database.getSubscription();
+
     if (!appConfig.appId)
       throw new InvalidStateError(InvalidStateReason.MissingAppId);
     if (!(await OneSignal.isPushNotificationsEnabled()))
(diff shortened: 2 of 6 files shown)


src/OneSignal.ts
```diff
@@ -713,16 +714,16 @@ export default class OneSignal {
...
-                              url: string = `${new URL(location.href).origin}?_osp=do_not_open`,
-                              icon: URL,
-                              data: Map<String, any>,
-                              buttons: Array<NotificationActionButton>): Promise<void> {
+  static async sendSelfNotification(title = 'OneSignal Test Message',
+                              message = 'This is an example notification.',
+                              url = `${new URL(location.href).origin}?_osp=do_not_open`,
+                              icon?: string,
+                              data?: Record<string, any>,
+                              buttons?: Array<NotificationAction>): Promise<void> {
```
jkasten2, line 722: Changing the type from `NotificationActionButton` to `NotificationAction` is a breaking change. The format is different between JS and the OneSignal API but we need to keep the backwards compatibility:
* https://documentation.onesignal.com/reference/push-channel-properties#action-buttons
* https://developer.mozilla.org/en-US/docs/Web/API/Notification/actions

src/OneSignal.ts
```diff
@@ -731,11 +732,32 @@ export default class OneSignal {
...
+      // get default icon
+      const icons = await MainHelper.getNotificationIcons();
+      icon = getPlatformNotificationIcon(icons);
+    }
 
-    if (subscription.deviceId) {
-      await OneSignalApi.sendNotification(appConfig.appId, [subscription.deviceId], { en : title }, { en : message },
-                                               url, icon, data, buttons);
+    data = {
+      ...data,
```
jkasten2, line 742: We needed actually to nest the data in the data object, as we want the click event to produce the format in the docs:
https://documentation.onesignal.com/v9.0/docs/web-push-sdk#addlistenerfornotificationopened-event

However this is what I am seeing instead for the click event when I used `OneSignal.sendSelfNotification("title", "msg", "https://localhost:4001/?app_id=a3268fe7-cc94-4f25-ab09-867b957d7fe2", "https://localhost:4001/icon", {key1: "value1"})` to create it.
```js
{
    "key1": "value1",
    "url": "https://localhost:4001/?app_id=a3268fe7-cc94-4f25-ab09-867b957d7fe2"
}
```

jkasten2 (changes requested):

src/OneSignal.ts
rgomezp: [That type](https://github.com/OneSignal/OneSignal-Website-SDK/blob/player-model-main/src/models/NotificationActionButton.ts#L3) has always been wrong then cause it doesn't match our API.

```ts
{
   action: string;
   title: string;
   icon: URL;
   url: URL;
}
```

This results in the error:

```
No overload matches this call.
  Overload 1 of 2, '(title: string, options?: NotificationOptions): Promise<void>', gave the following error.
    Argument of type '{ body: string; data: { data: Record<string, any>; url: string; buttons: NotificationActionButton[]; }; icon: string; actions: NotificationActionButton[]; }' is not assignable to parameter of type 'NotificationOptions'.
      Types of property 'actions' are incompatible.
        Type 'NotificationActionButton[]' is not assignable to type 'NotificationAction[]'.
          Type 'NotificationActionButton' is not assignable to type 'NotificationAction'.
            Types of property 'icon' are incompatible.
              Type 'URL' is not assignable to type 'string'.
  Overload 2 of 2, '(title: string, options?: NotificationOptions): Promise<void>', gave the following error.
    Argument of type '{ body: string; data: { data: Record<string, any>; url: string; buttons: NotificationActionButton[]; }; icon: string; actions: NotificationActionButton[]; }' is not assignable to parameter of type 'NotificationOptions'.

759       registration.showNotification(title, options);
```

express_webpack/index.html
```diff
@@ -154,6 +154,17 @@
         })
     }
 
+    function sendSelfNotification() {
+        OneSignal.push(function() {
+          OneSignal.sendSelfNotification('Title', 'message', 'https://localhost:4001/?app_id=90ab655e-f1b6-4a72-9b0e-69e842f24033', 'https://t3.ftcdn.net/jpg/03/08/73/34/360_F_308733458_QBzwMVu8ZzdGEp9Wwq1fAYaDgtP3UVwl.jpg', { test: 'foo' }, [{
```
jkasten2, line 159: If we leave this new test function in here we should pull the `app_id` instead of hard coding one.

src/OneSignalApi.ts
```diff
@@ -17,10 +17,6 @@ export default class OneSignalApi {
     return OneSignalApiShared.updatePlayer(appId, playerId, options);
   }
 
-  static sendNotification(appId: string, playerIds: Array<string>, titles, contents, url, icon, data, buttons) {
-    return OneSignalApiShared.sendNotification(appId, playerIds, titles, contents, url, icon, data, buttons);
```
jkasten2, line 21: We should also clean up sendNotification from the `OneSignalApiShared` class too.

jkasten2 (changes requested):

jkasten2 (approved):

PR #1189, opened 2024-09-03
[Fix] External Id hydration

shepherd-l:
# Description
## 1 Line Summary
Fixes outdated external ids caused by OneSignal returning the user's previous external id in its response for user GET requests when called quickly after the user was created.

## Details
The server may still be updating and can return an outdated external Id from the getUser request. Instead of waiting for the updated external Id, we can hydrate the user with the local external Id as getting a response from the server means that the login was successful.

This PR also includes:
Adding a delay after creating a new user or push subscription as the OneSignal's backend can return incorrect 404 responses. 
- Adds a new records state to keep track of recently created users and subscriptions inorder to delay appropriate operations afterwards. The delay is needed because after the server creates the user/subscription, there is a small amount of time where the data is not up to date.
- The fix done in https://github.com/OneSignal/OneSignal-Android-SDK/pull/2059 and https://github.com/OneSignal/OneSignal-iOS-SDK/pull/1470

# Systems Affected
   - [x] WebSDK
   - [ ] Backend
   - [ ] Dashboard

# Validation
### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [x] No Typescript warnings

Other:

### Checklist

---

## Related Tickets

---

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Website-SDK/1189)

### DIFF
--- a/src/core/executors/ExecutorBase.ts
+++ b/src/core/executors/ExecutorBase.ts
@@ -9,12 +9,14 @@ import { ExecutorResult } from './ExecutorResult';
 import Log from '../../shared/libraries/Log';
 import Database from '../../shared/services/Database';
 import LocalStorage from '../../shared/utils/LocalStorage';
+import { NewRecordsState } from '../../shared/models/NewRecordsState';
 
 const RETRY_AFTER = 5_000;
 
 export default abstract class ExecutorBase {
   protected _deltaQueue: CoreDelta<SupportedModel>[] = [];
   protected _operationQueue: Operation<SupportedModel>[] = [];
+  protected _newRecordsState: NewRecordsState;
 
   protected _executeAdd?: (
     operation: Operation<SupportedModel>,
@@ -31,7 +33,10 @@ export default abstract class ExecutorBase {
   static OPERATIONS_BATCH_PROCESSING_TIME = 5;
   static RETRY_COUNT = 5;
 
-  constructor(executorConfig: ExecutorConfig<SupportedModel>) {
+  constructor(
+    executorConfig: ExecutorConfig<SupportedModel>,
+    newRecordsState: NewRecordsState,
+  ) {
     setInterval(() => {
       Log.debug('OneSignal: checking for operations to process from cache');
       const cachedOperations = this.getOperationsFromCache();
@@ -48,6 +53,8 @@ export default abstract class ExecutorBase {
     this._executeAdd = executorConfig.add;
     this._executeUpdate = executorConfig.update;
     this._executeRemove = executorConfig.remove;
+
+    this._newRecordsState = newRecordsState;
   }
 
   abstract processDeltaQueue(): void;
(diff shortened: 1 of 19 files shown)


src/core/executors/ExecutorBase.ts
```diff
@@ -188,4 +189,20 @@ export default abstract class ExecutorBase {
       this._processOperationQueue.call(this);
     }
   }
+
+  private _prepareforExecution(operation: Operation<SupportedModel>): void {
```
jkasten2, line 193: I like that when you added more code you split this out to keep functions smaller, however I think this function name is misleading. It is doing more than prepare, it also executes.

One suggestion that comes to mind is rename this function to `canExecute` returning a boolean. Then move  `this._processOperation` back to the while loop in `_processOperationQueue`.

src/shared/models/NewRecordsState.ts
```diff
@@ -0,0 +1,29 @@
...
+  OP_REPO_POST_CREATE_DELAY: number;
+  // Key = a string id
+  // Value = A Timestamp in ms of when the id was created
+  private records: Map<string, number> = new Map();
+
+  constructor(time = 3000) {
+    this.OP_REPO_POST_CREATE_DELAY = time;
+  }
+
+  public add(key: string, overwrite = false): void {
```
jkasten2, line 17: ah good idea on the overwrite logic. Maybe we should add this to the Android SDK as well.

src/page/managers/LoginManager.ts
```diff
@@ -305,6 +336,15 @@ export default class LoginManager {
   static async fetchAndHydrate(onesignalId: string): Promise<void> {
     logMethodCall('LoginManager.fetchAndHydrate', { onesignalId });
 
+    const newRecordsState = OneSignal.coreDirector.getNewRecordsState();
+
+    if (!newRecordsState) {
+      Log.error(`fetchAndHydrate: NewRecordsState is undefined`);
+    } else {
+      const delay = newRecordsState.OP_REPO_POST_CREATE_DELAY;
+      await awaitableTimeout(delay);
```
jkasten2, line 345: The `delay` variable doesn't add anything here, recommend we just use OP_REPO_POST_CREATE_DELAY directly.
* _The rules of thumb I try to use in combining lines of code or splitting them out is; split up long lines / very nested patterns to access a value, logic getting a value (unless it is very simply), and some subjectiveness; less code is normally better; less bugs and quicker to read (but if it isn't obviously clear to other reads then splitting lines up is better)_

jkasten2 (changes requested):

src/shared/models/NewRecordsState.ts
shepherd-l: Got it from iOS, iOS has this logic: https://github.com/OneSignal/OneSignal-iOS-SDK/pull/1470/files#diff-478e4e86941c7e4be07cbc74b0f7d337844d1c7254ab37ab92a1d50c4da2b006R49

jkasten2 (approved):

jkasten2 (approved):

PR #1192, opened 2024-09-11
Fix prompting again after closing notification permission prompt if site has its own ServiceWorker

jkasten2:
# Description
## 1 Line Summary
Fix prompting again after closing notification permission prompt if site has its own ServiceWorker.

## Details
This PR fixes two things related to this issue:
1. requestNotificationPermission() - this was always returning `undefined` due to a conversion bug
2. getRegistration() - did not check if the service worker was a OneSignal one, resulting in wrong assumptions being made.

# Validation
## Tests
Added a unit test for `requestNotificationPermission()`, this was always returning `undefined` before the changes in the first commit.

### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [X] No Typescript warnings

Other:

### Checklist

---

## Related Tickets

---

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Website-SDK/1192)

### DIFF
--- a/__test__/unit/notifications/subscriptionmanager.test.ts
+++ b/__test__/unit/notifications/subscriptionmanager.test.ts
@@ -1,0 +1,32 @@
+import MockNotification from '../../support/mocks/MockNotification';
+import { SubscriptionManager } from '../../../src/shared/managers/SubscriptionManager';
+import { NotificationPermission } from '../../../src/shared/models/NotificationPermission';
+
+describe('SubscriptionManager', () => {
+  describe('requestNotificationPermission', () => {
+    beforeEach(() => {
+      window.Notification = MockNotification;
+    });
+
+    test('default', async () => {
+      MockNotification.permission = 'default';
+      expect(await SubscriptionManager.requestNotificationPermission()).toBe(
+        NotificationPermission.Default,
+      );
+    });
+
+    test('denied', async () => {
+      MockNotification.permission = 'denied';
+      expect(await SubscriptionManager.requestNotificationPermission()).toBe(
+        NotificationPermission.Denied,
+      );
+    });
+
+    test('granted', async () => {
+      MockNotification.permission = 'granted';
+      expect(await SubscriptionManager.requestNotificationPermission()).toBe(
+        NotificationPermission.Granted,
+      );
+    });
+  });
+});
--- a/src/shared/helpers/ServiceWorkerHelper.ts
+++ b/src/shared/helpers/ServiceWorkerHelper.ts
@@ -279,12 +279,6 @@ export enum ServiceWorkerActiveState {
    * No service worker is installed.
    */
   None = 'None',
-  /**
-   * Service workers are not supported in this environment. This status is used
-   * on HTTP pages where it isn't possible to know whether a service worker is
-   * installed or not or in any of the other states.
-   */
-  Indeterminate = 'Indeterminate',
 }
 
 export interface ServiceWorkerManagerConfig {
--- a/src/shared/libraries/WorkerMessenger.ts
+++ b/src/shared/libraries/WorkerMessenger.ts
@@ -197,7 +197,7 @@ export class WorkerMessenger {
     );
 
     const workerRegistration =
-      await this.context?.serviceWorkerManager.getRegistration();
+      await this.context?.serviceWorkerManager.getOneSignalRegistration();
     if (!workerRegistration) {
       Log.error(
         '`[Worker Messenger] [Page -> SW] Could not get ServiceWorkerRegistration to postMessage!',
--- a/src/shared/managers/ServiceWorkerManager.ts
+++ b/src/shared/managers/ServiceWorkerManager.ts
@@ -1,9 +1,7 @@
 import Environment from '../helpers/Environment';
 import { WorkerMessengerCommand } from '../libraries/WorkerMessenger';
 import Path from '../models/Path';
-import SdkEnvironment from './SdkEnvironment';
 import Database from '../services/Database';
-import { WindowEnvironmentKind } from '../models/WindowEnvironmentKind';
 import Log from '../libraries/Log';
 import OneSignalEvent from '../services/OneSignalEvent';
 import ServiceWorkerRegistrationError from '../errors/ServiceWorkerRegistrationError';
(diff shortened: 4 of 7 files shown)


shepherd-l (approved):
:lgtm:

Can confirm that the native permission prompt does not show again after dismissing with "X" with this fix. Tested with the example web app using a using a dummy non-OneSignal service worker.

Bug reproduced by:
Using the example web app using a dummy non-OneSignal service worker.
1. Clear web cache
2. Open website
3. Click "Subscribe" on slide down push prompt
4. Dismiss native permission prompt with "X"

- Main: Shows native permission prompt again
- This PR: Does not show native permission prompt

> *[Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Website-SDK/1192#-:-O6XKLYOCvsj6MCZGodt:b-pn2u2q)* status: 0 of 7 files reviewed, 2 unresolved discussions (waiting on @emawby, @jennantilla, @jinliu9508, @jkasten2, @nan-li, and @rgomezp)

___
*[`src/shared/managers/ServiceWorkerManager.ts` line 42 at r3](https://reviewable.io/reviews/OneSignal/OneSignal-Website-SDK/1192#-O6XM0eS9q8byvXv2kdh:-O6XM0eT6RW7MDL_f7Pd:b-5yp9ul) ([raw file](https://github.com/OneSignal/OneSignal-Website-SDK/blob/d2302b6026138db358ba786a7da6ecae159af01c/src/shared/managers/ServiceWorkerManager.ts#L42)):*
> ```XML
>     ServiceWorkerRegistration | null | undefined
>   > {
>     return await ServiceWorkerUtilHelper.getRegistration(
> ```

Why was this await removed? Isn't `getRegistration()` an async method that should be awaited?
___
*[`src/shared/managers/SubscriptionManager.ts` line 336 at r1](https://reviewable.io/reviews/OneSignal/OneSignal-Website-SDK/1192#-O6XKTyJ7XlJlvWzabwa:-O6XKTyJ7XlJlvWzabwb:b8nah3) ([raw file](https://github.com/OneSignal/OneSignal-Website-SDK/blob/a8286383dcad17fe91785fa24ff3fa5346697938/src/shared/managers/SubscriptionManager.ts#L336)):*
> ```XML
>     // TODO: Clean up our custom NotificationPermission enum
>     //         in favor of TS union type NotificationPermission instead of converting
>     return results as NotificationPermission;
> ```

Nice find! I noticed that this fixes `Error: OneSignal service worker not found!` from the tests: `registerForPushNotifications: before OneSignal.initialized` and `registerForPushNotifications: after OneSignal.initialized` even though they still passed before, and now log `PushPermissionNotGrantedError: The user dismissed the permission prompt.`

jkasten2 (commented):
Thanks for testing!

> *[Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Website-SDK/1192#-:-O6Xei0W8qGCZBFM9iej:b9v7vwa)* status: 0 of 7 files reviewed, 2 unresolved discussions (waiting on @emawby, @jennantilla, @jinliu9508, @nan-li, @rgomezp, and @shepherd-l)

___
*[`src/shared/managers/ServiceWorkerManager.ts` line 42 at r3](https://reviewable.io/reviews/OneSignal/OneSignal-Website-SDK/1192#-O6XM0eS9q8byvXv2kdh:-O6XdwJC108MpgNyyCDx:b-k8tdv3) ([raw file](https://github.com/OneSignal/OneSignal-Website-SDK/blob/d2302b6026138db358ba786a7da6ecae159af01c/src/shared/managers/ServiceWorkerManager.ts#L42)):*
<details><summary><i>Previously, shepherd-l (Shepherd) wrote…</i></summary><blockquote>

Why was this await removed? Isn't `getRegistration()` an async method that should be awaited?
</blockquote></details>

The function doesn't read the value, nor does it do anything after which would affect any order of operations.

Here is an example I put together to show there is no-side effects here:
```ts
async function asyncResult(): Promise<Number> {
  return Promise.resolve(123);
}

// Preferred (more concise) - This method doesn't consume the value, so we don't need 'await'.
async function noAwaitExample(): Promise<Number> {
  return asyncResult();
}

// Unnecessary (more verbose) - You can call 'await' but there isn't a point if the function doesn't need to read the value, or wait on order of operations
async function awaitExample(): Promise<Number> {
  return await asyncResult();
}

// Code showing both work the same, no-side effects.
async function printResults(): Promise<void> {
  console.log("awaitExample: " + await awaitExample());
  console.log("noAwaitExample: " + await noAwaitExample());
}
printResults();

```

Sandbox to try it out:
https://www.typescriptlang.org/play/?#code/IYZwngdgxgBAZgV2gFwJYHsI1JKAlAUxAQBtkAKASgC4YAFAJ3QFtUQCAeAOQWYCMCDAHwwA3gCgYMBgWQIGWRizYEAdDJDoSANwLkAjACYAzJQDc4gL7jxAelv0ZcQTIAmMcs3QyYUTFBVKGABaGAAVAAs2GGZZCPR3V3QiCAByZF9MYliYZAiCGG1gEgQCABoYTRgAdwKktIyIAgJ3VOBq4FRkVNVxHGh4JCg0TBgIdABBDq6AUQAPYGYABxI9Gkdldm5eAWExSWlZeSx+-CJSCnMrG3sYAFUIJqgiEGAGMA8vH10GPnR2IKhACa6AQvmAWCgxRIMDa026MD4CAyeUEBTYDWwMCW6FQEAyqDguXygxQGCwSRS6TGzXcyHQh2AdJJRRK5Rg3hqnQyo28rkEHKJ6CWgmAIwgID64AGiDJo3a3PmixWa1oSlYWx4-EEIgkUhkcgU2Hh2GlZ2IZCoFmsdgcAGEEgUQPFqniAOaI9B5GreADWxKdi3Z42CIFQ-JgBDgzmGIF6p1Jw3J2IYeOQhAtyBAVDVTA1nG0uNcuoOfglWjUJHQbvIACIFbMFstVrRazAANTG7ldxvK1ZUK5SMuaVaqKs12vjKaKpsq1sdnuNSbwpXNtZXaxLVP4jMXbNXIA
___
*[`src/shared/managers/SubscriptionManager.ts` line 336 at r1](https://reviewable.io/reviews/OneSignal/OneSignal-Website-SDK/1192#-O6XKTyJ7XlJlvWzabwa:-O6XeY7W14T_056bxNuB:b5ocjkj) ([raw file](https://github.com/OneSignal/OneSignal-Website-SDK/blob/a8286383dcad17fe91785fa24ff3fa5346697938/src/shared/managers/SubscriptionManager.ts#L336)):*
<details><summary><i>Previously, shepherd-l (Shepherd) wrote…</i></summary><blockquote>

Nice find! I noticed that this fixes `Error: OneSignal service worker not found!` from the tests: `registerForPushNotifications: before OneSignal.initialized` and `registerForPushNotifications: after OneSignal.initialized` even though they still passed before, and now log `PushPermissionNotGrantedError: The user dismissed the permission prompt.`
</blockquote></details>

ya, this is why I created a new unit test file to make sure the conversion is done correctly.

PR #1194, opened 2024-09-19
Fix logout() when push subscription doesn't exists

jkasten2:
# Description
## One-Line Summary
Fix OneSignal.logout() so it doesn't throw an error if there isn't a push subscription.

## Details
A push subscription only exists if the end-user accepted notifications at some point. `OneSignal.logout()` made this bad assumption it always existed.

The fix is to check this and create a local only user instead of making a call to OneSignal if there isn't a push subscription.

# Validation
## Tests
Tested on Chrome Version 128.0.6613.138 on Windows 11 23H2
Scenarios:
1. OneSignal.login("A");
2. OneSignal.User.addEmail("a@a.com");
3. OneSignal.logout();
4. OneSignal.User.addEmail("b@b.com");
5. Observe b@b.com was added to the correct User.

### Checklist
      - No current login / logout tests, there is another project to add test coverage to this flow.

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [X] No Typescript warnings

Other:

---

## Related Tickets
* https://github.com/OneSignal/onesignal-vue3/issues/43
* https://github.com/OneSignal/react-onesignal/issues/140

---

### DIFF
--- a/src/page/managers/LoginManager.ts
+++ b/src/page/managers/LoginManager.ts
@@ -143,12 +143,20 @@ export default class LoginManager {
     const pushSubModel =
       await OneSignal.coreDirector.getPushSubscriptionModel();
     await OneSignal.coreDirector.resetModelRepoAndCache();
+
+    // Initialize as a local User, as we don't have a push subscription to create a remote anonymous user.
+    if (pushSubModel === undefined) {
+      await UserDirector.initializeUser(true);
+      return;
+    }
+
     // add the push subscription model back to the repo since we need at least 1 sub to create a new user
     OneSignal.coreDirector.add(
       ModelName.PushSubscriptions,
       pushSubModel as OSModel<SupportedModel>,
       false,
     );
+    // Initialize as non-local, make a request to OneSignal to create a new anonymous user
     await UserDirector.initializeUser(false);
   }
 


shepherd-l (approved):
:lgtm:

Tested on Chrome Version 128.0.6613.138 on macOS Sonoma 14.6.1 with the web example app

Steps:
1. Clear web cache
2. Open example web app and do not accept the push notification permission (inorder to not create a push sub)
3. OneSignal.login("TestUser");
4. OneSignal.logout();

This PR does not produce an error
Reproduced error on main:
```
Dev-OneSignalSDK.page.es6.js?v=160202:2072 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'modelId')  
    at OSModelStore.subscribeUpdateListener (Dev-OneSignalSDK.page.es6.js?v=160202:2072:41)  
    at OSModelStore.add (Dev-OneSignalSDK.page.es6.js?v=160202:2056:14)  
    at CoreModuleDirector.add (Dev-OneSignalSDK.page.es6.js?v=160202:915:32)  
    at LoginManager.<anonymous> (Dev-OneSignalSDK.page.es6.js?v=160202:6476:87)  
    at Generator.next (<anonymous>)  
    at fulfilled (Dev-OneSignalSDK.page.es6.js?v=160202:18277:58)
```

> *[Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Website-SDK/1194#-:-O7FEZXW8IuWdq462vs8:b-k66rd9)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @emawby and @rgomezp)

PR #1195, opened 2024-09-21
[Dev] Enable compile TypeScript checks; noUnusedLocals, noUnusedParameters, & strictFunctionTypes

jkasten2:
# Description
## One-Line Summary
Enforce some additional Typescript compile errors that were only IDE warnings.

## Details
Enable these TS config compile settings. noUnusedLocals, noUnusedParameters, & strictFunctionTypes

# Validation
### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [X] No Typescript warnings

Other:

### Checklist

---

## Related Tickets

---

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Website-SDK/1195)

### DIFF
--- a/build/config/tsconfig.base.json
+++ b/build/config/tsconfig.base.json
@@ -27,6 +27,7 @@
     "sourceRoot": "../..",
     "rootDir": "../..",
     "esModuleInterop": true,
+    "strictFunctionTypes": true,
     "baseUrl": "../..",
     "types": [
       "node"
--- a/build/config/tsconfig.base.json
+++ b/build/config/tsconfig.base.json
@@ -8,7 +8,7 @@
     "noImplicitAny": false,
     "noImplicitReturns": true,
     "noImplicitThis": true,
-    "noUnusedLocals": false,
+    "noUnusedLocals": true,
     "noUnusedParameters": false,
     "allowSyntheticDefaultImports": true,
     "allowUnreachableCode": false,
--- a/build/config/tsconfig.base.json
+++ b/build/config/tsconfig.base.json
@@ -9,7 +9,7 @@
     "noImplicitReturns": true,
     "noImplicitThis": true,
     "noUnusedLocals": true,
-    "noUnusedParameters": false,
+    "noUnusedParameters": true,
     "allowSyntheticDefaultImports": true,
     "allowUnreachableCode": false,
     "allowUnusedLabels": false,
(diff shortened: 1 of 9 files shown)


shepherd-l (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Website-SDK/1195#-:-O7Uz4-aATj4lu0XWGUA:bnfp4nl)* status: 0 of 9 files reviewed, 1 unresolved discussion (waiting on @emawby, @jkasten2, and @rgomezp)

___
*[`src/shared/utils/Encoding.ts` line 83 at r3](https://reviewable.io/reviews/OneSignal/OneSignal-Website-SDK/1195#-O7UvVb8CQ9QBVDvcbfi:-O7UvVb8CQ9QBVDvcbfj:b-ads47w) ([raw file](https://github.com/OneSignal/OneSignal-Website-SDK/blob/700f8ec66c6bd9c890a9aa58b828c2effc39d372/src/shared/utils/Encoding.ts#L83)):*
> ```XML
>     encodeURIComponent(str).replace(
>       /%([0-9A-F]{2})/g,
>       function toSolidBytes(_match, p1) {
> ```

Should we just remove the method instead? Both `base64Encode()` and `base64Decode()`
I don't see them being referenced anywhere and the methods don't look like this in the docs referenced in the comment above: https://developer.mozilla.org/en-US/docs/Web/API/WindowBase64/Base64_encoding_and_decoding

jkasten2 (commented):
> *[Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Website-SDK/1195)* status: 0 of 9 files reviewed, 1 unresolved discussion (waiting on @emawby, @rgomezp, and @shepherd-l)

___
*[`src/shared/utils/Encoding.ts` line 83 at r3](https://reviewable.io/reviews/OneSignal/OneSignal-Website-SDK/1195#-O7UvVb8CQ9QBVDvcbfi:-O7V3IYE6M9Qd05NVhJf:b30sf6p) ([raw file](https://github.com/OneSignal/OneSignal-Website-SDK/blob/700f8ec66c6bd9c890a9aa58b828c2effc39d372/src/shared/utils/Encoding.ts#L83)):*
<details><summary><i>Previously, shepherd-l (Shepherd) wrote…</i></summary><blockquote>

Should we just remove the method instead? Both `base64Encode()` and `base64Decode()`
I don't see them being referenced anywhere and the methods don't looking like this in the docs referenced above today: https://developer.mozilla.org/en-US/docs/Web/API/WindowBase64/Base64_encoding_and_decoding
</blockquote></details>

ah good call, didn't look into the usage when I made the change. I confirmed they are not used and just made a commit here ot remove them in this PR.

PR #1206, opened 2024-11-14
test: explicitly assert for Log.error calls in `registerForPush`

sherwinski:
# Description
## 1 Line Summary

Explicitly tests for an error in the `registerForPush` test.

## Details

When running current tests, the following error is emitted:

```
  console.error
    PushPermissionNotGrantedError: The user dismissed the permission prompt.
        at SubscriptionManager.subscribeFcmFromPage (/Website-SDK/src/shared/managers/SubscriptionManager.ts:461:17)
        at SubscriptionManager.subscribe (/Website-SDK/src/shared/managers/SubscriptionManager.ts:162:13)
        at Function.internalRegisterForPush (/Website-SDK/src/shared/helpers/SubscriptionHelper.ts:30:35)
        at Function.registerForPush (/Website-SDK/src/shared/helpers/SubscriptionHelper.ts:20:12)
        at Function.registerForPushNotifications (/Website-SDK/src/shared/helpers/InitHelper.ts:104:5)
        at PromptsManager.internalShowNativePrompt (/Website-SDK/src/page/managers/PromptsManager.ts:188:5)
        at NotificationsNamespace.requestPermission (/Website-SDK/src/onesignal/NotificationsNamespace.ts:142:5)
        at PushSubscriptionNamespace.optIn (/Website-SDK/src/onesignal/PushSubscriptionNamespace.ts:96:7)
        at Object.<anonymous> (/Website-SDK/__test__/unit/push/registerForPush.test.ts:33:5) {
      reason: 1
    }
```

This change adds an assertion since the error is part of the test's [expected behavior](https://github.com/OneSignal/OneSignal-Website-SDK/pull/1206#issuecomment-2483933629).

~~Even though the error occurs, the [tests](https://github.com/OneSignal/OneSignal-Website-SDK/blob/main/__test__/unit/push/registerForPush.test.ts) still pass. For that reason, I assumed it's not part of the test's expected behavior. Please feel free to correct me if my understanding there is wrong.~~
 
~~**Note:** If needed we can add something like `expect(Log.error).toHaveBeenCalled()` to ensure that we are testing for the existence of the error.~~

# Systems Affected
   - [x] WebSDK
   - [ ] Backend
   - [ ] Dashboard

# Validation
### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [x] No Typescript warnings

Other:

### Checklist

---

## Related Tickets

---

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Website-SDK/1206)

### DIFF
--- a/__test__/unit/push/registerForPush.test.ts
+++ b/__test__/unit/push/registerForPush.test.ts
@@ -7,10 +7,18 @@ import BrowserUserAgent from '../../support/models/BrowserUserAgent';
 //stub dismisshelper
 jest.mock('../../../src/shared/helpers/DismissHelper');
 
+//stub log
+jest.mock('../../../src/shared/libraries/Log', () => ({
+  debug: jest.fn(),
+  trace: jest.fn(),
+  info: jest.fn(),
+  warn: jest.fn(),
+  error: jest.fn(),
+}));
+
 describe('Register for push', () => {
   beforeEach(async () => {
     jest.useFakeTimers();
-    jest.spyOn(Log, 'error').mockImplementation(() => {})
     await TestEnvironment.initialize({
       addPrompts: true,
       userAgent: BrowserUserAgent.Default,
@@ -31,6 +39,7 @@ describe('Register for push', () => {
     expect(spy).not.toHaveBeenCalled();
     OneSignalEvent.trigger(OneSignal.EVENTS.SDK_INITIALIZED);
     await promise;
+    expect(Log.error).toHaveBeenCalled();
     expect(OneSignal.initialized).toBe(true);
     expect(spy).toHaveBeenCalledTimes(1);
   });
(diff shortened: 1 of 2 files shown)


shepherd-l:
> Even though the error occurs, the [tests](https://github.com/OneSignal/OneSignal-Website-SDK/blob/main/__test__/unit/push/registerForPush.test.ts) still pass. For that reason, I assumed it's not part of the test's expected behavior. Please feel free to correct me if my understanding there is wrong.

I was also unsure at first but I think it's part of the test's expected behaviour as discussed here: https://github.com/OneSignal/OneSignal-Website-SDK/pull/1192#pullrequestreview-2298504262

>  Note: If needed we can add something like expect(Log.error).toHaveBeenCalled() to ensure that we are testing for the existence of the error.

Good idea, this makes sense to do inorder to avoid confusion

shepherd-l (approved):
Lgtm!

PR #1209, opened 2024-11-18
Chrome macOS 15 extra notification workaround

jkasten2:
# Description
## One Line Summary
Work around for [extra "This site has been updated in the background" notification being shown on Chrome on macOS 15](https://issues.chromium.org/issues/378103918).

## Details
macOS 15 made some unknown changes that causes Chrome's fragile notification display detection to fail, which means an extra notification with the body "This site has been updated in the background" displays.

This was reported to Google under this issue:
https://issues.chromium.org/issues/378103918

A workaround was found that if you delay ending the service worker event for a bit then Chrome will detect the notification from macOS. However this work around won't work 100% of the time since if a notification is closed very quickly (within the 1000ms). This window is also much easier to hit if notifications are send back-to-back. Lastly it is unknown if 1000ms is always enough time, but to much time isn't good either due what was noted earlier.

We scoped this workaround to the known target of Chromium on macOS 15 so this ~1% chance doesn't cause issues on other platforms. We can't detect the macOS version so this will effect all versions.

What this does not fix:
There is also an issue where if the OneSignal feature received receipts / confirmed deliveries is on and the notification is clicked or closed with in it's wait window this extra notification will display. We will address this in a different PR.

# Validation
## Tests
Tested on macOS 15.1 with Chrome and Firefox.

### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [X] No Typescript warnings

Other:

### Checklist

---

## Related Tickets

---


sherwinski (approved):
Suggestion: If it's low-effort, I'd recommend adding a test. But without actual browser e2e tests it might not provide a lot of value.

jkasten2:
> Suggestion: If it's low-effort, I'd recommend adding a test. But without actual browser e2e tests it might not provide a lot of value.

I added some tests in this commit
https://github.com/OneSignal/OneSignal-Website-SDK/pull/1209/commits/c3a3b09e9fed582c4bb7077c53b96ebfce87785f

sherwinski (approved):
Thanks for adding that 👍

PR #1213, opened 2024-12-13
160205 release

rgomezp:
# Description
## 1 Line Summary
Add ability to override remote service worker config via new override param `serviceWorkerOverrideForTypical`.

### Service Worker Configuration
- Introduced a new parameter, **serviceWorkerOverrideForTypical**, to allow users to override server settings for typical sites.

#### `docker-compose.yml`
- Fixed **volume mounting** issues to ensure the `releases` directory on the host accurately reflects the container contents.
- Updated the **version property**, as the previous one was deprecated.

#### `publish.sh`
- Modified the script to ensure **sourceMappingUrl** comments are correctly applied in output files, improving compatibility across environments.

### Release Version
- Updated the version to `160205` as part of the release process.

# Systems Affected
   - [x] WebSDK
   - [ ] Backend
   - [ ] Dashboard

# Validation
## Tests
Added test coverage. Tested locally.

### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [x] No Typescript warnings

Other:

### Checklist

---

## Related Tickets

---

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Website-SDK/1213)

### DIFF
--- a/__test__/support/environment/TestContext.ts
+++ b/__test__/support/environment/TestContext.ts
@@ -134,7 +134,6 @@ export default class TestContext {
             enable: false,
           },
           serviceWorker: {
-            customizationEnabled: false,
             path: '/',
             workerName: 'OneSignalSDKWorker.js',
             registrationScope: '/',
@@ -294,7 +293,6 @@ export default class TestContext {
           kind: configIntegrationKind,
         },
         serviceWorker: {
-          customizationEnabled: false,
           path: '/',
           workerName: 'OneSignalSDKWorker.js',
           registrationScope: '/',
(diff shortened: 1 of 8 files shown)


__test__/support/environment/TestContext.ts
```diff
@@ -294,10 +294,10 @@ export default class TestContext {
           kind: configIntegrationKind,
         },
         serviceWorker: {
-          path: undefined,
-          workerName: undefined,
-          registrationScope: undefined,
-          customizationEnabled: true,
+          customizationEnabled: false,
```
jkasten2, line 297: Looks like `customizationEnabled` is pre-existing dead code, so we could just clean this up instead of changing the value, since it has no effect anywhere.

jkasten2 (approved):

PR #1220, opened 2025-02-03
Add Service Worker Tests Part 1

fadi-george:
# Description
## 1 Line Summary
Part 1 of adding better tests for ServiceWorker.

## Details
- move service worker tests next to file its testing
- update service worker tests to use `event dispatch` to trigger to the event listeners and avoid mocking as much as we can
- added type-fest package for useful type utilities like 'PartialDeep`

# Systems Affected
   - [x] WebSDK
   - [ ] Backend
   - [ ] Dashboard

# Validation
### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [x] No Typescript warnings

Other:

### Checklist

---

## Related Tickets

---

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Website-SDK/1220)


src/sw/serviceWorker/ServiceWorker.ts
```diff
@@ -56,6 +56,7 @@ const MAX_CONFIRMED_DELIVERY_DELAY = 25;
  * allows notification permissions, and is a pre-requisite to subscribing for push notifications.
  */
 export class ServiceWorker {
+  // TODO: remove? UNSUBSCRIBED_FROM_NOTIFICATIONS is not set anywhere
```
sherwinski, line 59: Would be good to get an answer to this before moving forward

src/sw/serviceWorker/ServiceWorker.test.ts
sherwinski, line 1: 1. Is there a reason we're colocating these tests with the src code? Feels like we should stick to all tests living in the `__test__` directory. Although I did also notice [this file](https://github.com/OneSignal/OneSignal-Website-SDK/blob/main/src/shared/services/indexedDb.test.ts) floating in the src folder as well.
2. Mind combing through the old `serviceWorker.test.ts` file and making sure that all tests got migrated? Off the top of my head, I noticed that the `requiresMacOS15ChromiumAfterDisplayWorkaround` tests didn't make it over.

sherwinski (changes requested):

src/sw/serviceWorker/ServiceWorker.test.ts
fadi-george: 1. Can discuss more Tuesday but I think its helpful to see which files have tests by having tests file next to them. Exceptions can usually be e2e like playwright that would be better off with a separate folder. 
2. I did not cover all the events since I didn't want this pr to get too large so im splitting it up into multiple parts. E.g. next work part will have pushsubscription change and worker messages.

package.json
```diff
@@ -5,7 +5,6 @@
   "dependencies": {
     "@types/intl-tel-input": "^17.0.1",
     "bowser": "github:OneSignal/bowser#fix-android8-opr6-build-detection",
-    "core-js-bundle": "^3.1.3",
```
sherwinski, line 8: Want to specifically call this out to ensure it was intentional since it's a dependency.

sherwinski (commented):
Looks good 👍 

Would you mind documenting in the PR description why we need to use `dispatchEvent`? Is it due to migrating from jest to vitest, or something else?

fadi-george:
> Looks good 👍
> 
> Would you mind documenting in the PR description why we need to use `dispatchEvent`? Is it due to migrating from jest to vitest, or something else?

Yeah its in the pr details.

package.json
fadi-george: Theres no import of this library. It might've been accidentally added.

src/shared/services/IndexedDb.ts
```diff
@@ -3,7 +3,7 @@ import Utils from '../context/Utils';
 import Emitter from '../libraries/Emitter';
 import Log from '../libraries/Log';
 
-const DATABASE_VERSION = 6;
+export const DATABASE_VERSION = 6;
```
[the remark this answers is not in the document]
fadi-george: I changed the tests up a bit so no longer need to export it.

sherwinski (approved):

PR #1221, opened 2025-02-10
Replace yarn with npm install

fadi-george:
# Description
## 1 Line Summary
The yarn lock hasn't been updated in 2 years and modern npm install is comparable in terms of speed.

## Details
- Replace instances of `yarn` with `npm run` or `npm ci/install`
- The install step on the workflows take 11s instead of 22/23s
- Had to skip some tests for login and will need to be redone/fixed in a separate pr

# Systems Affected
   - [x] WebSDK
   - [ ] Backend
   - [ ] Dashboard

# Validation
### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [x] No Typescript warnings

Other:

### Checklist

---

## Related Tickets

---

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Website-SDK/1221)


.gitignore
```diff
@@ -6,12 +6,12 @@ build/bundles
...
 build/size-analysis.html
 /config.json
-package-lock.json
 
 .DS_STORE
 *.pem
 *.key
 *.crt
+yarn.lock
 yarn-error.log
```
jkasten2, line 15: Since we are removing yarn completely, there isn't a reason anyone should be using them, so we can remove these lines.

jkasten2 (changes requested):

__test__/unit/user/login.test.ts
```diff
@@ -216,9 +217,9 @@ describe('Login tests', () => {
     }
   });
 
-  test('If login with JWT token, save it to the database', async () => {});
+  // test('If login with JWT token, save it to the database', async () => {});
```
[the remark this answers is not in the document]
fadi-george: Done

.gitignore
fadi-george: Done

jkasten2 (approved):
Tested this PR with production settings, such as minification, with the express_webpack and ensure the slide down works, can subscribe, displays a welcome notification, can receive a push, and clicking on it opens the site.

To test the "production" like build I used this commit:
https://github.com/OneSignal/OneSignal-Website-SDK/commit/ea23d41e9851446795da6749477bcb1fe6a330fd

PR #1222, opened 2025-02-11
Replace jest with vitest

fadi-george:
# Description
## Details

# Systems Affected
   - [ ] WebSDK
   - [ ] Backend
   - [ ] Dashboard

# Validation
### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [ ] No Typescript warnings

Other:

### Checklist

---

## Related Tickets

---


__test__/support/environment/TestEnvironmentHelpers.ts
```diff
@@ -147,65 +145,3 @@ function addCustomEventPolyfill(windowDef) {
...
-  // Simulate a delayed request
-  const timeout = setTimeout(function () {
-    callback();
-  }, delay);
-  return {
-    abort: function () {
-      clearTimeout(timeout);
-    },
-  };
-}
```
sherwinski, line 211: Did this get replaced/refactored anywhere? Or was it just dead code?

__test__/unit/core/modelRepo.test.ts
```diff
@@ -109,120 +105,129 @@ describe('ModelRepo tests', () => {
...
-    core.modelRepo?.subscribe((delta: CoreDelta<SupportedModel>) => {
-      broadcastCount += 1;
-      if (isPropertyDelta(delta)) {
-        expect(delta.changeType).toBe(CoreChangeType.Update);
-        expect(delta.model).toEqual(newSub);
-        expect(delta.property).toBe('sdk');
-        expect(delta.newValue).toBe('mySdk');
-        done();
-      }
+    await new Promise<void>((resolve) => {
```
sherwinski, line 205: Could you provide some more context on why a lot of these checks are wrapped in a Promise. Was there test flakiness caused by the checks happening asynchronously with the coreDirector operations right after?

__test__/README.md
```diff
@@ -22,6 +22,12 @@ describe('My tests', () => {
 });
 ```
 
+# Extension
+
+You can use the [ms-playwright](https://marketplace.visualstudio.com/items?itemName=ms-playwright.playwright) extension to run individual by running a "Play" button next to the test case.
```
sherwinski, line 27:
```suggestion
You can use the [ms-playwright](https://marketplace.visualstudio.com/items?itemName=ms-playwright.playwright) extension to run individual tests by running a "Play" button next to the test case.
```

sherwinski (commented):
Wow, huge improvement in test speed. Thanks for doing this! 🚀 

Just a request for next time: it would be helpful to separate out any linting changes into their own commit so it's easier to pick out actual code changes from the diffs. Appreciate it!

__test__/support/environment/TestEnvironmentHelpers.ts
fadi-george: It's not referenced anywhere, seems to be dead code.

__test__/unit/core/modelRepo.test.ts
fadi-george: It's in the PR description but the reason is that Vitest doesn't provide a done callback so the alternative is a simple promise wrapper.

fadi-george:
> Wow, huge improvement in test speed. Thanks for doing this! 🚀
> 
> Just a request for next time: it would be helpful to separate out any linting changes into their own commit so it's easier to pick out actual code changes from the diffs. Appreciate it!

Will try but I was trying to change some lines from console.log -> console.info so I'll comment on those files to be more clear.

express_webpack/server.js
```diff
@@ -2,43 +2,51 @@ const path = require('path');
 const express = require('express');
```
fadi-george, line 2: Change is only console.log -> .info

express_webpack/amp/index.html
```diff
@@ -1,9 +1,9 @@
-<!DOCTYPE html>
-  <html lang="en">
+<!doctype html>
```
fadi-george, line 1: Change is only console.log -> .info

sherwinski (approved):

PR #1223, opened 2025-02-14
Fix Sync Call

fadi-george:
# Description
## 1 Line Summary
Updates the one signal api base requestHasAppId method to have a special condition for the sync/ call since it does not start `/apps` or have `app_id` in the body which leads to an error when calling `downloadServerAppConfig.`

Ticket: https://app.asana.com/0/1208777190614537/1209414057669993
Related issue: https://github.com/OneSignal/OneSignal-WordPress-Plugin/issues/351

## Details
- updates `requestHasAppId` as described above

# Systems Affected
   - [x] WebSDK
   - [ ] Backend
   - [ ] Dashboard

# Validation
### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [x] No Typescript warnings

Other:

### Checklist

---

## Related Tickets

---

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Website-SDK/1223)

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -8,6 +8,10 @@ module.exports = async () => {
     forceExit: true,
     bail: true,
     testEnvironment: 'jsdom',
+    moduleNameMapper: {
+      '^src/(.*)$': '<rootDir>/src/$1',
+      '^__test__/(.*)$': '<rootDir>/__test__/$1',
+    },
     // Run these files after jest has been
     // installed in the environment
     setupFilesAfterEnv: ['<rootDir>/__test__/jest/jest.setup.ts'], // use .js if you prefer JavaScript,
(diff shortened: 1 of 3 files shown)


sherwinski (approved):
LGTM

sherwinski (changes requested):
Sorry before I forget, could we add a test for this first before merging?

fadi-george:
> Sorry before I forget, could we add a test for this first before merging?

Done.

sherwinski (approved):
Assuming that we may move this test back to the test directory (pending [this comment](https://github.com/OneSignal/OneSignal-Website-SDK/pull/1220#discussion_r1956867678)), I'm good with this 👍

PR #1226, opened 2025-02-19
Service Worker Tests Part 2

fadi-george:
# Description
## 1 Line Summary
Improve test coverage of service worker to 73%

## Details
- Adds more test cases to service worker test file mainly testing the message listener
- Replace flushPromise calls with vi fake timers to improve test speed 
- Tests upsert and deactive session flows
- Updated vitest to fix issue with msw freezing up when using fake timers

# Systems Affected
   - [x] WebSDK
   - [ ] Backend
   - [ ] Dashboard

# Validation
### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [ ] No Typescript warnings

Other:

### Checklist

---

## Related Tickets

---

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Website-SDK/1226)


build/config/webpack.config.js
```diff
@@ -14,25 +15,29 @@ setBuildEnvironment();
...
+    case 'staging':
+    case 'production':
       break;
     default:
-      process.env.ENV = "development";
+      process.env.ENV = 'development';
   }
 
-  console.log("Build Environment:", process.env.ENV);
+  console.info('Build Environment:', process.env.ENV);
```
fadi-george, line 26: Only change

build/config/sdk.config.js
```diff
@@ -1,39 +1,40 @@
 const webpack = require('webpack');
```
fadi-george, line 1: Change is only console.log -> console.info

__test__/support/environment/TestContext.ts
```diff
@@ -19,8 +19,9 @@ export default class TestContext {
     overrideServerConfig: RecursivePartial<ServerAppConfig> | null = null,
     appId: string = APP_ID,
   ): ServerAppConfig {
+    let remoteConfigMockDefaults: ServerAppConfig;
     if (configIntegrationKind === ConfigIntegrationKind.Custom) {
-      return {
+      remoteConfigMockDefaults = {
```
fadi-george, line 24: To allow for passing custom properties

__test__/support/environment/TestContext.ts
```diff
@@ -169,177 +170,178 @@ export default class TestContext {
...
-        origin: 'https://example.com',
-        subdomain: undefined,
-        http_use_onesignal_com: false,
-        autoResubscribe: false,
-        staticPrompts: {
-          native: {
-            enabled: false,
-            autoPrompt: false,
+    } else {
+      remoteConfigMockDefaults = {
```
fadi-george, line 174: no change with these proprties

sherwinski (approved):
🚀

PR #1227, opened 2025-02-20
Service Worker Tests Part 3

fadi-george:
# Description
## 1 Line Summary
Improve test coverage of service worker from 73% - 86%

## Details
- Adds tests for the worker message events
- Deletes some unneeded mock service worker files
- Updates the isSecureContext check in `checkUnsupportedSubdomain` that's called when downloading the server config
  - This is related the AMP setup issue here: https://github.com/OneSignal/OneSignal-WordPress-Plugin/issues/351?email_source=slack
- Had to wrap workerMessenger.listen with setTimeout since the `listen` fires right away when importing the file for the test
- Had to fix some api usage typings like the ones typed with OneSignalApiBaseResponse to fix build issues

### Amp Issue
I was able to replicate the amp issue:
Screenshot 2025-02-20 at 4 12 29 PM

I built the sdk locally, copied it over to my amp setup, and pointed the service-worker-url to my local sdk.
I was able to subscribe. 
Screenshot 2025-02-20 at 4 14 00 PM

# Systems Affected
   - [x] WebSDK
   - [ ] Backend
   - [ ] Dashboard

# Validation
### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [ ] No Typescript warnings

Other:

### Checklist

---

## Related Tickets

---

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Website-SDK/1227)


src/shared/managers/SdkEnvironment.ts
```diff
@@ -74,15 +74,11 @@ export default class SdkEnvironment {
...
    */
   public static getWindowEnv(): WindowEnvironmentKind {
-    if (typeof window === 'undefined') {
-      if (
-        typeof self !== 'undefined' &&
-        typeof ServiceWorkerGlobalScope !== 'undefined'
-      ) {
-        return WindowEnvironmentKind.ServiceWorker;
-      } else {
-        throw Error('OneSignalSDK: Unsupported JS runtime!');
```
sherwinski, line 84: Is there a specific reason for removing this condition?

src/sw/serviceWorker/ServiceWorker.ts
```diff
@@ -410,12 +411,6 @@ export class ServiceWorker {
         )
         .catch((e) => {
           Log.debug('Failed to display a notification:', e);
-          if (ServiceWorker.UNSUBSCRIBED_FROM_NOTIFICATIONS) {
-            Log.debug(
-              'Because we have just unsubscribed from notifications, we will not show anything.',
-            );
-            return undefined;
-          }
```
sherwinski, line 418: Is there a reason for removing this log?

sherwinski (commented):
Looks like there's a handful of changes within the `src` files: could you add a little more detail in the PR description about those for tracking purposes? You don't need to document changes due to refactoring or linting.

__test__/support/environment/TestContext.ts
```diff
@@ -449,7 +449,7 @@ export default class TestContext {
       persistNotification: false,
       webhooks: {
         cors: true,
-        'notification.displayed':
+        'notification.willDisplay':
```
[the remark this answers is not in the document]
fadi-george: Thanks.

__test__/support/environment/TestEnvironmentHelpers.ts
```diff
@@ -112,12 +112,6 @@ export async function stubDomEnvironment(config: TestEnvironmentConfig) {
   });
 
   const windowDef = dom.window;
-  // Node has its own console; overwriting it will cause issues
-  delete (windowDef as any)['console'];
-  (windowDef as any).navigator.serviceWorker =
-    new MockServiceWorkerContainerWithAPIBan();
-  // (windowDef as any).TextEncoder = TextEncoder;
-  // (windowDef as any).TextDecoder = TextDecoder;
```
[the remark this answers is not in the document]
fadi-george: It's not needed currently for the vitest tests. window.TextEncoder and window.TextDecoder are provided by jsdom.

src/core/requestService/IdentityRequests.ts
```diff
@@ -65,7 +64,7 @@ export default class IdentityRequests {
   }
 
   private static _processIdentityResponse(
-    response?: OneSignalApiBaseResponse,
+    response?: Awaited<ReturnType<typeof RequestService.deleteAlias>>,
```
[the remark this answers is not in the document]
fadi-george: It would be equivalent to 
```
    response?: OneSignalApiBaseResponse<{
    identity: SupportedIdentity;
}>,
```

src/shared/managers/SdkEnvironment.ts
fadi-george: To make it easier to test since ServiceWorkerGlobalScope is only available in service worker context. So technically we don't need the self check either. 
```
if (
      typeof self !== 'undefined' &&
      typeof ServiceWorkerGlobalScope !== 'undefined'
    ) {
      return WindowEnvironmentKind.ServiceWorker;
    }
    if (typeof window === 'undefined') {
      throw Error('OneSignalSDK: Unsupported JS runtime!');
    }
    return WindowEnvironmentKind.Host;
    ```

File: src/sw/serviceWorker/ServiceWorker.ts (hunk 2 above)
Comment (fadi-george) [reply src/sw/serviceWorker/ServiceWorker.ts:418]: UNSUBSCRIBED_FROM_NOTIFICATIONS is not set anywhere and seems to be legacy code. And it doesn't do anything too useful.

**Comment by fadi-george** [comment]
> Looks like there's a handful of changes within the `src` files: could you add a little more detail in the PR description about those for tracking purposes? You don't need to document changes due to refactoring or linting.

There were some build issues so had to fix the typings for some api calls.

File: src/shared/helpers/ConfigHelper.ts [hunk 6]
```diff
@@ -75,7 +75,7 @@ export class ConfigHelper {
   // The os.tc domain feature is no longer supported in v16, so throw if the
   // OneSignal app is still configured this way after they migrated from v15.
   private static checkUnsupportedSubdomain(appConfig: AppConfig): void {
-    const isHttp = !window.isSecureContext;
+    const isHttp = !self.isSecureContext;
```
fadi-george, line 78: Main fix for the AMP issue.

sherwinski (approved):

PR #1229, opened 2025-02-24
Remove Old Ava Tests

fadi-george:
# Description
## 1 Line Summary
- Remove old ava tests and its related packages

## Details
- Removes ava and other unused packages
- Remove init helper polyfillSafariFetch logic since fetch has been widely [available](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API) since 2017
- Removes old ava tests folder, can have these be ported over later
- Ported over simpler ava tests to vitest

# Systems Affected
   - [x] WebSDK
   - [ ] Backend
   - [ ] Dashboard

# Validation
### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [ ] No Typescript warnings

Other:

### Checklist

---

## Related Tickets

---

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Website-SDK/1229)


src/shared/utils/utils.ts
```diff
@@ -194,13 +193,6 @@ export function nothing(): Promise<any> {
   return Promise.resolve();
 }
 
-export function timeoutPromise(
```
fadi-george, line 197: unused

__test__/support/environment/TestEnvironmentHelpers.ts
```diff
@@ -1,22 +1,22 @@
-import Database from '../../../src/shared/services/Database';
-import Random from '../utils/Random';
-import Emitter from '../../../src/shared/libraries/Emitter';
-import { TestEnvironmentConfig } from './TestEnvironment';
-import MockNotification from '../mocks/MockNotification';
+import bowser from 'bowser';
```
fadi-george, line 1: auto sorted imports

src/onesignal/NotificationsNamespace.test.ts
```diff
@@ -0,0 +1,44 @@
+import { TestEnvironment } from '__test__/support/environment/TestEnvironment';
```
fadi-george, line 1: Migrated some tests from ava

jkasten2 (changes requested):
~We need to keep `package-lock.json`, so we keep the locking so have reproducible builds.~
* My mistake, I see the file is still there, lines were just deleted.

jkasten2 (approved):

PR #1231, opened 2025-02-27
New Vite Setup

fadi-george:
# Description
## 1 Line Summary
- Removes webpack config in favor of a simpler Vite config setup

## Details
- Removes all webpack packages, configs, and references
- Adds new vite setup with an index.html for testing locally instead of needing to build
  - Adds .env to set a custom app_id for testing locally
  - Remove various build scripts and logic for build flags to instead rely on env vars like BUILD_ORIGIN
- Replaced bundlesize2 with more actively maintained size-limit package
- Add vite-bundle-analyzer to help with analyzing bundles and store them to build/stats
- Fix sass deprecations for stylesheet (e.g. @import -> @use)
- Updates tsconfig to use es2022 to use iffe format

# Systems Affected
   - [x] WebSDK
   - [ ] Backend
   - [ ] Dashboard

# Validation
### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [ ] No Typescript warnings

Other:

### Checklist

---

## Related Tickets

---

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Website-SDK/1231)


src/entries/stylesheet.scss
```diff
@@ -1,5 +1,5 @@
-@import '../page/stylesheets/reset';
-@import '../page/stylesheets/bell';
-@import '../page/stylesheets/customlink';
-@import '../page/stylesheets/helpers';
-@import '../page/stylesheets/slidedown';
+@use '../page/stylesheets/reset';
```
fadi-george, line 1: used sass migrator

index.html
```diff
@@ -0,0 +1,29 @@
+<!doctype html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <link rel="icon" type="image/svg+xml" href="/vite.svg" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <title>Vite + React + TS</title>
+    <script>
+      window.OneSignalDeferred = [];
```
jkasten2, line 9: Remove this `window.OneSignalDeferred = [];` line, the line below `window.OneSignalDeferred = window.OneSignalDeferred || [];` is all the is needed.

index.html
```diff
@@ -0,0 +1,29 @@
+<!doctype html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <link rel="icon" type="image/svg+xml" href="/vite.svg" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <title>Vite + React + TS</title>
```
jkasten2, line 7: Change title to state this is a OneSignal dev env.

index.html
```diff
@@ -0,0 +1,29 @@
+<!doctype html>
```
jkasten2, line 1: Why this new `index.html`?

If this has benefits for those how develop the SDK can we add comments here to explain that? And that also call out `preview/index.html` is more production like test if that is needed.

index.html
```diff
@@ -0,0 +1,29 @@
...
+
+      OneSignalDeferred.push(async function (OneSignal) {
+        await OneSignal.init({
+          appId,
+        });
+      });
+    </script>
+  </head>
+  <body>
+    <script type="module" src="/src/entries/pageSdkInit.ts"></script>
```
jkasten2, line 26: Does this allow testing the SDK without have to do a full build?

typings/globals/missing.d.ts
```diff
@@ -1,51 +0,0 @@
...
-
-interface Window {
-  safari?: {
-    pushNotification?: SafariRemoteNotification
-  };
-}
-
-/**
- * END: window.safari definition
- */
```
jkasten2, line 30: These Safari types are used in SDK code base so we should keep these. I don't believe they are defined in the standard TypeScript types.

jkasten2 (changes requested):

.gitignore
```diff
@@ -1,23 +1,17 @@
...
-dist/Dev-OneSignalSDK.page.js
-dist/Dev-OneSignalSDK.page.js.map
 .vscode/**
 !.vscode/launch.json
 
 coverage
-.nyc_output
 
-ngrok_last_url
-ngrok.log
```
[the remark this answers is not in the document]
fadi-george: Yeah mb ill add it back since I originally removed ngrok

index.html
fadi-george: Yes this will allow for hot-reloading and easier debugging.

index.html
fadi-george: Done

index.html
fadi-george: Done

index.html
fadi-george: Added a description in the readme and this file.

index.html
fadi-george: Can also debug the service worker.
Screenshot 2025-03-04 at 2 21 26 PM

typings/globals/missing.d.ts
fadi-george: Added back.

jkasten2 (approved):

PR #1232, opened 2025-03-06
More Vite Setup Changes

fadi-george:
# Description
## Details

# Systems Affected
   - [ ] WebSDK
   - [ ] Backend
   - [ ] Dashboard

# Validation
### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [ ] No Typescript warnings

Other:

### Checklist

---

## Related Tickets

---


src/global.d.ts
```diff
@@ -31,19 +31,3 @@ interface Window {
 
 declare let OneSignal: any;
 declare let OneSignalDeferred: any;
-
-// These __*__ variables are defined from Webpack to change resulting JS
-declare let __VERSION__: string;
```
fadi-george, line 36: already in vite-end.d.ts

src/page/services/DynamicResourceLoader.ts
```diff
@@ -38,9 +37,7 @@ export class DynamicResourceLoader {
     const cssFileForEnv = SdkEnvironment.getOneSignalCssFileName();
     return this.loadIfNew(
       ResourceType.Stylesheet,
```
fadi-george, line 39: @jkasten2 updated to use package json version instead

fadi-george: @sherwinski

src/page/services/DynamicResourceLoader.test.ts
```diff
@@ -29,7 +29,7 @@ describe('DynamicResourceLoader', () => {
       'link[rel="stylesheet"]',
     );
 
-    const url = `${cssURL}?v=2`;
+    const url = `${cssURL}?v=1`;
```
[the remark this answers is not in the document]
sherwinski: Makes sense, figured I'd ask just to be sure.

sherwinski (approved):

src/page/services/DynamicResourceLoader.ts
jkasten2: I don't see any references the css file at all on the backend, so this version should be safe to change

PR #1233, opened 2025-03-06
Prevent race with switching users & handleOnFocus

jkasten2:
# Description
## 1 Line Summary
Prevent bad states & errors with SessionManager.handleOnFocus if the site's focus changes while a login or logout is being processed.

## Details
If the site focus changes when the login or logout fn is still running we can get errors with missing ids. To prevent this always await to ensure they are completed.

The most common way this can happen is if you call window.prompt() or window.alert() with 100ms before or after calling OneSignal.login()

# Validation
### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [X] No Typescript warnings

Other:

### Checklist

---

## Related Tickets

---

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Website-SDK/1233)

### DIFF
--- a/src/page/managers/LoginManager.ts
+++ b/src/page/managers/LoginManager.ts
@@ -17,19 +17,17 @@ import LocalStorage from '../../shared/utils/LocalStorage';
 import { logMethodCall } from '../../shared/utils/utils';
 
 export default class LoginManager {
+  // Other internal classes should await on this if they access users
   static switchingUsersPromise: Promise<void> = Promise.resolve();
 
   static async login(externalId: string, token?: string): Promise<void> {
-    try {
-      // Prevents overlapting login/logout calls
-      await this.switchingUsersPromise;
-      this.switchingUsersPromise = this._login(externalId, token);
-    } catch (e) {
-      this.switchingUsersPromise = Promise.reject(e);
-    }
+    await (this.switchingUsersPromise = LoginManager._login(externalId, token));
   }
 
-  static async _login(externalId: string, token?: string): Promise<void> {
+  private static async _login(
+    externalId: string,
+    token?: string,
+  ): Promise<void> {
     const consentRequired = LocalStorage.getConsentRequired();
     const consentGiven = await Database.getConsentGiven();
 
@@ -142,16 +140,10 @@ export default class LoginManager {
   }
 
   static async logout(): Promise<void> {
-    try {
-      // Prevents overlapting login/logout calls
-      await this.switchingUsersPromise;
-      this.switchingUsersPromise = this._logout();
-    } catch (e) {
-      this.switchingUsersPromise = Promise.reject(e);
-    }
+    await (this.switchingUsersPromise = LoginManager._logout());
   }
 
-  static async _logout(): Promise<void> {
+  private static async _logout(): Promise<void> {
     // check if user is already logged out
     const identityModel = OneSignal.coreDirector?.getIdentityModel();
     if (
--- a/src/page/managers/LoginManager.ts
+++ b/src/page/managers/LoginManager.ts
@@ -17,7 +17,19 @@ import LocalStorage from '../../shared/utils/LocalStorage';
 import { logMethodCall } from '../../shared/utils/utils';
 
 export default class LoginManager {
+  static switchingUsersPromise: Promise<void> = Promise.resolve();
+
   static async login(externalId: string, token?: string): Promise<void> {
+    try {
+      // Prevents overlapting login/logout calls
+      await this.switchingUsersPromise;
+      this.switchingUsersPromise = this._login(externalId, token);
+    } catch (e) {
+      this.switchingUsersPromise = Promise.reject(e);
+    }
+  }
+
+  static async _login(externalId: string, token?: string): Promise<void> {
     const consentRequired = LocalStorage.getConsentRequired();
     const consentGiven = await Database.getConsentGiven();
 
@@ -130,6 +142,16 @@ export default class LoginManager {
   }
 
   static async logout(): Promise<void> {
+    try {
+      // Prevents overlapting login/logout calls
+      await this.switchingUsersPromise;
+      this.switchingUsersPromise = this._logout();
+    } catch (e) {
+      this.switchingUsersPromise = Promise.reject(e);
+    }
+  }
+
+  static async _logout(): Promise<void> {
     // check if user is already logged out
     const identityModel = OneSignal.coreDirector?.getIdentityModel();
     if (
--- a/src/shared/managers/sessionManager/SessionManager.test.ts
+++ b/src/shared/managers/sessionManager/SessionManager.test.ts
@@ -1,45 +1,27 @@
-import { SessionManager } from './SessionManager';
 import { TestEnvironment } from '__test__/support/environment/TestEnvironment';
+import { SessionManager } from './SessionManager';
 
-import { stub } from '__test__/support/helpers/general';
-import { PropertiesExecutor } from 'src/core/executors/PropertiesExecutor';
-import { IdentityExecutor } from 'src/core/executors/IdentityExecutor';
-import { SubscriptionExecutor } from 'src/core/executors/SubscriptionExecutor';
-
-import { setupLoginStubs } from '__test__/support/helpers/login';
-import LoginManager from 'src/page/managers/LoginManager';
 import {
   DUMMY_EXTERNAL_ID,
   DUMMY_ONESIGNAL_ID,
 } from '__test__/support/constants';
-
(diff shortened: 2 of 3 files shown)


fadi-george:
Let's add some tests

src/page/managers/LoginManager.ts
```diff
@@ -17,7 +17,19 @@ import LocalStorage from '../../shared/utils/LocalStorage';
...
+  static switchingUsersPromise: Promise<void> = Promise.resolve();
+
   static async login(externalId: string, token?: string): Promise<void> {
+    try {
+      // Prevents overlapting login/logout calls
+      await this.switchingUsersPromise;
+      this.switchingUsersPromise = this._login(externalId, token);
+    } catch (e) {
+      this.switchingUsersPromise = Promise.reject(e);
+    }
```
fadi-george, line 29: Wont this loop ie if switchingUsersPromise fails one time, now its always promise reject. Maybe it should be
```
 try {
      // Prevents overlapting login/logout calls
      await this.switchingUsersPromise;
    } catch (e) {
      Log...
    }
    this.switchingUsersPromise = this._login(externalId, token);
```

jkasten2: I addressed this in commit 7d3e033f04b83f275a282b00ca62d2a6c1169755

src/page/managers/LoginManager.ts
```diff
@@ -130,6 +142,16 @@ export default class LoginManager {
   }
 
   static async logout(): Promise<void> {
+    try {
```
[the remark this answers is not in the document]
jkasten2: I addressed this in commit 7d3e033f04b83f275a282b00ca62d2a6c1169755

jkasten2:
@fadi-george I addressed all your feedback

src/shared/managers/sessionManager/SessionManager.test.ts
```diff
@@ -0,0 +1,77 @@
...
+
+vi.mock('../../../src/shared/libraries/Log');
+stub(PropertiesExecutor.prototype, 'getOperationsFromCache', []);
+stub(IdentityExecutor.prototype, 'getOperationsFromCache', []);
+stub(SubscriptionExecutor.prototype, 'getOperationsFromCache', []);
+
+describe('SessionManager', () => {
+  afterEach(() => {
+    vi.resetAllMocks();
+  });
```
fadi-george, line 24: dont need

src/shared/managers/sessionManager/SessionManager.test.ts
```diff
@@ -0,0 +1,77 @@
...
+import LoginManager from 'src/page/managers/LoginManager';
+import {
+  DUMMY_EXTERNAL_ID,
+  DUMMY_ONESIGNAL_ID,
+} from '__test__/support/constants';
+
+vi.mock('../../../src/shared/libraries/Log');
+stub(PropertiesExecutor.prototype, 'getOperationsFromCache', []);
+stub(IdentityExecutor.prototype, 'getOperationsFromCache', []);
+stub(SubscriptionExecutor.prototype, 'getOperationsFromCache', []);
```
fadi-george, line 19: These dont seem needed if we have `useMockIdentityModel: true` for TestEnvironmentInitialize

src/shared/managers/sessionManager/SessionManager.test.ts
```diff
@@ -0,0 +1,77 @@
...
+  });
+
+  describe('Switching Users', () => {
+    beforeEach(async () => {
+      await TestEnvironment.initialize({
+        useMockIdentityModel: true,
+        useMockPushSubscriptionModel: true,
+      });
+      setupLoginStubs();
+      stub(
```
fadi-george, line 33: The stub methods are deprecated and we recommended using vi.spyOn instead e.g.
```
   vi.spyOn(LoginManager, 'identifyOrUpsertUser').mockResolvedValue({
        identity: {
          external_id: DUMMY_EXTERNAL_ID,
          onesignal_id: DUMMY_ONESIGNAL_ID,
        },
      });
```

src/shared/managers/sessionManager/SessionManager.test.ts
```diff
@@ -0,0 +1,77 @@
...
+describe('SessionManager', () => {
+  afterEach(() => {
+    vi.resetAllMocks();
+  });
+
+  describe('Switching Users', () => {
+    beforeEach(async () => {
+      await TestEnvironment.initialize({
+        useMockIdentityModel: true,
+        useMockPushSubscriptionModel: true,
```
fadi-george, line 30: doesnt seem like you need `useMockPushSubscriptionModel`

src/shared/managers/sessionManager/SessionManager.test.ts
```diff
@@ -0,0 +1,77 @@
...
+import { SubscriptionExecutor } from 'src/core/executors/SubscriptionExecutor';
+
+import { setupLoginStubs } from '__test__/support/helpers/login';
+import LoginManager from 'src/page/managers/LoginManager';
+import {
+  DUMMY_EXTERNAL_ID,
+  DUMMY_ONESIGNAL_ID,
+} from '__test__/support/constants';
+
+vi.mock('../../../src/shared/libraries/Log');
```
fadi-george, line 16: doesnt seem needed for thistest

src/shared/managers/sessionManager/SessionManager.test.ts
```diff
@@ -0,0 +1,77 @@
...
+        return 'login';
+      })();
+
+      const sessionManager = new SessionManager(OneSignal.context);
+      const sessionPromise = (async function () {
+        await sessionManager.handleOnFocus(new Event('{}'));
+        return 'session';
+      })();
+
+      const winner = await Promise.race([loginPromise, sessionPromise]);
```
fadi-george, line 57: This is pretty unique, I have seen this used that often but makes sense for usecase. I would just the result different e.g. `winner` -> `firstResult` or `firstCompleted`

src/shared/managers/sessionManager/SessionManager.test.ts
```diff
@@ -0,0 +1,77 @@
...
+        return 'logout';
+      })();
+
+      const sessionManager = new SessionManager(OneSignal.context);
+      const sessionPromise = (async function () {
+        await sessionManager.handleOnFocus(new Event('{}'));
+        return 'session';
+      })();
+
+      const winner = await Promise.race([loginPromise, sessionPromise]);
```
fadi-george, line 73: See other comment about naming.

src/shared/managers/sessionManager/SessionManager.ts
```diff
@@ -16,6 +16,7 @@ import { RequestService } from '../../../core/requestService/RequestService';
 import AliasPair from '../../../core/requestService/AliasPair';
 import { UpdateUserPayload } from '../../../core/requestService/UpdateUserPayload';
 import Utils from '../../../shared/context/Utils';
+import LoginManager from '../../../page/managers/LoginManager';
```
fadi-george, line 19: nit: minor thing but this setting would help organize imports:
```
{
  "editor.codeActionsOnSave": {
    "source.fixAll.eslint": "explicit",
    "source.organizeImports": "always"
  },
}
```
which we could update gitignore to allow sharing these settings.

jkasten2: Addressed in commit 457f8235698c2f006e8329399d0c35520f2bb3fc
* I made this it's own commit since it moved around all the imports in the file and it isn't really a change related to the issue being fixed in this PR.

src/shared/managers/sessionManager/SessionManager.test.ts
jkasten2: I removed this

src/shared/managers/sessionManager/SessionManager.test.ts
jkasten2: I removed these

src/shared/managers/sessionManager/SessionManager.test.ts
jkasten2: I removed this.

src/shared/managers/sessionManager/SessionManager.test.ts
jkasten2: I updated this.

src/shared/managers/sessionManager/SessionManager.test.ts
jkasten2: See my other follow up comment above.

fadi-george (approved):

Issue #1210, opened 2024-11-19
[Bug]: Uncaught TypeError: e is not iterable at o.processOneSignalDeferredArray


Jonathan-Ha:
![Screenshot 2024-11-19 104542](https://github.com/user-attachments/assets/99ab1e9d-0cbb-4bc0-82f4-19e4c9ad8c08)
![Screenshot 2024-11-19 104614](https://github.com/user-attachments/assets/6109908b-73a8-4d13-bbc5-9ef07ca4ae9b)

sherwinski:
Hi @Jonathan-Ha,

Could you please share/paste the code snippet of where you have the SDK init code defined?

Reference: https://documentation.onesignal.com/docs/web-sdk-setup#7-add-code-to-site

vdr-bcg-robinweston:
We have started seeing this too

Here's our initialisation code - part of a React component:

```
useEffect(() => {
    if (typeof window !== undefined) {
      window.OneSignalDeferred = window?.OneSignalDeferred || [];
      window.OneSignalDeferred.push(async function (OneSignal) {
        await OneSignal.init({
          appId: env('NEXT_PUBLIC_ONESIGNAL_APP_ID') || getEnvironmentVariable('ONESIGNAL_APP_ID'),
        });
        OneSignal.User.addTag('user_id', userId);
      });
    }
  }, []);
  ```
  
  Here is the script tag
  ```
  <script src="https://cdn.onesignal.com/sdks/web/v16/OneSignalSDK.page.js" defer></script>
  ```

**Comment by jihad28** [comment]
I have the same issue, here is my code

``` js
window.OneSignalDeferred = window.OneSignalDeferred || [];
window.OneSignalDeferred.push(async function (OneSignal) {
    await OneSignal.init({
        appId: import.meta.env.VITE_ONESIGNAL_APP_ID,
        serviceWorkerPath: "service-worker.js",
        promptOptions: {
            slidedown: {
                prompts: [{
                    type: "push",
                    autoPrompt: false,
                    text: {
                        actionMessage: LA.send_notification_text,
                        acceptButton: LA.yes,
                        cancelButton: LA.no
                    }
                }]
            }
        }
    });
});
```

and the script tag
``` html
<script src="https://cdn.onesignal.com/sdks/web/v16/OneSignalSDK.page.js" defer></script>
```

[...]

PR #1240, opened 2025-03-21
Update page sdk init to handle delayed setting of window.OneSignalDeferred

fadi-george:
# Description
## 1 Line Summary
- Update logic for processOneSignalDeferredArray to address #1210 

## Details
If `window.OneSignalDeferred` is not set in time before processOneSignalDeferredArray is called, it will lead to an `is not iterable` error.
- So added a nullish coalescing check before the process call
- Removed OneSignalDeferred class and just update the push property of `window.OneSignalDeferred` for better typings / simplicity  
- Improved global typings for OneSignal and OneSignalDeferred

# Systems Affected
   - [x] WebSDK
   - [ ] Backend
   - [ ] Dashboard

# Validation
### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [ ] No Typescript warnings

Other:

### Checklist

---

## Related Tickets

---

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Website-SDK/1240)


src/global.d.ts
```diff
@@ -19,15 +20,17 @@ interface SafariRemoteNotification {
   ): void;
 }
 
-interface Window {
-  safari?: {
-    pushNotification?: SafariRemoteNotification;
-  };
-}
```
sherwinski, line 26: what's the reason for removing this?

sherwinski (approved):

src/global.d.ts
fadi-george: I moved it under global see below

PR #1265, opened 2025-04-11
Web SDK Refactor: Add Tests for OperationRepo

fadi-george:
# Description
Continuation of the [Web SDK Refactor project](https://docs.google.com/document/d/1JsbmxVM_n6K9nRXwbJf6mQr5h88qy_vIEiBGY0Ub_lE/edit?tab=t.0#heading=h.ojv84gskof17). This pr is focused on adding tests for the OperationRepo refactor.

## Details
- adds unit tests for the operation repo class
- added some methods like `isPaused` and `records` to make it easier to test
- changed `executeOperations` and `getGroupableOperations` to public to make it easier to test
- export delay values like `OP_REPO_*` for easier testing 
- removed `enqueueAndWait` as there doesnt need to be a need for it

# Systems Affected
   - [x] WebSDK
   - [ ] Backend
   - [ ] Dashboard

# Validation
### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [ ] No Typescript warnings

Other:

### Checklist

---

## Related Tickets

---

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Website-SDK/1265)

---
  - https://app.asana.com/0/0/1208801496090558

### DIFF
--- a/src/core/operationRepo/OperationRepo.test.ts
+++ b/src/core/operationRepo/OperationRepo.test.ts
@@ -1,4 +1,5 @@
-import { describe, expect, vi } from 'vitest';
+import Log from 'src/shared/libraries/Log';
+import { describe, expect, Mock, vi } from 'vitest';
 import {
   ExecutionResult,
   GroupComparisonType,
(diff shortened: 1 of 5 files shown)


src/core/operationRepo/OperationRepo.test.ts
```diff
@@ -0,0 +1,443 @@
...
+        operations: additionalOps,
+      });
+
+      const opRepo = getNewOpRepo();
+      const modelAddSpy = vi.spyOn(mockOperationModelStore, 'add');
+      const modelRemoveSpy = vi.spyOn(mockOperationModelStore, 'remove');
+
+      opRepo.enqueue(mockOperation);
+      expect(modelAddSpy).toHaveBeenCalledWith(mockOperation);
+      modelAddSpy.mockClear();
```
sherwinski, line 191: Is it possible to test that these spys are called with the IDs fro L176-177?

sherwinski (approved):

src/core/operationRepo/OperationRepo.test.ts
fadi-george: The ids get changed to something random in the OperationRepo.  So by the time we check lines 200, 201, theyre  different from 3 and 4. But the id mutates the original data, so the test will pass.

PR #1267, opened 2025-04-11
Web SDK Refactor: Add Model Class

fadi-george:
# Description
Continuation of the [Web SDK Refactor project](https://docs.google.com/document/d/1JsbmxVM_n6K9nRXwbJf6mQr5h88qy_vIEiBGY0Ub_lE/edit?tab=t.0#heading=h.ojv84gskof17). This pr is focuses on porting the follow classes/types from the Android SDK: [Model](https://github.com/OneSignal/OneSignal-Android-SDK/blob/5249d8d78bbec6fb783f3f377bea29a55fabb797/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/modeling/Model.kt) , [EventProducer](https://github.com/OneSignal/OneSignal-Android-SDK/blob/5249d8d78bbec6fb783f3f377bea29a55fabb797/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/events/EventProducer.kt#L13) , [IEventNotifier](https://github.com/OneSignal/OneSignal-Android-SDK/blob/5249d8d78bbec6fb783f3f377bea29a55fabb797/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/common/events/IEventNotifier.kt#L14) , [IPreferencesService](https://github.com/OneSignal/OneSignal-Android-SDK/blob/5249d8d78bbec6fb783f3f377bea29a55fabb797/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/core/internal/preferences/IPreferencesService.kt#L9) 

# Details
- adds new Model class based on the Android SDK one
- added two new methods to the model class ( `getProperty` and `setProperty` that take in a generic)
- removed extraneous methods `getEnumProperty`, `getMapModelProperty`, etc. since we can just use a generic when getting or setting a property
- moved `ModelChangedArgs` to same place as Model otherwise they would import eachother
- did not implement `PreferenceStores` or `PreferenceOneSignalKeys` as we dont need multiple stores e.g. OneSignal vs PlayerPurchases or other legacy info

# Systems Affected
   - [x] WebSDK
   - [ ] Backend
   - [ ] Dashboard

# Validation
### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [ ] No Typescript warnings

Other:

### Checklist

---

## Related Tickets

---

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Website-SDK/1267)

---
  - https://app.asana.com/0/0/1208801496090558


src/core/models/Model.ts
```diff
@@ -0,0 +1,278 @@
...
+   * class of Model for this property is determined by the implementor and should depend on
+   * the property provided.
+   *
+   * @param {string} property The property that is to contain the Model created by this method.
+   * @param {object} jsonObject The JSON object that the Model will be created/initialized from.
+   *
+   * @return The created Model, or null if the property should not be set.
+   */
+  protected createModelForProperty(): Model | null {
+    return null;
```
sherwinski, line 173: You'll need to double check this, but I believe this line should return either the Model if it exists or `null`. I understand that in the Java implementation `createModelForProperty` is meant to be overriden but I'm not completely sure if that translates 1-to-1 to typescript.

sherwinski: Also, it looks like the jsdoc mentions two parameters but none are passed in.

src/core/models/Model.ts
```diff
@@ -0,0 +1,278 @@
...
+   * indicating the property value should be set to an Array. The specific concrete class
+   * inside the Array for this property is determined by the implementor and should depend
+   * on the property provided.
+   *
+   * @param {string} property The property that is to contain the Array created by this method.
+   * @param {unknown[]} jsonArray The JSON array that the Array will be created/initialized from.
+   *
+   * @return The created Array, or null if the property should not be set.
+   */
+  protected createListForProperty(): unknown[] | null {
```
sherwinski, line 187: jsdocs mention two parameters but none are passed in.

src/types/models.ts
```diff
@@ -0,0 +1,77 @@
+import type { Model, ModelChangedArgs } from 'src/core/models/Model';
+import type { IEventNotifier } from './events';
+
+// MODEL STORE
```
sherwinski, line 4: unnecessary comment?

sherwinski (changes requested):

src/core/models/Model.ts
```diff
@@ -0,0 +1,278 @@
+import { EventProducer } from 'src/shared/helpers/EventProducer';
+import { IEventNotifier } from 'src/types/events';
+import { ModelChangeTags } from 'src/types/models';
+
+// MODEL CHANGED
```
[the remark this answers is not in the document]
fadi-george: Ill remove

src/core/models/Model.ts
fadi-george: Typescript would complain they are unused so I figured jsdoc would be fine here.

src/core/models/Model.ts
fadi-george: Typescript would complain its unused. Either I add ts-ignores or leave it empty.

fadi-george: Actually I tested in Typescript Playground, so it seems like I'll need to add them back even though they are unused.

src/core/modelRepo/Model.ts
```diff
@@ -0,0 +1,281 @@
+import { EventProducer } from 'src/shared/helpers/EventProducer';
+import { IEventNotifier } from 'src/types/events';
+import { ModelChangeTags } from 'src/types/models';
+
+// MODEL CHANGED
```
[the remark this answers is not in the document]
fadi-george: Removed

src/types/models.ts
fadi-george: Ill remove.

src/core/models/Model.ts
fadi-george: The override keyword is optional and is more compile time check.

sherwinski (approved):

PR #1279, opened 2025-04-21
Web SDK Refactor: Login Operation Executor Tests

fadi-george:
# Description
## 1 Line Summary

## Details
- add tests for Login Operation Executor
- changed from Object.assign to variable assignment for `subscriptions` variable since delete subscription would need to remove properties and Object.assign will fail to do that
- for create subscription operation, made it more like the Android SDK implementation with passing just `enabled`, `notifications_types` (status), `type`, `token` (address)
- for update subscription operation, made it pass along `notification_types` & `token`
- added `BaseFullSubscriptionOperation` which has more subscription properties used for CreateSubscriptionOperation and UpdateSubscriptionOperation

# Systems Affected
   - [x] WebSDK
   - [ ] Backend
   - [ ] Dashboard

# Validation
### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [ ] No Typescript warnings

Other:

### Checklist

---

## Related Tickets

---

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Website-SDK/1279)

### DIFF
--- a/src/core/operations/TransferSubscriptionOperation.ts
+++ b/src/core/operations/TransferSubscriptionOperation.ts
@@ -1,19 +1,11 @@
-import { IDManager } from 'src/shared/managers/IDManager';
 import { OPERATION_NAME } from '../constants';
-import {
-  GroupComparisonType,
-  GroupComparisonValue,
-  Operation,
-} from './Operation';
-
-type TransferOp = {
-  subscriptionId: string;
-};
+import { BaseSubscriptionOperation } from './BaseSubscriptionOperation';
+import { GroupComparisonType, GroupComparisonValue } from './Operation';
 
 /**
  * An Operation to transfer a subscription to a new owner on the OneSignal backend.
  */
-export class TransferSubscriptionOperation extends Operation<TransferOp> {
+export class TransferSubscriptionOperation extends BaseSubscriptionOperation {
   constructor(appId: string, onesignalId: string, subscriptionId: string);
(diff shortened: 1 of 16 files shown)


src/core/operations/TransferSubscriptionOperation.ts
```diff
@@ -6,6 +6,7 @@ import { GroupComparisonType, GroupComparisonValue } from './Operation';
  * An Operation to transfer a subscription to a new owner on the OneSignal backend.
  */
 export class TransferSubscriptionOperation extends BaseSubscriptionOperation {
+  constructor(appId: string, onesignalId: string, subscriptionId: string);
```
sherwinski, line 9: I'm not sure what adding this line accomplishes from a user's perspective. From how I'm reading it, you can either call the constructor with all parameters, no parameters, or some combination of any of them. Is that not the same as all parameters being optional?

fadi-george: Its useful in the case you try to do something like this
``` 
new TransferSubscriptionOperation(APP_ID)
```
where now Typescript will complain that youre missing parameters.

sherwinski: So it's more like: include all parameters or include none?

fadi-george: Yes

sherwinski (approved):

PR #1284, opened 2025-04-23
Web SDK Refactor: Update Core Module 

fadi-george:
# Description
## 1 Line Summary

## Details
- Implements [RebuildUserService](https://github.com/OneSignal/OneSignal-Android-SDK/blob/5.1.31/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/builduser/impl/RebuildUserService.kt) similar to Android SDK
- Update CoreModule logic to use the new executors
- Delete some unused files/classes
- Removed usage preference stores since I added logic ModelStore to store data in IndexedDB similar to how was it done before

# Systems Affected
   - [x] WebSDK
   - [ ] Backend
   - [ ] Dashboard

# Validation
### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [ ] No Typescript warnings

Other:

### Checklist

---

## Related Tickets

---

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Website-SDK/1284)


src/shared/services/IndexedDb.ts
```diff
@@ -346,6 +353,7 @@ export default class IndexedDb {
           reject(e);
         };
       } catch (e) {
+        console.trace('uhh', { table, method, keyOrValue });
```
sherwinski, line 356: Let's either remove this or make a note to remove it later

sherwinski (approved):

src/shared/services/IndexedDb.ts
fadi-george: Thanks, removed

PR #1285, opened 2025-04-24
Web SDK Refactor: Core Module & User Director

fadi-george:
# Description
## 1 Line Summary
Continuation of the [Web SDK Refactor project](https://docs.google.com/document/d/1JsbmxVM_n6K9nRXwbJf6mQr5h88qy_vIEiBGY0Ub_lE/edit?tab=t.0#heading=h.ojv84gskof17). Tests will be fixed later.

## Details
- Removed legacy code (OSModel, OSModelStore, modelCache, modelRepo, operationCache & tests)
- Exposed ` subscriptionModelStore`, `identityModelStore`, `propertiesModelStore` in CoreModule
- For `generatePushSubscriptionModel`, had it return the SubscriptionModel, it doesnt have a external_id or onesignal_id property so may need to be revisited later 
- Removed `_hydrateSubscriptions` as there a similar issue with not having external_id or onesignal_id for subscription models
- Removed `getModelByTypeAndId` as it was used by LegacyOperation class
- Removed `OSModel` usage in CoreModuleDirector 
- Updated Model class to have modelId and get a unique id in the constructor 
- Updated tests accordingly to account for model_id
- Update SubscriptionModel to have both `model_id` (via Model base) and `id` property
- Added new `mergeData` for Model class for adding data to a model (different from initalizeFromJSON since that clears the original data)
- Updated `device_os` to string or number but may revisit to just be string
- Added comments to `ModelChangeTags` but we want to remove the usage altogether 
- Removed `copyOneSignalIdPromiseFromIdentityModel` and `updateModelWithCurrentUserOneSignalId` as it relies on the awaitOneSignalIdAvailable promise but the new IdentityModel does not have logic similar to that
- Removed `resetModelRepoAndCache`, will revisit it later if needed 
- Some tests will remain broken till we update more of the codebase

# Systems Affected
   - [x] WebSDK
   - [ ] Backend
   - [ ] Dashboard

# Validation
### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [ ] No Typescript warnings

Other:

### Checklist

---

## Related Tickets

---

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Website-SDK/1285)


src/core/models/Model.ts
```diff
@@ -82,20 +82,20 @@ export interface ModelChangedArgs<T extends object = object> {
...
+type BaseModel = { modelId: string };
 
 export class Model<
   U extends object = BaseModel,
   T extends U & BaseModel = U & BaseModel,
 > implements IEventNotifier<IModelChangedHandler>
 {
   /**
-   * A unique identifier for this model.
+   * Legacy id used a keypath for the IndexedDB tables. A unique identifier for this model.
```
sherwinski, line 93:
```suggestion
   * Legacy Id used as keypath for the IndexedDB tables. A unique identifier for this model.
```

sherwinski (approved):

PR #1287, opened 2025-04-25
Web SDK Refactor: Update Managers

fadi-george:
# Description
## 1 Line Summary

## Details
- moved simple model store and singleton model store to core/modelStores
- removed `SubscriptionStateKind` usage in favor `NotificationType/NotifcationTypeValue`
- updated `SubscriptionModel` to have `web_auth` and `web_p256` properties
- updated `LoginManager` to handle new models logic
- fixed broken tests and test helpers, skipped few tests that will need to revisited

# Systems Affected
   - [x] WebSDK
   - [ ] Backend
   - [ ] Dashboard

# Validation
### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [ ] No Typescript warnings

Other:

### Checklist

---

## Related Tickets

---

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Website-SDK/1287)

---
  - https://app.asana.com/0/0/1210104365482126


src/shared/managers/SubscriptionManager.ts
```diff
@@ -820,15 +805,16 @@ export class SubscriptionManager {
   private async getSubscriptionStateFromBrowserContext(): Promise<PushSubscriptionState> {
     const { optedOut, subscriptionToken } = await Database.getSubscription();
 
-    const pushSubscriptionOSModel: OSModel<SupportedSubscription> | undefined =
+    const pushSubscriptionModel =
       await OneSignal.coreDirector.getPushSubscriptionModel();
-    const isValidPushSubscription =
-      isCompleteSubscriptionObject(pushSubscriptionOSModel?.data) &&
-      !!pushSubscriptionOSModel?.onesignalId;
```
sherwinski, line 827: Double checking, is the check no longer needed? `!!pushSubscriptionOSModel?.onesignalId`

fadi-george: Android's SubscriptionModel doesnt have onesignal_id, Ive kept track of this in my notes but could revisit this later when going back to fix the old tests/write new tests.

sherwinski (approved):

PR #1294, opened 2025-05-08
Web SDK Refactor: Enqueue Login Operation

fadi-george:
# Description
## 1 Line Summary

## Details
- removes `LoginUserFromSubscriptionOperationExecutor` as that is a special case for [legacy player id](https://github.com/OneSignal/OneSignal-Android-SDK/blob/aaf6693dd6570701133aaf69932fead6686c2481/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/internal/OneSignalImp.kt#L264)
- exposes operationRepo, identityModelStore, propertiesModelStore, and subscriptionModelStore from core module director
- adds new Database.getPushId static method to retrieve last known push id from Options
- remove config model & config model store in favor of `Database.getPushId`
- removes `_path` option in store listener as we dont care about nested fields
- adds new `getBySubscriptionId` as we dont want to get by modelid for getting specific subscription models
- updates `Model` class to not save modelId to data map
- updates OperationQueueItem to be pass in an object rather than multiple params
- removes `initializeUser` in User class as we need push subscription to be ready when attempting to create a user
- removes usage `isTemporary` for initialize user
- updates `createUserOnServer` to follow similar logic to Android SDK with enqueue a login operation and added create subscription op otherwise it would fail

# Systems Affected
   - [x] WebSDK
   - [ ] Backend
   - [ ] Dashboard

# Validation
### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [ ] No Typescript warnings

Other:

### Checklist

---

## Related Tickets

---

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Website-SDK/1294)


src/shared/managers/SubscriptionManager.ts
```diff
@@ -177,7 +177,7 @@ export class SubscriptionManager {
 
     if (!pushModel) {
       OneSignal.coreDirector.generatePushSubscriptionModel(rawPushSubscription);
-      await UserDirector.createAndHydrateUser();
+      UserDirector.createAndHydrateUser();
```
sherwinski, line 180: I'm not super familiar with the entire user-creation flow but is there any possible for a race condition by doing this?

fadi-george: I will update it to be awaited in a new pr (login changes) since this would happen in the init call and login would need to await for operations to finish.

sherwinski (approved):

PR #1297, opened 2025-05-14
Web SDK Refactor: Fix AddEmail and AddSms

fadi-george:
# Description
## 1 Line Summary

## Details
- Update generatePushSubscriptionModel method to not propagate as we want have init call enqueue a login operation and a create subscription operation in that order
- Adds refresh operation for execution response for `createSubscription` action in SubscriptionOperationExecutor. This is since addSms and addEmail dont get an empty response from the backend and so well just get the full subscription details through a get user call. 
- Removes bootstrap method in ModelStoreListener and subscribe right away in the constructor. 
- Update clear call for `ModelStore` to remove the IndexedDB entry, otherwise youd have duplicate items
- Removes modelid check in ModelStore load method since modelId are randomly generated but in AndroidSDK you could have singleton ids.
- Updates `SimpleModelStore` create method to set the modelId if present in the object param.
- Updates `SubscriptionModel` to have defined sdk, device_model, and device_os values
- Adds tests for `addEmail`, `addSms`, `removeEmail`, and `removeSMS` user actions
- Updates User `addSms` and `addEmail` methods to check if subscription item already exists to avoid creating dupkicates

# Systems Affected
   - [x] WebSDK
   - [ ] Backend
   - [ ] Dashboard

# Validation
### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [ ] No Typescript warnings

Other:

### Checklist

---

## Related Tickets

---

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Website-SDK/1297)


src/core/modelRepo/ModelStore.ts
```diff
@@ -96,17 +96,23 @@ export abstract class ModelStore<TModel extends Model>
     );
   }
 
-  replaceAll(
+  async replaceAll(
     newModels: TModel[],
     tag: ModelChangeTagValue = ModelChangeTags.NORMAL,
-  ): void {
+  ): Promise<void> {
     this.clear(tag);
```
sherwinski, line 102: should be awaited?

src/core/modelStores/SubscriptionModelStore.ts
```diff
@@ -18,10 +18,10 @@ export class SubscriptionModelStore extends SimpleModelStore<SubscriptionModel>
...
   }
 
-  override replaceAll(
+  override async replaceAll(
     models: SubscriptionModel[],
     tag?: ModelChangeTagValue,
-  ): void {
+  ): Promise<void> {
     if (tag !== ModelChangeTags.HYDRATE) {
       return super.replaceAll(models, tag);
```
sherwinski, line 26: should be awaited?

src/core/modelRepo/ModelStore.ts
fadi-george: remove promise logic, updated clear to call Database.remove with modelId

src/core/modelStores/SubscriptionModelStore.ts
fadi-george: updated to not need promise

sherwinski (approved):

PR #1303, opened 2025-05-29
fix(api): update User `getTags` and `getLanguage` to return a Promise

sherwinski:
# Description
## 1 Line Summary

Updates the User `getTags` and `getLanguage` functions in the API spec to return a Promise.

## Details

In order to address https://github.com/OneSignal/react-onesignal/pull/173 we need to accomplish two things:
- update the spec in this repo so that https://github.com/OneSignal/web-shim-codegen can generate the correct types and function signature
  - this PR addresses this point
- update the templates in web-shim ([example](https://github.com/OneSignal/web-shim-codegen/blob/main/src/support/react/oneSignalFunctionTemplates.ts#L35-L54)) so that the function body is generated using a Promise-style
  - change needs to be made in web-shim directly

The end result should be as follows:
**Before:**
```ts
function userGetTags(): { [key: string]: string } {
  let retVal: { [key: string]: string };
  window.OneSignalDeferred?.push((OneSignal: IOneSignalOneSignal) => {
    retVal = OneSignal.User.getTags();
  });
  return retVal;
}
```

**After:**
```ts
function userGetTags(): Promise<{ [key: string]: string }> {
  return new Promise((resolve) => {
    window.OneSignalDeferred?.push((OneSignal: IOneSignalOneSignal) => {
      resolve(OneSignal.User.getTags());
    });
  });
}
```
# Systems Affected
   - [ ] WebSDK
   - [ ] Backend
   - [ ] Dashboard

# Validation
### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [ ] No Typescript warnings

Other:

### Checklist

---

## Related Tickets

---

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Website-SDK/1303)

### DIFF
--- a/api.json
+++ b/api.json
@@ -455,7 +455,7 @@
         "name": "getTags",
         "isAsync": false,
         "args": [],
-        "returnType": "{ [key: string]: string }"
+        "returnType": "Promise<{ [key: string]: string }>"
       },
       {
         "name": "addEventListener",
@@ -507,7 +507,7 @@
         "name": "getLanguage",
         "isAsync": false,
         "args": [],
-        "returnType": "string"
+        "returnType": "Promise<string>"
       }
     ],
     "properties": [


fadi-george:
So the api spec is just for the web shim packages. We dont need to change the web code to be async for getTags and getLanguage?

rgomezp:
Yeah I'm confused. It sounds like the issue is that folks want getTags and getLanguage to force-fetch the tags?

If they're getting out of sync that sounds like a different problem where they're not being hydrated properly on session start.

Also, you are updating the return type but you left it `"isAsync": false`.

Needs more investigation

sherwinski:
I get the feeling that the [wrapper code](https://github.com/OneSignal/react-onesignal/blob/main/index.ts#L872-L878) is trying to set the returned value synchronously but the operation itself is async, which could be related to [this change](https://github.com/OneSignal/OneSignal-Website-SDK/commit/70f1a3909edbc406bb5bf9dbefeca272c485d4d2#diff-b0293c86322369fe531c409a2a8c334e57c099b481e3dc78258875a4e6923fb3) from a while back. I tested this by comparing the returned value from the wrapper (`OneSignal.User.getTags()`) against the web SDK directly (`window.OneSignal.User.getTags()`).

> Also, you are updating the return type but you left it "isAsync": false

I originally changed it to be marked async but after realizing it would be a breaking change, rolled it back and started a conversation internally to figure out what's the best way forward.

PR #1306, opened 2025-06-03
Web SDK Refactor: Fix Add/Remove Tags

fadi-george:
# Description
## 1 Line Summary
Continuation of the [Web SDK Refactor project](https://docs.google.com/document/d/1JsbmxVM_n6K9nRXwbJf6mQr5h88qy_vIEiBGY0Ub_lE/edit?tab=t.0#heading=h.ojv84gskof17). Fixes bug with add/remove tag.

## Details
Before, the addTag would bug out and instead do a delete tag operation due to this `if (newValue != null && typeof newValue === 'string')` check in properties model store listener. But decided to simplify the logic by removing `SetTagOperation` and `DeleteTagOperation` in favor of the regular `SetPropertyOperation`. This is since we have tags being stored as an object vs a `MapModel` in Android for the PropertiesModel.

- removed Set/Delete Tag ops and references
- updated User methods (remove tag/tags) to set tag value to empty string since BE expects this value if youre attempting to delete tags
- updated UpdateUserOperationExecutor to remove empty string tags when updating properties model after successful network call
- removes deltas 

# Systems Affected
   - [x] WebSDK
   - [x] Backend
   - [ ] Dashboard

# Validation
### Info
The test suite has been updated to reflect the changes in property handling.

### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [x] No Typescript warnings

Other:

### Info
N/A

### Checklist

---

## Related Tickets

---

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Website-SDK/1306)

### DIFF
--- a/src/core/executors/UpdateUserOperationExecutor.ts
+++ b/src/core/executors/UpdateUserOperationExecutor.ts
@@ -5,10 +5,7 @@ import {
 import { PropertyOperationHelper } from 'src/shared/helpers/PropertyOperationHelper';
 import Log from 'src/shared/libraries/Log';
 import { OPERATION_NAME } from '../constants';
-import {
-  IPropertiesModelKeys,
-  IPropertiesModelValues,
-} from '../models/PropertiesModel';
+import { IPropertiesModelKeys } from '../models/PropertiesModel';
 import { type IdentityModelStore } from '../modelStores/IdentityModelStore';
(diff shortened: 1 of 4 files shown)


sherwinski (approved):

sherwinski:
It wasn't mentioned in the PR description, but why are the deltas removed?

PR #1307, opened 2025-06-04
Web SDK Refactor: Adjust Login Logic

fadi-george:
# Description
## 1 Line Summary
- Removes unnecessary delay after creating a subscription, this was due to not using local ids via IdManager

## Details
Currently, if we have logic like:
```
 OneSignalDeferred.push(async function (OneSignal) {
        await OneSignal.init({
          appId,
        });
      });
      //uncomment to test login
      OneSignalDeferred.push(async function (OneSignal) {
        await OneSignal.login('fadi-11111');
      });
```
It will error like so, if we dont have the `await delay(OP_REPO_EXECUTION_INTERVAL * 2);` in the UserDirector as init will enqueue login op but so will OneSignal.login. This will group the ops together since (previously) onesignal would be undefined.
Screenshot 2025-06-04 at 12 07 51 PM

And so updated logic to mimic Android and use id manager to generate local ids.
- Add resetUserModels helper to replace Identity and Properties Model for operations like login and logout
- Uses createUserOnServer for logout since were now using local ids when resetting the models. initializeUser has a check for onesignalid property and will not create a user if that field exists. 

This leads to the grouped ops being different and the "last login" to be in a different group due to the different create comparison key (since the id is different).
Screenshot 2025-06-04 at 1 58 49 PM

# Systems Affected
   - [x] WebSDK

# Validation
### Info
Existing tests cover the user login flow.

### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [x] No Typescript warnings

Other:

### Info
N/A - no visual changes.

### Checklist

---

## Related Tickets

---

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Website-SDK/1307)

### DIFF
--- a/src/onesignal/OneSignal.test.ts
+++ b/src/onesignal/OneSignal.test.ts
@@ -31,6 +31,7 @@ import { SubscriptionModel } from 'src/core/models/SubscriptionModel';
 import { ICreateUserSubscription } from 'src/core/types/api';
 import { ModelChangeTags } from 'src/core/types/models';
 import Log from 'src/shared/libraries/Log';
+import { IDManager } from 'src/shared/managers/IDManager';
 import Database, {
   IdentityItem,
   PropertiesItem,
@@ -38,6 +39,7 @@ import Database, {
 } from 'src/shared/services/Database';
 import LocalStorage from 'src/shared/utils/LocalStorage';
 
+vi.spyOn(Log, 'error').mockImplementation(() => '');
 const debugSpy = vi.spyOn(Log, 'debug');
 
 const getIdentityItem = async () =>
@@ -54,6 +56,7 @@ describe('OneSignal', () => {
 
     await Database.put('identity', {
       modelId: '123',
+      modelName: 'identity',
       onesignal_id: DUMMY_ONESIGNAL_ID,
     });
     await window.OneSignal.init({ appId: APP_ID });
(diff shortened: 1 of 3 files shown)


src/onesignal/OneSignal.test.ts
```diff
@@ -558,24 +566,31 @@ describe('OneSignal', () => {
 
         // identity model should be reset
         identityModel = window.OneSignal.coreDirector.getIdentityModel();
-        expect(identityModel.toJSON()).toEqual({}); // no one signal id
+        const onesignalId = identityModel.onesignalId;
+        expect(identityModel.toJSON()).toEqual({
+          onesignal_id: expect.any(String),
+        });
+        expect(IDManager.isLocalId(onesignalId)).toBe(true);
```
sherwinski, line 573: I'm not following the logic of these test changes. So originally the tests shows that a logged-out user has an undefined onesignalId but now it's supposed to exist? Where is the SDK deriving it from?

fadi-george: Updated pr description.

sherwinski (approved):

src/onesignal/OneSignal.test.ts
fadi-george: Reposting here:
The op repo groups ops together by either their create comparison or modify comparison key. So they can be executed together like you can see in LoginOpExecutor.
The previous logic didnt set the local ids so youd have multiple ops group together like
loginop (from init but its create the user), create sub, ..., and then the login with (externalid) op
The login op executor sees theres two login ops being executed together and throws the error.
We got around this by having the await delay (oprepointerval*2) in the init/user director.

PR #1309, opened 2025-06-11
Web SDK Refactor: Operations IndexedDB Table

fadi-george:
# Description
## 1 Line Summary
Continuation of the [Web SDK Refactor project](https://docs.google.com/document/d/1JsbmxVM_n6K9nRXwbJf6mQr5h88qy_vIEiBGY0Ub_lE/edit?tab=t.0#heading=h.ojv84gskof17). Updates logic to persist stored operations to the model store.

## Details
- Fixes tests for operation repo class & adds test cases for initialize data from IndexedDB
- Moves modelid changes for initalizeFromJSON (from simplemodlestore) to be inside the model class itself since operation uses a regular model store
- Remove unused rebuild method for Database class
- Updates IndexedDB to have a new database version where we create a new operations table. And so I updated the Database Version 7. Should we ever need to revert, we would have update revert the changes but also increment the version to 8 otherwise there would be a unrecoverable version error.

# Systems Affected
   - [x] WebSDK
   - [ ] Backend
   - [ ] Dashboard

# Validation
### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [ ] No Typescript warnings

Other:

### Checklist

---

## Related Tickets

---

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Website-SDK/1309)

---
  - https://app.asana.com/0/0/1210527513506671

### DIFF
--- a/src/shared/services/IndexedDb.ts
+++ b/src/shared/services/IndexedDb.ts
@@ -87,19 +87,6 @@ export default class IndexedDb {
       );
     } else {
       Log.warn('OneSignal: Fatal error opening IndexedDb database:', error);
-
-      // version error will occur if the existing db version is higher than the version we are trying to open
-      // so we will need to nuke the db in this case
-      if (error.name.includes('VersionError')) {
-        Log.warn(
-          'OneSignal: IndexedDb version mismatch, deleting database',
-          error,
-        );
-        const req = indexedDB.deleteDatabase(this.databaseName);
-        req.onsuccess = () => {
-          this.open(this.databaseName);
-        };
-      }
     }
   }
 
@@ -197,9 +184,6 @@ export default class IndexedDb {
     }
     if (newDbVersion >= 7 && event.oldVersion < 7) {
       db.createObjectStore(ModelName.Operations, { keyPath: 'modelId' });
-    }
-    if (newDbVersion >= 8 && event.oldVersion < 8) {
-      // Make sure to update the database version at the top of the file
     }
     // Wrap in conditional for tests
     if (typeof OneSignal !== 'undefined') {
--- a/src/shared/services/IndexedDb.ts
+++ b/src/shared/services/IndexedDb.ts
@@ -3,7 +3,7 @@ import Utils from '../context/Utils';
 import Emitter from '../libraries/Emitter';
 import Log from '../libraries/Log';
 
-const DATABASE_VERSION = 6;
+const DATABASE_VERSION = 7;
 
 export const LegacyModelName = {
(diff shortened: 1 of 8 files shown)


src/shared/services/IndexedDb.ts
```diff
@@ -87,6 +87,19 @@ export default class IndexedDb {
...
       Log.warn('OneSignal: Fatal error opening IndexedDb database:', error);
+
+      // version error will occur if the existing db version is higher than the version we are trying to open
+      // so we will need to nuke the db in this case
+      if (error.name.includes('VersionError')) {
+        Log.warn(
+          'OneSignal: IndexedDb version mismatch, deleting database',
+          error,
+        );
+        const req = indexedDB.deleteDatabase(this.databaseName);
```
fadi-george: @jkasten2 This change would nuke the DB unless we release a new version.

jkasten2: What happened before adding this code?
* Did the indexDb just fail to open and the SDK doesn't work?

Also want to callout this code won't help if we need to rollback this version (before this User Model refactor), as the `deleteDatabase` didn't exist in version we want the option to rollback too.

Lastly, I don't think we should delete the whole database if it's possible to avoid it. Doing so will cause existing subscriptions to become detached from existing Users and new User records being created, as the SDK will make network calls to recreate them.

fadi-george: Thats a fair point. We could always just update the version instead. Removed the changes @jkasten2 .

sherwinski: If I'm following the changes correctly, it looks like you removed all of the code to persist operations in [b37bf65](https://github.com/OneSignal/OneSignal-Website-SDK/pull/1309/commits/b37bf65e3e734aaecbcba12976a4d9889728cfe9). Is that correct? We should update the PR description because it doesn't seem accurate anymore.

sherwinski: Ah maybe the key part is here https://github.com/OneSignal/OneSignal-Website-SDK/pull/1309/commits/c3dc8d9d74e05ceb87cfaa3d62a42a62ec41d35d#diff-dd2c30ab15a2323012089532df9ab057c1861af04fa6022f1af434781885d39bR199

sherwinski: Either way I think the PR description needs to be updated.

fadi-george: Updated pr description.

sherwinski (approved):

PR #1310, opened 2025-06-16
Web SDK Refactor: Fix Async

fadi-george:
# Description
## 1 Line Summary
Continuation of the [Web SDK Refactor project](https://docs.google.com/document/d/1JsbmxVM_n6K9nRXwbJf6mQr5h88qy_vIEiBGY0Ub_lE/edit?tab=t.0#heading=h.ojv84gskof17). Preserve order when actions like `login` and `addTag` when called without await.

## Details
When making calls like so:
```
OneSignal.login(...)
OneSignal.User.addTag(... , ...)
```
The addTag gets executed first before the login. This is due to login action having too many async operations.
- moves consent check to login user operation executor 
- makes MainHelper sync by having get app id from OneSignal.config

# Systems Affected
   - [x] WebSDK
   - [ ] Backend
   - [ ] Dashboard

# Validation
### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [ ] No Typescript warnings

Other:

### Checklist

---

## Related Tickets

---

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Website-SDK/1310)

---
  - https://app.asana.com/0/0/1210562926736778

### DIFF
--- a/src/shared/helpers/MainHelper.ts
+++ b/src/shared/helpers/MainHelper.ts
@@ -121,7 +121,7 @@ export default class MainHelper {
   }
 
   static async getNotificationIcons() {
-    const appId = await MainHelper.getAppId();
+    const appId = MainHelper.getAppId();
     if (!appId) {
       throw new InvalidStateError(InvalidStateReason.MissingAppId);
     }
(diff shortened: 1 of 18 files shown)


src/shared/helpers/MainHelper.ts
```diff
@@ -223,13 +223,8 @@ export default class MainHelper {
...
   }
 
-  static async getAppId(): Promise<string> {
-    if (OneSignal.config.appId) {
-      return Promise.resolve(OneSignal.config.appId);
-    } else {
-      const appId = await Database.get<string>('Ids', 'appId');
-      return appId;
-    }
+  static getAppId(): string {
```
fadi-george, line 226: @jkasten2 can you review

sherwinski (approved):
seems good to me

PR #1316, opened 2025-06-30
Web SDK Refactor

fadi-george:
# Description
## 1 Line Summary
Addresses initial goal of [spec](https://docs.google.com/document/d/1JsbmxVM_n6K9nRXwbJf6mQr5h88qy_vIEiBGY0Ub_lE/edit?tab=t.0#heading=h.ojv84gskof17) to group operations and halt operations when errors are encounted. 

## Details
- Implements a queueing system similar to Android SDK with operation repo, executors, model stores, and model store listeners
- Updates and adds new tests related to operation repo and executor logic 
- Allows operations to be enqueued in the right order e.g.:
```
OneSignal.login("fadi-1");
OneSignal.User.addTags({'A': '1', 'B': ''})
OneSignal.login("fadi-2");
OneSignal.User.addTag('C', '3')
```
Before, it will do an order like creating user 1 then user 2 (error) then batch the tag updates. 

# Systems Affected
   - [x] WebSDK
   - [ ] Backend
   - [ ] Dashboard

# Validation
### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [ ] No Typescript warnings

Other:

### Checklist

---

## Related Tickets

---

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Website-SDK/1316)


src/entries/pageSdkInit2.test.ts
sherwinski, line 1: Can we rename this file to something a bit clearer?

fadi-george: The test file is next pageSdkInit

PR #1318, opened 2025-06-30
Updates OneSignal API URL

fadi-george:
# Description
## Details
Addresses ticket: https://app.asana.com/0/home/1209206699711965/1207914352046248
Update api url to be `api.onesignal.com` 

# Systems Affected
   - [x] WebSDK

# Validation
### Info
Existing tests cover the functionality, and they have been updated to reflect the correct API URL.

### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [x] No Typescript warnings

Other:

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Website-SDK/1318)

### DIFF
--- a/__test__/README.md
+++ b/__test__/README.md
@@ -62,7 +62,7 @@ import { server } from '__test__/support/mocks/server';
 import { http, HttpResponse } from 'msw';
 
 server.use(
-  http.get('https://api.onesignal.com/v1/notifications', () =>
+  http.get('https://api.onesignal.com/notifications', () =>
     HttpResponse.json({ result: {}, status: 200 }),
   ),
 );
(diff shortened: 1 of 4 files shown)


sherwinski (commented):
Nit: looks like there are two other references that should be changed:

- https://github.com/OneSignal/OneSignal-Website-SDK/blob/ba2927fff17ae1f0dfd8091bd5b0fcdc108df07c/preview/README.md?plain=1#L69
- https://github.com/OneSignal/OneSignal-Website-SDK/blob/352f16c07f07b14064a17d05ff8dd00738aabe6d/src/shared/managers/SdkEnvironment.test.ts#L37

sherwinski (approved):

PR #1320, opened 2025-07-01
Standardizes OneSignal API endpoint URLs

fadi-george:
# Description
## Details
This pull request standardizes how the application interacts with the OneSignal API across different environments. It achieves this by:

-   Ensuring consistency in API endpoint URLs, by explicitly using `api.onesignal.com` as the base URL for production environments.
-   Corrects the staging API URL to include the `onesignal-staging.com` domain, aligning with the intended testing environment.
-   Adjusts the sync call endpoint to remove the extra forward slash, resolving potential routing issues.

These changes improve the reliability and predictability of API calls, which reduces errors and facilitates easier debugging and maintenance.

# Systems Affected
   - [x] WebSDK
   - [x] Backend

# Validation
### Info
Unit tests have been updated to reflect the changes in API URLs and ensure correct behavior across environments.

### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [x] No Typescript warnings

Other:

### Info
N/A

### Checklist

---

## Related Tickets

---


src/shared/managers/SdkEnvironment.ts
```diff
@@ -92,13 +92,13 @@ export default class SdkEnvironment {
...
+          return new URL(`http://${apiOrigin}:${TURBINE_API_URL_PORT}/api/v1/`);
         }
-        return new URL(`http://${apiOrigin}:${API_URL_PORT}/api/v1`);
+        return new URL(`http://${apiOrigin}:${API_URL_PORT}/api/v1/`);
       case EnvironmentKind.Staging:
-        return new URL(`https://${apiOrigin}/api/v1`);
+        return new URL(`https://${apiOrigin}/api/v1/`);
       case EnvironmentKind.Production:
-        return new URL('https://onesignal.com/api/v1');
+        return new URL('https://api.onesignal.com/');
```
jkasten2: This is used in a few places where this would be a breaking change:
1. subscribeSafariPromptPermission() - This URL has to match exactly with the .zip files generated for macOS Safari legacy push support. I'll include the full internal details in the Asana ticket.
2. getNotificationIcons() - This fetches a icon, so api.onesignal.com may not work. If it does work, I still don't think it should be used that way, unless someone from the backend can confirm this won't break in the future.

jkasten2 (changes requested):
There are two problems with

src/shared/managers/SdkEnvironment.ts
fadi-george: The icon route seems to be the same for new and legacy.
The new api route does not seem to work for safari request permission so I added a legacy api option.

PR #1321, opened 2025-07-01
Enables logging by default in production builds

fadi-george:
# Description
## Details
The changes ensure that logging is enabled by default in production builds. 
This is achieved by providing a default value for the `__LOGGING__` Vite define variable, ensuring that it is set to `true` when the `LOGGING` environment variable is not explicitly defined in production.

# Systems Affected
   - [ ] WebSDK
   - [ ] Backend
   - [ ] Dashboard

# Validation
### Info
Manual verification to ensure logging works correctly in production builds.

### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [ ] No Typescript warnings

Other:

### Info
N/A

### Checklist

---

## Related Tickets

---


sherwinski (approved):

jkasten2 (approved):
Code changes look good.

PR Title, you mean disables logging by default in production builds 😁

PR #1327, opened 2025-07-09
Login Bug Fix

fadi-george:
# Description
## 1 Line Summary
Fixes login call for situations where login is called before a user accepts notification permissions.

## Details
With web refactor release, there exists a bug where the developer might call login for a user without granted notification permissions. This will cause an identity entry to be made a local-id and a subscription model will also be created  with a local-id (after accepting permissions).

Login will be called before a push subscription is generated and so the `_login` in LoginManager will set `isCreatingUser` to true. But push subscription is not found so it will return early:
```
    const pushOp = await OneSignal.coreDirector.getPushSubscriptionModel();
    if (!pushOp) return Log.info('No push subscription found');
```

`_updatePushSubscriptionModelWithRawSubscription` will be called on init which attempts to create a user. But `user.isCreatingUser` is true and so it will do an early return when it shouldn't be the case.

So did the following changes:
- removes `isCreatingUser` since calling login with same id is already handled and OneSignal.init should wait for OneSignal.init (if granted permission). Otherwise login operations are enqueued.
- if the case for users without data, we wouldnt need any other changes
- but for users with buggy data (local id for both identity and push sub in IndexeDB), we need to create the user i.e. add this check:
```
 if (!pushModel) {
    OneSignal.coreDirector.generatePushSubscriptionModel(rawPushSubscription);
    return UserDirector.createUserOnServer();
  } else if (!pushModel.id) {
    return UserDirector.createUserOnServer();
  }
```

# Systems Affected
   - [x] WebSDK
   - [ ] Backend
   - [ ] Dashboard

# Validation
### Info
**You can clear your data then open branch `fg/login-bug-ref` and run the dev server `npm run dev`. Then see that the data created is buggy and no users api call is made.
You can then check out this branch to see the call gets made.**

### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [ ] No Typescript warnings

Other:

### Checklist

---

## Related Tickets

---

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Website-SDK/1327)

### DIFF
--- a/src/core/executors/LoginUserOperationExecutor.ts
+++ b/src/core/executors/LoginUserOperationExecutor.ts
@@ -1,6 +1,5 @@
 import { ModelChangeTags } from 'src/core/types/models';
 import { ExecutionResult, IOperationExecutor } from 'src/core/types/operation';
-import User from 'src/onesignal/User';
 import OneSignalError from 'src/shared/errors/OneSignalError';
 import Environment from 'src/shared/helpers/Environment';
 import EventHelper from 'src/shared/helpers/EventHelper';
@@ -60,9 +59,7 @@ export class LoginUserOperationExecutor implements IOperationExecutor {
     const startingOp = operations[0];
 
     if (startingOp instanceof LoginUserOperation)
-      return this.loginUser(startingOp, operations.slice(1)).finally(() => {
-        User.createOrGetInstance().isCreatingUser = false;
-      });
+      return this.loginUser(startingOp, operations.slice(1));
 
     throw new Error(`Unrecognized operation: ${startingOp.name}`);
   }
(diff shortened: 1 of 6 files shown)


sherwinski (commented):
- Let's add a test for this
- What was the original intention for `isCreatingUser`? I'm just trying to think through the potential side effects of removing it completely

fadi-george:
> * isCreatingUser
Was used for addSms and addEmail but seemed redundant to use since you could check for hasOneSignalId.
It was used in createUserOnServer to prevent making call if another was in progress.

sherwinski (approved):

PR #1329, opened 2025-07-09
Have Safari use Standard Web Notifications

fadi-george:
# Description
## 1 Line Summary
Removes legacy safari notifications logic in favor of new standard web notifications.

## Details
`window.safari.pushNotification` was released in Safari 7 to allow push notifications for safari browsers. We added a lot of logic (now considered legacy) to have custom logic for chromium+firefox browsers and safari browsers. But since then Safari has pushed updates (v16+) to conform to standard [window.Notifications api](https://developer.mozilla.org/en-US/docs/Web/API/Notification).
- Removes all legacy safari notification logic
- Adds porting logic in `_updatePushSubscriptionModelWithRawSubscription` to change from the old token format to the standard format
- Add logic to save token and id in IndexedDB 

Old Data:
Screenshot 2025-07-09 at 4 12 43 PM

New/Migrated Data:
Screenshot 2025-07-09 at 4 42 23 PM

# Systems Affected
   - [x] WebSDK
   - [ ] Backend
   - [ ] Dashboard

# Validation
### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [ ] No Typescript warnings

Other:

### Checklist

---

## Related Tickets

---

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Website-SDK/1329)

### DIFF
--- a/src/shared/managers/PermissionManager.ts
+++ b/src/shared/managers/PermissionManager.ts
@@ -1,10 +1,5 @@
-import {
-  InvalidArgumentError,
-  InvalidArgumentReason,
-} from '../errors/InvalidArgumentError';
-import { NotificationPermission } from '../models/NotificationPermission';
 import OneSignalError from '../errors/OneSignalError';
-import Environment from '../helpers/Environment';
+import { NotificationPermission } from '../models/NotificationPermission';
 
 /**
  * A permission manager to consolidate the different quirks of obtaining and evaluating permissions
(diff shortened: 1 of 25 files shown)


vitest.config.ts
```diff
@@ -4,7 +4,7 @@ import { defineConfig } from 'vitest/config';
 export default defineConfig({
   define: {
     __API_ORIGIN__: JSON.stringify('onesignal.com'),
-    __API_TYPE__: JSON.stringify('staging'),
+    __API_TYPE__: JSON.stringify('production'),
```
sherwinski, line 7: what's the purpose of this change?

sherwinski (commented):
Overall the code looks good to me but I think we should let @jkasten2 weigh in on this one in case there are extra considerations I might be missing.

src/shared/managers/PermissionManager.ts
```diff
@@ -26,47 +21,15 @@ export default class PermissionManager {
...
-   * Returns the Safari browser's notification permission as reported by the browser.
-   *
-   * @param safariWebId The Safari web ID necessary for Legacy Safari on macOS.
-   */
-  private static getLegacySafariNotificationPermission(
-    safariWebId?: string,
-  ): NotificationPermission {
-    if (safariWebId)
-      return window.safari.pushNotification.permission(safariWebId)
-        .permission as NotificationPermission;
```
jkasten2: We need to check if `window.safari.pushNotification.permission(safariWebId)` returns false, and only then use the new VAPID push. This is due to a bug that still exists today in Safari that doesn't allow migrating someone who has accepted Safari legacy push for the domain to standard notifications.

See this Asana ticket for more details:
https://app.asana.com/1/780103692902078/project/1204295426001934/task/1204303093244836?focus=true

jkasten2 (changes requested):

vitest.config.ts
fadi-george: This is just for the tests and will affect the api url is. It doesnt matter too much in this case since we use msw to mock the requests.

fadi-george:
I think one option to push a breaking change v17 and remove the remaining legacy safari specific code.

sherwinski:
Closing in favor of #1441

PR #1330, opened 2025-07-12
fix: allow email/SMS subscription creation without requiring push subscription

sherwinski:
# Description
## 1 Line Summary

Allow Email/SMS subscription creation without a web push subscription

## Details

Currently, users cannot create email or SMS subscriptions after login unless they also have a push subscription. This is because the user creation logic in UserDirector.createUserOnServer() and LoginManager._login() only proceeds when a push subscription exists.

The root cause is that CreateSubscriptionOperation cannot execute when the onesignalId is still a local ID (temporary). User creation is needed to replace the local ID with a real backend ID, but user creation was being prevented when no push subscription existed.

# Systems Affected
   - [ ] WebSDK
   - [ ] Backend
   - [ ] Dashboard

# Validation
### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [ ] No Typescript warnings

Other:

### Checklist

---

## Related Tickets

---

### DIFF
--- a/src/page/managers/LoginManager.ts
+++ b/src/page/managers/LoginManager.ts
@@ -48,14 +48,15 @@ export default class LoginManager {
     const appId = MainHelper.getAppId();
 
     const pushOp = await OneSignal.coreDirector.getPushSubscriptionModel();
-    if (!pushOp) return Log.info('No push subscription found');
-    OneSignal.coreDirector.operationRepo.enqueue(
-      new TransferSubscriptionOperation(
-        appId,
-        newIdentityOneSignalId,
-        pushOp.id,
-      ),
-    );
+    if (pushOp) {
+      OneSignal.coreDirector.operationRepo.enqueue(
+        new TransferSubscriptionOperation(
+          appId,
+          newIdentityOneSignalId,
+          pushOp.id,
+        ),
+      );
+    }
 
     await OneSignal.coreDirector.operationRepo.enqueueAndWait(
       new LoginUserOperation(
--- a/src/onesignal/UserDirector.ts
+++ b/src/onesignal/UserDirector.ts
@@ -11,27 +11,43 @@ export default class UserDirector {
     const identityModel = OneSignal.coreDirector.getIdentityModel();
     const appId = MainHelper.getAppId();
 
+    const allSubscriptions = await OneSignal.coreDirector.getAllSubscriptionsModels();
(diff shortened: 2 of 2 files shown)


src/page/managers/LoginManager.ts
```diff
@@ -48,14 +48,15 @@ export default class LoginManager {
     const appId = MainHelper.getAppId();
 
     const pushOp = await OneSignal.coreDirector.getPushSubscriptionModel();
```
fadi-george, line 50: Did you manually test the case of resetting all data then having index.html have both init and login calls. 
If you had multiple users w/ external_id (user-1, user-2) and were using login user-2 then we might get two subscriptions with user-2.

fadi-george: Should compare against v1603 versions

fadi-george:
Also for people that didnt accept permissions but did get added a subscription (& stored in IndexedDB) we should create them on the server.

sherwinski:
> Also for people that didnt accept permissions but did get added a subscription (& stored in IndexedDB) we should create them on the server.

Can you elaborate on this? I'm not sure what you specifically mean by "subscription" in this case: email, sms, or something else?

src/page/managers/LoginManager.ts
sherwinski: can you add specific steps for what you'd like to test here? It sounds like you want to test adding sms/email for an already-existing user after clearing site data.

fadi-george:
> > Also for people that didnt accept permissions but did get added a subscription (& stored in IndexedDB) we should create them on the server.
> 
> Can you elaborate on this? I'm not sure what you specifically mean by "subscription" in this case: email, sms, or something else?

Just email and sms.

src/page/managers/LoginManager.ts
fadi-george: Im saying for the normal case of not adding email or not adding sms and just using login w/o accepting permissions.

fadi-george:
> Also for people that didnt accept permissions but did get added a subscription (& stored in IndexedDB) we should create them on the server.

This is still unhandled I think.

sherwinski:
> > Also for people that didnt accept permissions but did get added a subscription (& stored in IndexedDB) we should create them on the server.
> 
> This is still unhandled I think.

Testing this one now

fadi-george (approved):

PR #1333, opened 2025-07-22
Add Fix for Login before Accepting Permissions

fadi-george:
# Description
## 1 Line Summary
Addresses ticket: https://app.asana.com/1/780103692902078/project/1208777190614537/task/1210836212667492
Adds check for any existing login calls then create users with a subscription.

Before it would error since `OneSignal.login` would enqueue a login op. And `createUserOnServer` would also create a login op. So this would throw an in op repo since login operation executor expects just one login op.

## Details

# Systems Affected
   - [x] WebSDK
   - [ ] Backend
   - [ ] Dashboard

# Validation
### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [ ] No Typescript warnings

Other:

### Checklist

---

## Related Tickets

---

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Website-SDK/1333)

### DIFF
--- a/src/shared/managers/SubscriptionManager.ts
+++ b/src/shared/managers/SubscriptionManager.ts
@@ -37,6 +37,7 @@ import { bowserCastle } from '../utils/bowserCastle';
 import { base64ToUint8Array } from '../utils/Encoding';
 import { PermissionUtils } from '../utils/PermissionUtils';
 import { executeCallback, logMethodCall } from '../utils/utils';
+import { IDManager } from './IDManager';
 import SdkEnvironment from './SdkEnvironment';
 export const DEFAULT_DEVICE_ID = '99999999-9999-9999-9999-999999999999';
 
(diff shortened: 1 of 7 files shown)


src/shared/managers/SubscriptionManager.ts
```diff
@@ -170,30 +172,37 @@ export class SubscriptionManager {
...
+        OneSignal.coreDirector.generatePushSubscriptionModel(
+          rawPushSubscription,
+        );
       return UserDirector.createUserOnServer();
-
-      // Bug w/ v160400 release where isCreatingUser was improperly set and never reset
-      // so a check if pushModel id is needed to recreate the user
     }
-    if (!pushModel.id) {
+    // for users with data failed to create a user or user+subscriptipn on the server
```
sherwinski:
```suggestion
    // for users with data failed to create a user or user+subscription on the server
```

src/shared/managers/SubscriptionManager.ts
```diff
@@ -170,30 +172,37 @@ export class SubscriptionManager {
...
+    if (IDManager.isLocalId(pushModel.id)) {
       return UserDirector.createUserOnServer();
     }
 
-    // resubscribing. update existing push subscription model
+    // in case of notifcation state changes, we need to update its web_auth, web_p256, and other keys
```
sherwinski:
```suggestion
    // in case of notification state changes, we need to update its web_auth, web_p256, and other keys
```

PR #1336, opened 2025-07-23
Add Custom Events

fadi-george:
# Description
## Details

# Systems Affected
   - [ ] WebSDK
   - [ ] Backend
   - [ ] Dashboard

# Validation
### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [ ] No Typescript warnings

Other:

### Checklist

---

## Related Tickets

---


src/core/executors/RefreshUserOperationExecutor.ts
```diff
@@ -106,9 +106,9 @@ export class RefreshUserOperationExecutor implements IOperationExecutor {
         model.type = sub.type;
         model.enabled =
           model.notification_types !== NotificationType.UserOptedOut;
-        model.sdk = sub.sdk ?? '';
-        model.device_os = sub.device_os ?? '';
-        model.device_model = sub.device_model ?? '';
+        model.sdk = sub.sdk;
+        model.device_os = sub.device_os;
+        model.device_model = sub.device_model;
```
sherwinski, line 111: does this mean if any of these values are missing, they will get saved/sent as `undefined`?

src/core/operations/TrackEventOperation.ts
```diff
@@ -0,0 +1,88 @@
...
+    this.setProperty('externalId', value);
+  }
+
+  /**
+   * The timestamp when the event occurred.
+   */
+  get timestamp(): string {
+    return this.getProperty('timestamp');
+  }
+  private set timestamp(value: string) {
```
sherwinski: should this default to `now()`?

src/core/requestService/RequestService.ts
```diff
@@ -241,40 +241,6 @@ export class RequestService {
...
-    identity: IdentityModel,
-  ): Promise<OneSignalApiBaseResponse> {
-    const { appId } = requestMetadata;
-    return OneSignalApiBase.patch<{ identity: IUserIdentity }>(
-      `apps/${appId}/users/by/subscriptions/${subscriptionId}/identity`,
-      { identity },
-      requestMetadata.jwtHeader,
-    );
-  }
-
```
sherwinski, line 277: why was this removed?

src/core/operations/TrackEventOperation.ts
fadi-george: We set a custom iso string from the caller

src/core/requestService/RequestService.ts
fadi-george: Unused

src/core/executors/RefreshUserOperationExecutor.ts
fadi-george: refresh call should have the sdk, device_os and device_module information

fadi-george: Also updated device_os to be a string.

src/shared/environment/environment.ts
```diff
@@ -61,3 +71,45 @@ const isTurbineEndpoint = (action?: string): boolean => {
...
+export function getDeviceOS(): string {
+  const environment = EnvironmentInfoHelper.getEnvironmentInfo();
+  return String(
+    isNaN(environment.browserVersion) ? -1 : environment.browserVersion,
+  );
+}
+
+export function getDeviceModel(): string {
+  return navigator.platform;
+}
```
sherwinski, line 115: I'm not sure I follow: why was this moved from FuturePushSubscriptionRecord?

fadi-george: Because its used in multiple places like custom events and subscription op

sherwinski:
Sending a custom event without a `onesignalId` or `externalId` results in a 403. In this case it should await the `onesignalId` like `addTag`.

fadi-george:
> If I enqueue a `trackEvent` call and then refresh the page (before the operation gets executed) I get the following error:
> 
> ```
> ReplayCallsOnOneSignal.ts:15 TypeError: Cannot read properties of undefined (reading 'appId')
>     at new TrackEventOperation (TrackEventOperation.ts:29:46)
>     at OperationModelStore.create (OperationModelStore.ts:71:21)
>     at OperationModelStore.load (ModelStore.ts:169:29)
>     at async OperationRepo.loadSavedOperations (OperationRepo.ts:361:11)
>     at async OperationRepo.start (OperationRepo.ts:94:5)
>     at async OneSignal2._initializeCoreModuleAndOSNamespaces (OneSignal.ts:42:5)
>     at async OneSignal2.init (OneSignal.ts:145:5)
>     at async ?app_id=77e32082-ea27-42e3-a898-c72e141824ef:40:7
>     at async ReplayCallsOnOneSignal.processOneSignalDeferredArray (ReplayCallsOnOneSignal.ts:12:15)
> ```

Should be fixed now

src/core/executors/LoginUserOperationExecutor.ts
```diff
@@ -25,6 +25,7 @@ import { LoginUserOperation } from '../operations/LoginUserOperation';
 import { type Operation } from '../operations/Operation';
 import { RefreshUserOperation } from '../operations/RefreshUserOperation';
 import { SetAliasOperation } from '../operations/SetAliasOperation';
+import { TrackEventOperation } from '../operations/TrackEventOperation';
```
sherwinski: it looks like this is unused

sherwinski (approved):

src/core/operations/TrackEventOperation.ts
```diff
@@ -0,0 +1,88 @@
...
+  }
+  set event(value: ICustomEvent) {
+    this.setProperty('event', value);
+  }
+
+  override get createComparisonKey(): string {
+    return `${this.appId}.User.${this.onesignalId}.CustomEvent`;
+  }
+  override get modifyComparisonKey(): string {
+    return `${this.appId}.User.${this.onesignalId}.CustomEvent.${this.name}`;
```
jinliu9508: Here 'name' might be referring to the name of the operation. We should use event.name instead.

src/onesignal/User.ts
```diff
@@ -276,4 +282,28 @@ export default class User {
...
+    if (!isObjectSerializable(properties)) {
+      return Log.error(
+        'Custom event properties must be a JSON-serializable object',
+      );
+    }
+
+    // check onesignalId
+    const identityModel = OneSignal.coreDirector.getIdentityModel();
+    if (!identityModel.onesignalId) {
+      identityModel.setProperty(
```
jinliu9508: Assigning a local onesignalId here may create a race condition if another process is doing the same, resulting in two onChanged events being fired consecutively. Since the operation includes a canExecute() check that returns false when the onesignalId is local, can we utilize that instead to signal a retry, rather than assigning a local onesignalId here?

src/onesignal/UserDirector.ts
```diff
@@ -63,6 +67,17 @@ export default class UserDirector {
     newIdentityModel.onesignalId = sdkId;
     newPropertiesModel.onesignalId = sdkId;
 
+    // update onesignalId for all operations in the queue
+    const opList = OneSignal.coreDirector.operationRepo.queue;
```
jinliu9508: If the CustomEventOperation was enqueued before the onesignalId was set, a successful CreateUser request will translate the backend onesignalId to all operations previously saved with the local onesignalId. Therefore, this behavior can be handled by the general mechanism we use for other operations.

vite.config.ts
```diff
@@ -102,7 +102,7 @@ export default defineConfig(({ mode }) => {
     },
     server: {
       open: true,
-      port: 4000,
+      port: 4001,
```
jinliu9508: Does this change need to be included in the commit?

src/core/executors/LoginUserOperationExecutor.ts
fadi-george: Removed.

src/core/operations/TrackEventOperation.ts
fadi-george: Updated to use event.name

vite.config.ts
fadi-george: This was for testing ill reset the value.

src/onesignal/User.ts
fadi-george: Web doesnt create a user on init, i can this move logic there to set a local id. But login and logout call `resetUserModels` which create a new local id. So now all ops with the prev local id be will stuck in limbo and will never get executed unless we have something like:
```
    const opList = OneSignal.coreDirector.operationRepo.queue;
    opList.forEach((op) => {
      if (
        IDManager.isLocalId(op.operation.onesignalId) &&
        op.operation.onesignalId === prevOnesignalId
      ) {
        op.operation.setProperty('onesignalId', sdkId, ModelChangeTags.HYDRATE);
      }
    });
```

The only weird edge case would be (for a new user w/ no data) trackEvent then logout. Should the trackEvent work for an anonymous user.

src/onesignal/UserDirector.ts
fadi-george: This logic would really only be needed for a new user with no data. So in case a trackEvent is called before login, we need to update from old local id to new local id. But if onesignal id existed and was not local, this wont do anything i.e. trackEvent (some real onesignal id) then login, these would be separate onesignal ids.

src/core/operations/TrackCustomEventOperation.ts
```diff
@@ -0,0 +1,92 @@
...
+
+  private get key(): string {
+    return `${this.appId}.User.${this.onesignalId}.CustomEvent.${this.event.name}`;
+  }
+
+  override get createComparisonKey(): string {
+    return this.key;
+  }
+  override get modifyComparisonKey(): string {
+    return this.key;
```
nan-li, line 73: `createComparisonKey` and `modifyComparisonKey` should both be nothing (aka empty string). Having this set to a value can be misleading or confusing down the line if someone was trying to figure out how we were using these keys to batch. It is effectively no different than using a random UUID.

If we do batch in the future, it'll probably be more like `${this.appId}.User.${this.onesignalId}.CustomEvent` dropping the name identifier and batch per user.

jinliu9508 (approved):

src/core/operations/TrackCustomEventOperation.ts
fadi-george: In our meeting, we said this is fine for now.

PR #1337, opened 2025-07-24
Update Packages

fadi-george:
# Description
## 1 Line Summary
Update vite packages and use recommend tsconfig settings.

## Details

# Systems Affected
   - [ ] WebSDK
   - [ ] Backend
   - [ ] Dashboard

# Validation
### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [ ] No Typescript warnings

Other:

### Checklist

---

## Related Tickets

---


tsconfig.json
```diff
@@ -1,23 +1,23 @@
...
-    "lib": ["ESNext", "DOM", "DOM.Iterable", "WebWorker"],
     "module": "ESNext",
+    "lib": ["ES2022", "DOM", "DOM.Iterable"],
     "skipLibCheck": true,
 
     /* Bundler mode */
     "moduleResolution": "bundler",
     "allowImportingTsExtensions": true,
-    "isolatedModules": true,
+    "verbatimModuleSyntax": true,
```
sherwinski, line 16: It would be good to include notes (in the PR description) on the motivation to use this flag vs `isoladedModules` e.g. how does it specific benefit developer experience, bundle size, etc. There are some comparisons in [this reference](https://www.typescriptlang.org/tsconfig/#verbatimModuleSyntax).

tsconfig.json
```diff
@@ -1,23 +1,23 @@
 {
   "compilerOptions": {
-    "target": "ES2020",
+    "target": "ES2022",
```
sherwinski: Are there any backwards compatibility concerns that we need to be aware of when targeting a newer ES version?

tsconfig.json
```diff
@@ -1,23 +1,23 @@
...
     "useDefineForClassFields": true,
-    "lib": ["ESNext", "DOM", "DOM.Iterable", "WebWorker"],
```
sherwinski, line 5: Why the choice to remove `WebWorker` here?

src/page/managers/tagManager/TagManager.ts
```diff
@@ -33,9 +36,8 @@ export default class TagManager implements ITagManager {
 
     const shouldSendUpdate = !TagUtils.isTagObjectEmpty(finalTagsObject);
     if (shouldSendUpdate) {
-      return (await OneSignal.User.addTags(
-        finalTagsObject,
-      )) as TagsObjectForApi;
+      await OneSignal.User.addTags(finalTagsObject);
+      return finalTagsObject;
```
sherwinski, line 40: Are these two functionally the same?

```
return (await OneSignal.User.addTags(
        finalTagsObject, 
      )) as TagsObjectForApi;
```

fadi-george: AddTags doesnt return anything but the outer function (sendTags) has this `Promise<TagsObjectForApi>`.

tsconfig.json
fadi-george: I still have it.

tsconfig.json
fadi-george: I changed it back to es2020

tsconfig.json
fadi-george: Saves a little of space by not using enums. And `import type` helps to different real classes from typescript types/interfaces.

PR #1338, opened 2025-07-24
Fix Bundle Bloat for Service Worker

fadi-george:
# Description
## 1 Line Summary
Reduces SW bundle size to near its original 38 kB size.
```
// 160404 bundle size
Size limit: 1 kB
Size: 631 B gzipped

Size limit: 65 kB
Size: 64.91 kB gzipped

Size limit: 37.5 kB
Size: 37.44 kB gzipped

Size limit: 8.9 kB
Size: 8.8 kB gzipped
```

// This branch (new size is 35kB):
```
  Size limit: 1 kB
  Size:       631 B gzipped
  
  Size limit: 65.2 kB
  Size:       64.02 kB gzipped
  
  Size limit: 69 kB
  Size:       34.91 kB gzipped
  
  Size limit: 8.81 kB
  Size:       8.8 kB  gzipped
 
```

## Details
- Removes environment and sdk environment class in favor of tree-shakable functions
- Updates vite config to introduce a new variable replacement: ` __IS_SERVICE_WORKER__: JSON.stringify(lib === 'worker')` and use this value in the subscribe call:
```
  public async subscribe(
    subscriptionStrategy: SubscriptionStrategyKind,
  ): Promise<RawPushSubscription> {
    let rawPushSubscription: RawPushSubscription;

    if (IS_SERVICE_WORKER) {
      rawPushSubscription =
        await this.subscribeFcmFromWorker(subscriptionStrategy);
    } else {
...
```
Since vite will set this to `true` for SW bundle. It will get rid of the else block (which uses LoginManager) 
- Updates vite config to use `treeshake: smallest` since LoginManager still has an import in the SubscriptionManager. This setting improves the dead code elimination since it will see that no method/properties of LoginManager is used in the SW bundle.

# Systems Affected
   - [x] WebSDK
   - [ ] Backend
   - [ ] Dashboard

## Screenshots
Current main:
Screenshot 2025-07-25 at 10 24 42 AM

Current branch:
Screenshot 2025-07-25 at 10 22 57 AM

v160404 (before login fix)
Screenshot 2025-07-25 at 10 27 43 AM

# Validation
### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [ ] No Typescript warnings

Other:

### Checklist

---

## Related Tickets

---

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Website-SDK/1338)


src/shared/models/Path.ts
```diff
@@ -51,12 +51,4 @@ export default class Path {
   getFullPath() {
     return this.path;
   }
-
-  getPathWithoutFileName() {
```
fadi-george, line 55: unused

__test__/support/helpers/sdkVersion.ts
```diff
@@ -5,7 +5,7 @@ export function expectHeaderToBeSent() {
     const requestInit = params[1] as RequestInit;
     const headers = requestInit.headers as Headers;
     expect(headers.get('sdk-version')).toMatch(
-      new RegExp(/^onesignal\/web\/[0-9]{6}$/),
+      new RegExp(/^onesignal\/web\/[0-9]$/),
```
sherwinski: Any particular reason for removing the string length `{6}`?

__test__/unit/models/path.test.ts
```diff
@@ -14,28 +13,24 @@ describe('Path tests', () => {
...
     expect(path.getFileName()).toBe('SeRvIcE-WoRkEr.js');
     expect(path.getFullPath()).toBe('/WeB-FoLdEr/AsSeTs/SeRvIcE-WoRkEr.js');
-    expect(path.getPathWithoutFileName()).toBe('/WeB-FoLdEr/AsSeTs');
   });
 
   test(`should return correct components for a double-extension path`, () => {
     const path = new Path('/web-folder/assets/service-worker.js.php');
     expect(path.getFileName()).toBe('service-worker.js.php');
     expect(path.getFullPath()).toBe('/web-folder/assets/service-worker.js.php');
-    expect(path.getPathWithoutFileName()).toBe('/web-folder/assets');
```
sherwinski, line 31: Why the decision to remove these assertions? And the other assertions in this test file.

src/core/executors/LoginUserOperationExecutor.ts
```diff
@@ -356,3 +355,35 @@ export class LoginUserOperationExecutor implements IOperationExecutor {
...
+      }
+      return 'zh-Hant'; // Return Chinese traditional by default
+    } else {
+      // Return the language subtag (it can be three characters, so truncate it down to 2 just to be sure)
+      return languageSubtags[0].substring(0, 2);
+    }
+  } else {
+    return 'en';
+  }
+};
```
sherwinski, line 389: Can you expand on why this `getLanguage` function is defined here? Also, why does it need specific hard-coded logic for the Chinese language?

src/shared/managers/SubscriptionManager.ts
```diff
@@ -66,6 +70,35 @@ export type SubscriptionStateServiceWorkerNotIntalled = Exclude<
...
+  // for new users, we need to create a new push subscription model and also save its push id to IndexedDB
+  if (!pushModel) {
+    pushModel =
+      OneSignal.coreDirector.generatePushSubscriptionModel(rawPushSubscription);
+    return UserDirector.createUserOnServer();
+  }
+  // for users with data failed to create a user or user + subscription on the server
+  if (IDManager.isLocalId(pushModel.id)) {
+    return UserDirector.createUserOnServer();
+  }
```
sherwinski, line 90: Is my understanding correct: if there is a pushModel ID that is `local-xxxx` then it is likely one of those users in that "stuck" state that we discussed previously?

sherwinski: Actually I just realized this function was just moved up from later on in the file. I thought it was some slight change to behavior but it seems like it's always done this.

src/sw/managers/sessionManager/SessionManager.ts
```diff
@@ -1,12 +1,16 @@
...
+    // TODO: how should it be implemented if called from inside of service worker?
   }
 
   async setupSessionEventListeners(): Promise<void> {
     // TO DO
   }
+
+  async sendOnSessionUpdateFromPage(): Promise<void> {
+    // TODO: how should it be implemented if called from inside of service worker?
+  }
```
sherwinski, line 15: Is there a plan to address these?

src/vite-env.d.ts
```diff
@@ -12,12 +12,12 @@ declare const __LOGGING__: boolean;
 
 declare const __IS_HTTPS__: boolean;
 
-declare const __API_ORIGIN__: string;
+declare var __API_ORIGIN__: string;
```
sherwinski: nit: consider using let or const

src/vite-env.d.ts
```diff
@@ -12,12 +12,12 @@ declare const __LOGGING__: boolean;
...
 
 /**
  * Which API environment to use (development, staging, production)
  */
-declare const __API_TYPE__: string;
+declare var __API_TYPE__: string;
```
sherwinski: nit: consider using let or const

__test__/unit/models/path.test.ts
fadi-george: getPathWithoutFileName is unused

src/core/executors/LoginUserOperationExecutor.ts
fadi-george: i just moved it from the environment helprs since its only used here. I don't want to alter this legacy logic.

src/sw/managers/sessionManager/SessionManager.ts
fadi-george: Ill have to ask Josh about these unimplemented methods but I only added sendOnSessionUpdateFromPage to match the interface.

src/vite-env.d.ts
fadi-george: Ill switch it back to (global as any).__API_TYPE__ for environment test

src/vite-env.d.ts
fadi-george: Switched it back.

__test__/support/helpers/sdkVersion.ts
fadi-george: We had this  vi.spyOn(Environment, 'version').mockReturnValue(160000) before. Ill just update __VERSION__ in vitest config to be 16000.

src/shared/managers/SubscriptionManager.ts
fadi-george: This is just the v160405 logic

PR #1339, opened 2025-07-25
Improve Web SDK Bundle Sizes

fadi-george:
# Description
## 1 Line Summary
Split out config helpers from having class methods to tree-shakable function exports

## Details

# Systems Affected
   - [ ] WebSDK
   - [ ] Backend
   - [ ] Dashboard

# Validation
### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [ ] No Typescript warnings

Other:

### Checklist

---

## Related Tickets

---


src/shared/config/types.ts
```diff
@@ -110,100 +93,41 @@ export interface AppUserConfigWelcomeNotification {
...
-      unsubscribe: string;
-    };
-    color: {
-      button: string;
-      text: string;
-    };
-  };
-}
-
+// server app config
```
sherwinski, line 130: nit: can probably remove this, it doesn't add much

sherwinski (approved):

PR #1345, opened 2025-08-04
SDK-2881 Break Up Static Methods and Remove Error Classes

fadi-george:
# Description
## 1 Line Summary
More optimizations to the [bundle size](https://docs.google.com/spreadsheets/d/1kbknuAcyxTrzP5IrYN9mJtWLS3oBRbzG3prTEua8yCs/edit?gid=0#gid=0).

## Details
Shave off more kBs off the bundles by breaking up static classes helpers like request service.
- moved subscriptions types/constants to /shared/subscriptions
- removed unused Log trace method
- make a shared Log class for both SW & Page
- break up context class helpers into shared/context tree-shakeable functions
- break up OneSignal errors to simpler errors with shorter messages

```bash
# SDK-2871
  Size limit: 631 B
  Size:       631 B gzipped
  
  Size limit: 60.6 kB
  Size:       60.53 kB gzipped
  
  Size limit: 25.25 kB
  Size:       25.19 kB gzipped
  
  Size limit: 8.81 kB
  Size:       8.8 kB  gzipped
```

Now:
```bash
  Size limit: 630 B
  Size:       630 B gzipped
  
  Size limit: 58.6 kB
  Size:       58.41 kB gzipped
  
  Size limit: 22.4 kB
  Size:       22.38 kB gzipped
  
  Size limit: 8.81 kB
  Size:       8.8 kB  gzipped
```

# Systems Affected
   - [x] WebSDK
   - [ ] Backend
   - [ ] Dashboard

# Validation
### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [ ] No Typescript warnings

Other:

### Checklist

---

## Related Tickets

---

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Website-SDK/1345)


src/core/requestService/index.ts
sherwinski, line 1: nit: is this file necessary? I was under the impression that barrel files can potentially cause issues with tree shaking. Curious if excluding it would have any effect on bundle size.

src/shared/config/config-version.ts
```diff
@@ -83,36 +80,21 @@ function convertConfigToVersionOne(
...
-   * Example:
-   * "promptOptions": {
-   *      "acceptButtonText": "",
-   *      "cancelButtonText": "",
-   *      "slidedown": {
-   *          "acceptButtonText": "", <--
-   *          "cancelButtonText": ""  <--
-   *      }
-   * }
-   */
```
sherwinski: Just my opinion but I think keeping these comments would be useful

src/shared/notifications/index.ts
sherwinski, line 1: same question re: barrel files

src/shared/subscriptions/index.ts
sherwinski, line 1: same question re: barrel files

src/core/executors/LoginUserOperationExecutor.ts
```diff
@@ -84,9 +83,7 @@ export class LoginUserOperationExecutor implements IOperationExecutor {
     const consentGiven = await Database.getConsentGiven();
 
     if (consentRequired && !consentGiven) {
-      throw new OneSignalError(
-        'Login: Consent required but not given, skipping login',
-      );
+      throw new Error('Consent required but not given');
```
sherwinski, line 86: Any particular reason for changing the error text itself? imo the effect of the error seemed more clear beforehand

src/shared/errors/errors.ts
```diff
@@ -0,0 +1,74 @@
...
+  new Error(`"${argName}" is empty`);
+
+export const MalformedArgumentError = (argName: string) =>
+  new Error(`"${argName}" is malformed`);
+
+export const EnumOutOfRangeArgumentError = (argName: string) =>
+  new Error(`"${argName}" is out of range`);
+
+export const WrongTypeArgumentError = (argName: string) =>
+  new Error(`"${argName}" is the wrong type`);
```
sherwinski: nit: might be nice to throw in a ` is the wrong type: expected {$expected} but got {$actual}`

src/core/executors/LoginUserOperationExecutor.ts
fadi-george: Having shorter messages would be better for reducing the bundle sizes. So striking a balance would be good.

src/shared/config/config-version.ts
fadi-george: Added back

src/shared/notifications/index.ts
fadi-george: Removed barrel files.

src/shared/errors/errors.ts
fadi-george: I'd expect people to read the docs first and that this error should be rare.

src/core/requestService/index.ts
fadi-george: Changed to requests/api

src/shared/context/types.ts
sherwinski, line 1: nit: should this go into general helpers?

PR #1349, opened 2025-08-11
fix: Ensure that notifications on legacy http integrations navigate to the origin

itrush:
# Ensure that notifications on legacy http integrations navigate to the origin

## Details
There was an issue with legacy HTTP integrations where the defaultNotificationUrl was never saved in the DB. As the result if a notification was sent without explicit launch url, the service worker logic would default to the worker's scope which is in the os.tc domain. This in turn would redirect users to OneSignal dashboard rather than to the customer's site.

To account for missing default notification url, we try to get the default URL from the app config on notification click and save it to the DB for the future use. Choosing between notification received and clicked to add this logic, decided on notification clicked to avoid doing extra api call when the user may never click the notification.

Verified in Chrome, Firefox and Safari that when clicking on a notification it redirects to the origin instead of the os.tc domain.

Note: will only merge it after Thanksgiving

# Systems Affected
   - [x] WebSDK
   - [ ] Backend
   - [ ] Dashboard

# Validation
## Tests

There seem to be no tests for the service worker flows so didn't add any new ones 🤔  Did an extensive manual test in Chrome, Safari and Firefox.

### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [x] No Typescript warnings

Other:

### Info

https://github.com/user-attachments/assets/b3302082-f07f-4572-93a6-259263723986

### Checklist

---

## Related Tickets
https://app.asana.com/1/780103692902078/project/1211950939908862/task/1210115044974433?focus=true

---

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Website-SDK/1349)


src/libraries/WorkerMessenger.ts
```diff
@@ -192,7 +192,7 @@ export class WorkerMessenger {
     Log.debug(`(${location.origin}) [Worker Messenger] Page is now listening for messages.`);
   }
 
-  onWorkerMessageReceivedFromPage(event: ServiceWorkerMessageEvent) {
+  onWorkerMessageReceivedFromPage(event: any) {
```
jkasten2: Instead of `any` I believe we can switch to `ExtendableMessageEvent` instead.

src/libraries/WorkerMessenger.ts
```diff
@@ -235,7 +235,7 @@ export class WorkerMessenger {
   message topic. If no one is listening to the message, it is discarded;
   otherwise, the listener callback is executed.
   */
-  onPageMessageReceivedFromServiceWorker(event: ServiceWorkerMessageEvent) {
+  onPageMessageReceivedFromServiceWorker(event: any) {
```
jkasten2: Instead of `any` I believe we can switch to `ExtendableMessageEvent` instead.

jkasten2 (changes requested):
Logic looks good to me, just one comment on types.

src/libraries/WorkerMessenger.ts
itrush: Updated

src/libraries/WorkerMessenger.ts
itrush: Updated

PR #1355, opened 2025-08-20
Allow Adding Sms / Email before Login

fadi-george:
# Description
## 1 Line Summary
- Allow adding email before login

## Details
- On user initialize, well check for a onesignal. If it doesnt exists, we will create a local id.
- For addSms or addEmail we will we need to enqueue a login operation if one isnt already in queue. If there are two login operations (either for addSms + addEmail or login + addEmail) then it will error.
- Trackevent still needs a login operation before it otherwise the very first track event op for a new user (local onesignal id) will get stuck and never executed.
- User tags (as how it was before) will get erased if user is logged in later.

# Systems Affected
   - [x] WebSDK
   - [ ] Backend
   - [ ] Dashboard

# Validation
### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [ ] No Typescript warnings

Other:

### Checklist

---

## Related Tickets

---

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Website-SDK/1355)

### DIFF
--- a/src/core/executors/RefreshUserOperationExecutor.test.ts
+++ b/src/core/executors/RefreshUserOperationExecutor.test.ts
@@ -183,7 +183,7 @@ describe('RefreshUserOperationExecutor', () => {
       pushSubModel.token = PUSH_TOKEN;
       pushSubModel.notification_types = NotificationType.Subscribed;
 
-      subscriptionModelStore.add(pushSubModel, ModelChangeTags.NO_PROPOGATE);
+      subscriptionModelStore.add(pushSubModel, ModelChangeTags.NO_PROPAGATE);
       await setPushToken(PUSH_TOKEN);
 
       const executor = getExecutor();
--- a/src/onesignal/User.ts
+++ b/src/onesignal/User.ts
@@ -35,7 +35,7 @@ export default class User {
         identityModel.setProperty(
           IdentityConstants.ONESIGNAL_ID,
           onesignalId,
-          ModelChangeTags.NO_PROPOGATE,
+          ModelChangeTags.NO_PROPAGATE,
         );
       }
     }
--- a/__test__/support/environment/TestEnvironmentHelpers.ts
+++ b/__test__/support/environment/TestEnvironmentHelpers.ts
@@ -90,7 +90,7 @@ export const setupSubModelStore = async ({
   await setPushToken(pushModel.token);
   OneSignal.coreDirector.subscriptionModelStore.replaceAll(
     [pushModel],
-    ModelChangeTags.NO_PROPOGATE,
+    ModelChangeTags.NO_PROPAGATE,
   );
 
   return pushModel;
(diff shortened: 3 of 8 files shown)


src/core/executors/RefreshUserOperationExecutor.test.ts
```diff
@@ -177,7 +176,6 @@ describe('RefreshUserOperationExecutor', () => {
     });
 
     test('should preserve cached push subscription when updating models', async () => {
-      setUpdateSubscriptionResponse({ subscriptionId: SUB_ID_2 });
```
fadi-george, line 180: getting rid of extraneous calls

src/page/managers/LoginManager.ts
```diff
@@ -39,33 +43,36 @@ export default class LoginManager {
...
-      new LoginUserOperation(
-        appId,
-        newIdentityOneSignalId,
-        externalId,
-        !currentExternalId ? currentOneSignalId : undefined,
       ),
-    );
+    ];
+
+    await Promise.all(promises);
```
sherwinski, line 75: Just double checking, do we not care about the order in which `TransferSubscriptionOperation` and `LoginUserOperation` are run?

src/onesignal/User.ts
```diff
@@ -43,9 +43,8 @@ export default class User {
     return User.singletonInstance;
   }
 
-  get onesignalId(): string | undefined {
-    const oneSignalId = OneSignal.coreDirector.getIdentityModel().onesignalId;
-    return !IDManager.isLocalId(oneSignalId) ? oneSignalId : undefined;
+  get onesignalId(): string {
```
sherwinski, line 46: My only worry about this is ensuring that we have a `IDManager.isLocalId(oneSignalId)` check anywhere needed that invokes this getter

src/onesignal/User.ts
```diff
@@ -25,14 +29,22 @@ export default class User {
...
     if (!User.singletonInstance) {
       User.singletonInstance = new User();
+      const identityModel = OneSignal.coreDirector.getIdentityModel();
+      if (!identityModel.onesignalId) {
+        const onesignalId = IDManager.createLocalId();
+        identityModel.setProperty(
+          IdentityConstants.ONESIGNAL_ID,
+          onesignalId,
+          ModelChangeTags.NO_PROPOGATE,
+        );
```
sherwinski, line 39: Non-blocking and very nit but I just realized: shouldn't this be `NO_PROPAGATE`?

src/onesignal/User.ts
fadi-george: Its a public api but its only used here.

src/onesignal/User.ts
fadi-george: NO_PROPOGATE means dont trigger api call changes

fadi-george: Youre right, updated!

src/page/managers/LoginManager.ts
fadi-george: Technically it doesnt matter since the operations get grouped anyway.

sherwinski (approved):

PR #1363, opened 2025-08-27
Wait for user to provide consent before executing operations

fadi-george:
# Description
## 1 Line Summary
Addresses: https://app.asana.com/1/780103692902078/project/1198177413178126/task/1211152956512052?focus=true
- move consent logic out of login operation executor and move it to operation repo

## Details
Current erorr:
Screenshot 2025-08-27 at 12 16 03 PM

- Currently if the developer sets consent required to true but user never gives consent. The display notification is still shown (but subscription is never created. Additionally the Login Operation executor throws an error. 

Changes:
- Updates default test context to not have prompts manager autoprompt to not trigger operations like login
- Adds new `isConsentRequiredButNotGiven` helper to check against consent required and consent given flags.
- Adds new prop `OneSignal._consentGiven` which is loaded in init . This is needed since we dont we to add delay for the login call and don want to make actions like addTags async.
- Removes 'consentGiven' from db schema as this was a typo when it shouldve been 'userConsent'
- Add isConsentRequiredButNotGiven checks to `addTag`, `addTags`, `addEmail`, `addSms`, etc.
- Adds user config check `requiresUserPrivacyConsent` to `getConsentRequired` call.

# Systems Affected
   - [x] WebSDK
   - [ ] Backend
   - [ ] Dashboard

# Validation
### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [ ] No Typescript warnings

Other:

### Checklist

---

## Related Tickets

---

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Website-SDK/1363)

### DIFF
--- a/package.json
+++ b/package.json
@@ -84,7 +84,7 @@
     },
     {
       "path": "./build/releases/OneSignalSDK.page.es6.js",
-      "limit": "52.15 kB",
+      "limit": "52.143 kB",
       "gzip": true
     },
     {
--- a/package.json
+++ b/package.json
@@ -5,8 +5,7 @@
   "type": "module",
   "dependencies": {
     "idb": "^8.0.3",
-    "jsonp": "github:OneSignal/jsonp#onesignal",
-    "uuid": "^11.1.0"
+    "jsonp": "github:OneSignal/jsonp#onesignal"
   },
   "scripts": {
     "dev": "LOGGING=false vite",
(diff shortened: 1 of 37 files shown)


package.json
```diff
@@ -5,8 +5,7 @@
   "type": "module",
   "dependencies": {
     "idb": "^8.0.3",
-    "jsonp": "github:OneSignal/jsonp#onesignal",
-    "uuid": "^11.1.0"
```
sherwinski, line 9: just double checking that we don't use this anywhere?

sherwinski (approved):

package.json
sherwinski: I assume we replaced it with `crypto.randomUUID()`?

fadi-george: Yes we use crypto.

PR #1369, opened 2025-09-02
More Size Reduction via _ Prefix

fadi-george:
# Description
## 1 Line Summary
- add more _ prefix for to class methods 

## Details
- The vite config settings currently use terser with a mangleProps that uses _ to denote which constants/properties to minify/mangle for helping with reducing bundle size

Main:
```
  Size limit: 450 B
  Size:       450 B gzipped
  
  Size limit: 52.18 kB
  Size:       52.18 kB gzipped
  
  Size limit: 14.54 kB
  Size:       14.54 kB gzipped
  
  Size limit: 8.81 kB
  Size:       8.8 kB  gzipped
```

Branch:
```
  Size limit: 450 B
  Size:       450 B gzipped
  
  Size limit: 51.85 kB
  Size:       51.85 kB gzipped
  
  Size limit: 13.93 kB
  Size:       13.93 kB gzipped
  
  Size limit: 8.81 kB
  Size:       8.8 kB  gzipped
```

# Systems Affected
   - [x] WebSDK
   - [ ] Backend
   - [ ] Dashboard

# Validation
### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [ ] No Typescript warnings

Other:

### Checklist

---

## Related Tickets

---

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Website-SDK/1369)


src/page/bell/Bell.ts
```diff
@@ -541,26 +539,22 @@ export default class Bell {
...
-        'style',
-        `filter: ${badgeShadow}; -webkit-filter: ${badgeShadow};`,
-      );
-      this.dialog.element.setAttribute(
-        'style',
-        `filter: ${dialogShadow}; -webkit-filter: ${dialogShadow};`,
-      );
-    } else {
-      this.badge.element.setAttribute('style', `display: none;`);
-    }
```
sherwinski, line 563: what's the reason for removing this?

sherwinski (approved):

src/page/bell/Bell.ts
fadi-george: Its very legacy code and safari can show boxshadows

PR #1370, opened 2025-09-02
Return undefined for local onesignal id

fadi-george:
# Description
## 1 Line Summary
- Return undefined for local onesignal id and push subscription id

## Details
- Update User and PushSubscription class to return undefined if id is local

# Systems Affected
   - [x] WebSDK
   - [ ] Backend
   - [ ] Dashboard

# Validation
### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [ ] No Typescript warnings

Other:

### Checklist

---

## Related Tickets

---

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Website-SDK/1370)

### DIFF
--- a/src/onesignal/User.ts
+++ b/src/onesignal/User.ts
@@ -30,11 +30,11 @@ export default class User {
   static createOrGetInstance(): User {
     if (!User.singletonInstance) {
       User.singletonInstance = new User();
-      const identityModel = OneSignal.coreDirector.getIdentityModel();
-      const propertiesModel = OneSignal.coreDirector.getPropertiesModel();
+      const identityModel = OneSignal._coreDirector._getIdentityModel();
+      const propertiesModel = OneSignal._coreDirector._getPropertiesModel();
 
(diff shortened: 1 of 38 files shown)


src/onesignal/User.ts
```diff
@@ -55,8 +55,9 @@ export default class User {
     return User.singletonInstance;
   }
 
-  get onesignalId(): string {
-    return OneSignal.coreDirector.getIdentityModel().onesignalId;
+  get onesignalId(): string | undefined {
+    const onesignalId = OneSignal._coreDirector._getIdentityModel().onesignalId;
+    return IDManager._isLocalId(onesignalId) ? undefined : onesignalId;
```
jkasten2, line 60: nit, on splitting up commits. A logic change like this mixed in with a refactor (a rename in this case) makes this hard to spot.

jkasten2 (approved):

PR #1372, opened 2025-09-03
Reduce Bundle Size

fadi-george:
# Description
## Details

# Systems Affected
   - [ ] WebSDK
   - [ ] Backend
   - [ ] Dashboard

# Validation
### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [ ] No Typescript warnings

Other:

### Checklist

---

## Related Tickets

---


src/shared/api/base.ts
```diff
@@ -122,7 +121,7 @@ async function executeFetch<T = unknown>(
     };
   } catch (e) {
     if (e instanceof Error && e.name === 'TypeError') {
-      await delay(RETRY_BACKOFF[retry]);
```
sherwinski: This change doesn't seem consistent with the old backoff strategy:

```ts
export const RETRY_BACKOFF: { [key: number]: number } = {
  5: 10_000,
  4: 20_000,
  3: 30_000,
  2: 30_000,
  1: 30_000,
};
```

Wouldn't this just increase by 10000 ms each time after the update?

sherwinski (changes requested):
I guess I'm a bit confused about one thing: we're opting to wildcard import all of these modules but I was under the impression that doing so would interfere with tree shaking? Am I missing something here?

src/shared/api/base.ts
fadi-george: Fixed it inline

src/shared/api/base.ts
```diff
@@ -121,7 +121,8 @@ async function executeFetch<T = unknown>(
     };
   } catch (e) {
     if (e instanceof Error && e.name === 'TypeError') {
-      await delay(retry * 1e4);
+      // start with 30 seconds, then 20 seconds, then 10 seconds
```
sherwinski: Doesn't the backoff strategy start with 10, then 20, then 30 seconds? Since we're decrementing `retry`.

fadi-george: Oops updated comment

PR #1373, opened 2025-09-03
Simplify bell elements

fadi-george:
# Description
## Details

# Systems Affected
   - [ ] WebSDK
   - [ ] Backend
   - [ ] Dashboard

# Validation
### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [ ] No Typescript warnings

Other:

### Checklist

---

## Related Tickets

---


src/page/bell/Button.ts
```diff
@@ -2,154 +2,170 @@ import { addDomElement, removeDomElement } from 'src/shared/helpers/dom';
...
+    this._activate();
+    this._bell._badge._activate();
   }
 
-  onEndTap() {
-    this.inactivate();
-    this.bell.badge.inactivate();
+  _onEndTap() {
+    this._inactivate();
+    this._bell._badge._inactivate();
```
sherwinski: Including changes like these in the same commit make it difficult to parse out the parts that actually affect Bell behavior. I generally recommend keeping any refactoring changes to their own separate commits.

src/page/bell/constants.ts
```diff
@@ -11,11 +11,11 @@ export const SUBSCRIBE_BUTTON_ID = 'subscribe-button';
...
-  UnsubscribeClick: 'notifyButtonUnsubscribeClick',
-  Hovering: 'notifyButtonHovering',
-  Hovered: 'notifyButtonHover',
+  StateChanged: 'notify0',
+  LauncherClick: 'notify1',
+  BellClick: 'notify2',
+  SubscribeClick: 'notify3',
+  UnsubscribeClick: 'notify4',
+  Hovering: 'notify5',
+  Hovered: 'notify6',
```
sherwinski, line 20: I'm not sure I understand the purpose of this change? Is it meant to help with bundle size by shortening the values?

Issue #409, opened 2018-09-20
Subscription bell and popup are transparent in opened state


floppyvlaDISK:
**Description:**
Sometimes when double clicking on subscription bell repeatedly bell widget and popup become out-of-sync, i.e. bell icon gets smaller(non-opened) but popup is still shown, and both are transparent

**Steps to Reproduce Issue:**

1. go to https://onesignal.com/
1. double clicking few times on subscription bell

**Problem:** Sometimes subscription bell is in closed(small) state, but popup is shown, both being transparent.

**Anything else:** Attaching a screenshot 
![selection!UNITO-UNDERSCORE!113](https://user-images.githubusercontent.com/25643386/45814760-4370e080-bcdf-11e8-97b1-d3e31ec07a2b.png)

jkasten2:
@floppyvlaDISK Thanks for reporting, I was able to reproduce the issue as well. Tagged this as a reproducible bug

jkasten2:
Still reproducible in version 150707, however the timing is very specific to reproduce so marking as low priority.

PR #1375, opened 2025-09-08
SDK-2923 Fix Bell Animation

fadi-george:
# Description
## 1 Line Summary
- Addresses #409 when you're clicking the notification bell multiple times in quick succession.

Before:

https://github.com/user-attachments/assets/a1d40d89-1f12-4b9c-9847-e845c8de2500

After:

https://github.com/user-attachments/assets/244c7f0b-60f0-4539-9a49-045f6b255616

## Details
- Combine active animated element and animated element into one class
- Uses modern element getAnimations api instead of timeout
- Removes internal animated element trigger events.

# Systems Affected
   - [x] WebSDK
   - [ ] Backend
   - [ ] Dashboard

# Validation
### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [ ] No Typescript warnings

Other:

### Checklist

---

## Related Tickets

---

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Website-SDK/1375)

### DIFF
--- a/src/page/bell/Badge.ts
+++ b/src/page/bell/Badge.ts
@@ -1,9 +1,6 @@
-import {
-  default as ActiveAnimatedElement,
-  default as AnimatedElement,
-} from './AnimatedElement';
+import AnimatedElement from './AnimatedElement';
 
-export default class Badge extends ActiveAnimatedElement {
+export default class Badge extends AnimatedElement {
   constructor() {
     super(
       '.onesignal-bell-launcher-badge',


src/page/bell/Badge.ts
```diff
@@ -1,15 +1,14 @@
-import ActiveAnimatedElement from './ActiveAnimatedElement';
-import AnimatedElement from './AnimatedElement';
+import {
+  default as ActiveAnimatedElement,
+  default as AnimatedElement,
+} from './AnimatedElement';
```
sherwinski: Is this valid syntax? Or would it cause a conflict?

fadi-george: Oops Ill fix

PR #1378, opened 2025-09-10
SDK-2937 Automate portion of web cd pipeline

fadi-george:
# Description
## 1 Line Summary
- automate creation of turbine pr
Screenshot 2025-09-10 at 4 08 41 PM

## Details
- adds script to the cd workflow to automate creating turbine pr:
  - clones turbine repo and creates/update the sdk release branch
  - creates the pr diff with the updated commit sha for user model field

# Systems Affected
   - [x] WebSDK
   - [ ] Backend
   - [ ] Dashboard

# Validation
### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [ ] No Typescript warnings

Other:

### Checklist

---

## Related Tickets

---

### DIFF
--- a/build/scripts/turbine.sh
+++ b/build/scripts/turbine.sh
@@ -40,19 +40,17 @@ JOB_MESSAGE="Job that successfully built the artifacts: https://github.com/OneSi
 git add .circleci/config.yml
 git commit -m "$RELEASE_MESSAGE" -m "$JOB_MESSAGE"
 
-# Push to origin
-git push -u origin "$BRANCH_NAME"
+# Push to origin (override any existing branch)
+git push --force origin "$BRANCH_NAME"
 
-exit 0
-
-# Create pull request - use full owner:branch format for head
+# Create pull request
 curl -sS -X POST \
   -H "Authorization: Bearer $GH_TURBINE_TOKEN" \
   -H "Content-Type: application/json" \
   -d "{
-\"title\": \"$RELEASE_MESSAGE\",
-\"body\": \"$JOB_MESSAGE\",
-\"head\": \"OneSignal:$BRANCH_NAME\",
-\"base\": \"main\"
-}" \
+    \"title\": \"$RELEASE_MESSAGE\",
+    \"body\": \"$JOB_MESSAGE\",
+    \"head\": \"OneSignal:$BRANCH_NAME\",
+    \"base\": \"main\"
+  }" \
   https://api.github.com/repos/OneSignal/turbine/pulls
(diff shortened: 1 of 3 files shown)


build/scripts/turbine.sh
```diff
@@ -0,0 +1,68 @@
+#!/bin/bash
```
abdulraqeeb33, line 1: @fadi-george  couldn't we make this work in the GA? instead of having a standalone script?

build/scripts/turbine.sh
```diff
@@ -0,0 +1,68 @@
...
+
+git clone "https://${GH_TURBINE_TOKEN}@github.com/OneSignal/turbine.git" turbine
+cd turbine
+
+# Configure git identity
+git config user.name "github-actions[bot]"
+git config user.email "github-actions[bot]@users.noreply.github.com"
+
+# Create new branch
+BRANCH_NAME="web-sdk-${SDK_VERSION}-release"
```
abdulraqeeb33, line 12: also, do we have a pre-defined pattern for creating release branches?
I used `rel/x.x.x` for Android, this will help in reusing other steps in a GA.

package.json
```diff
@@ -36,7 +36,7 @@
     "lint": "eslint src --ext .js,.jsx,.ts,.tsx; prettylint 'src/**/*' 'test/**/*' '__test__/**/*' --no-editorconfig"
   },
   "config": {
-    "sdkVersion": "160508"
+    "sdkVersion": "160509"
```
abdulraqeeb33: I am guessing you updated this manually for this PR. but for a release PR, it will automatically increment it right?

build/scripts/turbine.sh
fadi-george: The script is kind of long, I didn't want it to be all in the github acitons

package.json
fadi-george: Just for testing can remove change after approval. We don't have a good process for incrementing automatically.

build/scripts/turbine.sh
jkasten2: @abdulraqeeb33 the branch is created on a different repo, turbine, we need to include web sdk in the branch name somewhere. Turbine does more than host the Web SDK.

fadi-george: Was following this pattern, https://docs.google.com/document/d/1ekfGZHxYyu04Z5VcE_8vkTWJEQm2e8Ndp-hlFedFM-Q/edit?tab=t.0#heading=h.u6bz69tu00ad
but I suppose the branch could be that or rel/xxyyzz

fadi-george: Not sure what we can really reuse since this one goes through turbine

jkasten2: We can't use `rel/x.x.x` in the turbine repo, it serves much more than the Web SDK. So if we change it could be something like `web-sdk-rel/xxyyzz`

fadi-george: Or can keep it as is.

Issue #1198, opened 2024-09-24
[Question]: how to tell when Onesignal.Notifications.requestPermission() fails?


ggrochow-dh:
### How can we help?

Assuming I have blocked browser notification permission for the site I am doing this on.

when calling `Onesignal.Notifications.requestPermission()` I see an error output in the console, but get 0 feedback from the function call, and im unable to catch the error.

How can I tell when this has failed? I want to be able to instruct the user they have to manually change their browser settings, but it seems like there is no way to get any feedback around this very critical part of this whole process not working.

fadi-george:
Fix should now be available in 160509.

PR #1379, opened 2025-09-11
fix: expose errors from request permission

fadi-george:
# Description
## 1 Line Summary
- Addresses issue #1198 by exposing errors when request permission fails

## Details
- Removes try catch wrapper for subscription helper's `internalRegisterForPush` to expose errors
- Changes operation repo resolver to reject if value is false
- Rejects the enqueue resolve for operations that pause the operation repo
- Adds new tests for notification request permissions 

# Systems Affected
   - [x] WebSDK
   - [ ] Backend
   - [ ] Dashboard

# Validation
### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [ ] No Typescript warnings

Other:

### Checklist

---

## Related Tickets

---

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Website-SDK/1379)

### DIFF
--- a/src/core/operationRepo/constants.ts
+++ b/src/core/operationRepo/constants.ts
@@ -3,4 +3,4 @@ export const OP_REPO_DEFAULT_FAIL_RETRY_BACKOFF = 1000;
 export const OP_REPO_POST_CREATE_DELAY = 500;
 export const OP_REPO_EXECUTION_INTERVAL = 500;
 export const OP_REPO_POST_CREATE_RETRY_UP_TO = 15_000;
-export const RETRY_SECONDS = 10_000;
+export const RETRY_MS = 10_000;
--- a/src/shared/helpers/SubscriptionHelper.ts
+++ b/src/shared/helpers/SubscriptionHelper.ts
@@ -6,6 +6,7 @@ import type {
   SubscriptionChannelValue,
   SubscriptionTypeValue,
 } from 'src/shared/subscriptions/types';
+import Log from '../libraries/Log';
 import { checkAndTriggerSubscriptionChanged } from '../listeners';
 import { Subscription } from '../models/Subscription';
 import { SubscriptionStrategyKind } from '../models/SubscriptionStrategyKind';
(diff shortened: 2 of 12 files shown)


src/shared/helpers/SubscriptionHelper.ts
```diff
@@ -24,22 +23,17 @@ export default class SubscriptionHelper {
...
+    incrementPageViewCount();
+
+    await triggerNotificationPermissionChanged();
+    await checkAndTriggerSubscriptionChanged();
 
-      await triggerNotificationPermissionChanged();
-      await checkAndTriggerSubscriptionChanged();
-    } catch (e) {
-      Log._error(e);
-    }
```
jkasten2: Bubbling up exception publicly is a breaking change, it will cause customer's code after `Onesignal.Notifications.requestPermission()` to no longer run when this throws.

Throwing is also different behavior than what the mobile SDKs do, they return a boolean instead.

Can we change `Promise<void>` to instead return `Promise<boolean>` to match closer to the mobile SDKs public interface?

src/core/operationRepo/constants.ts
```diff
@@ -3,3 +3,4 @@ export const OP_REPO_DEFAULT_FAIL_RETRY_BACKOFF = 1000;
 export const OP_REPO_POST_CREATE_DELAY = 500;
 export const OP_REPO_EXECUTION_INTERVAL = 500;
 export const OP_REPO_POST_CREATE_RETRY_UP_TO = 15_000;
+export const RETRY_SECONDS = 10_000;
```
jkasten2: This should be milliseconds / ms, not seconds.

src/shared/helpers/SubscriptionHelper.ts
fadi-george: Yeah that makes sense.

src/core/operationRepo/constants.ts
fadi-george: True.

PR #1380, opened 2025-09-12
chore(api): add `User.trackEvent` definition to API spec

sherwinski:
# Description
## 1 Line Summary
Adds the definition for `User.trackEvent` to the API spec (see more details [on docs](https://documentation.onesignal.com/docs/web-sdk-reference#trackevent)). This will then generate and push the definitions downstream to https://github.com/OneSignal/web-shim-codegen and the web wrappers.

## Details

# Systems Affected
   - [x] WebSDK
   - [ ] Backend
   - [ ] Dashboard

# Validation
### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [x] No Typescript warnings

Other:

### Checklist

---

## Related Tickets

---

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Website-SDK/1380)

### DIFF
--- a/src/page/managers/slidedownManager/SlidedownManager.test.ts
+++ b/src/page/managers/slidedownManager/SlidedownManager.test.ts
@@ -314,6 +314,7 @@ export const mockPhoneLibraryLoading = () => {
 
     window.intlTelInputUtils = {
       numberType: { MOBILE: 1 },
+      // @ts-expect-error - mock intl-tel-input
       numberFormat: { E164: 0 },
     };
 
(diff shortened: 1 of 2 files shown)


src/page/managers/slidedownManager/SlidedownManager.test.ts
```diff
@@ -314,7 +314,6 @@ export const mockPhoneLibraryLoading = () => {
 
     window.intlTelInputUtils = {
       numberType: { MOBILE: 1 },
-      // @ts-expect-error - mock intl-tel-input
```
fadi-george: broke tests

PR #1388, opened 2025-09-16
chore: automate generating a release when release branches are merged

fadi-george:
# Description
## 1 Line Summary
- updates cd workflow to handle merge events from release branches

## Details
- updates cd workflow to handle workflow dispatch for merges of release branches
- adds job to check if action is valid
- adds job to run create-release-on-githun workflow

# Systems Affected
   - [x] WebSDK
   - [ ] Backend
   - [ ] Dashboard

# Validation
### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [ ] No Typescript warnings

Other:

### Checklist

---

## Related Tickets

---

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Website-SDK/1388)

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -1,11 +1,50 @@
 name: Build SDK Files & GCS Upload
 
 on:
+  push:
+    branches:
+      - main
+    paths-ignore:
+      - '**.md'
   workflow_dispatch:
 
 jobs:
+  check_release:
+    runs-on: ubuntu-22.04
+    outputs:
+      should_run: ${{ steps.check_release.outputs.should_run }}
+    steps:
+      - name: Check if release merge or workflow dispatch
+        id: check_release
+        uses: actions/github-script@v8
+        with:
+          script: |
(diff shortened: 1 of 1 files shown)


.github/workflows/cd.yml
```diff
@@ -1,11 +1,38 @@
...
+          script: |
+            // Allow workflow_dispatch and release PR merges
+            if (context.eventName !== 'push') {
+              core.setOutput('should_run', 'true');
+              return;
+            }
+
+            // For push events, check if it's a release merge
+            const isReleaseMerge = /Release \d{6}/.test(context.payload.head_commit.message);
+            core.setOutput('should_run', isReleaseMerge ? 'true' : 'false');
```
jkasten2: Instead of using this `should_run` variable can't we just fail the build instead to simplify this?

I think it is something like `core.setFailed(error.message);` for JS, or `exit 1` if it is a bash script works.

fadi-george: Made the changes

PR #1392, opened 2025-09-17
fix create release on github permissions

fadi-george:
# Description
## 1 Line Summary
Clean up create release notes workflow

## Details
- update create release notes workflow to clean up pr body notes (to remove reviewable section)
- update cd workflow to not error for pushes on main

# Systems Affected
   - [x] WebSDK
   - [ ] Backend
   - [ ] Dashboard

# Validation
### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [ ] No Typescript warnings

Other:

### Checklist

---

## Related Tickets

---

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Website-SDK/1392)

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -11,22 +11,29 @@ on:
 jobs:
   check_release:
     runs-on: ubuntu-22.04
+    outputs:
+      should_run: ${{ steps.check.outputs.should_run }}
     steps:
       - name: Check if release merge or workflow dispatch
-        id: check_release
+        id: check
         uses: actions/github-script@v8
         with:
           script: |
-            // Allow workflow_dispatch and release PR merges
-            if (context.eventName !== 'push') return;
+            let shouldRun = false;
 
-            // For push events, check if it's a release merge
-            const isReleaseMerge = /Release \d{6}/.test(context.payload.head_commit.message);
-            if (!isReleaseMerge) core.setFailed('Not a release merge - skipping workflow');
+            if (context.eventName === 'workflow_dispatch') {
+              shouldRun = true;
+            } else if (context.eventName === 'push') {
+              const isReleaseMerge = /Release \d{6}/.test(context.payload.head_commit.message);
+              shouldRun = isReleaseMerge;
+            }
+
+            core.setOutput('should_run', shouldRun);
 
   test_build_deploy:
     runs-on: ubuntu-22.04
     needs: check_release
+    if: needs.check_release.outputs.should_run == 'true'
     steps:
       - uses: actions/checkout@v5
       - name: Use Node.js
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -19,15 +19,8 @@ jobs:
         uses: actions/github-script@v8
         with:
           script: |
-            let shouldRun = false;
-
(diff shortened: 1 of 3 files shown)


.github/workflows/cd.yml
```diff
@@ -11,22 +11,29 @@ on:
...
-            const isReleaseMerge = /Release \d{6}/.test(context.payload.head_commit.message);
-            if (!isReleaseMerge) core.setFailed('Not a release merge - skipping workflow');
+            if (context.eventName === 'workflow_dispatch') {
+              shouldRun = true;
+            } else if (context.eventName === 'push') {
+              const isReleaseMerge = /Release \d{6}/.test(context.payload.head_commit.message);
+              shouldRun = isReleaseMerge;
+            }
+
+            core.setOutput('should_run', shouldRun);
```
jkasten2, line 24: Why not use `core.setFailed` instead?

.github/workflows/cd.yml
```diff
@@ -11,22 +11,29 @@ on:
...
 
   test_build_deploy:
     runs-on: ubuntu-22.04
     needs: check_release
```
jkasten2, line 28: id was changed to `check` but this wasn't updated.

.github/workflows/cd.yml
fadi-george: That would make it seem like theres something wrong with the workflow:
Screenshot 2025-09-17 at 1 32 13 PM

.github/workflows/cd.yml
fadi-george: the job is check_release

.github/workflows/cd.yml
jkasten2: ah I see, is there a skipped status we can use instead?

.github/workflows/cd.yml
jkasten2: I see, the `id` was renamed but didn't know that this `needs:` doesn't use it. I'll leave a separate comment then.

.github/workflows/cd.yml
```diff
@@ -11,22 +11,29 @@ on:
 jobs:
   check_release:
     runs-on: ubuntu-22.04
+    outputs:
+      should_run: ${{ steps.check.outputs.should_run }}
     steps:
       - name: Check if release merge or workflow dispatch
-        id: check_release
+        id: check
```
jkasten2, line 18: Do we need to set an `id:` here?

If this is needed for outputs, then this line below probably needs to be renamed too:
```
if: needs.check_release.outputs.should_run == 'true'
```

jkasten2 (changes requested):

.github/workflows/cd.yml
fadi-george: I don't believe so.

.github/workflows/cd.yml
fadi-george: its used here (line 14)
    outputs:
      should_run: ${{ steps.check.outputs.should_run }}

jkasten2: I see, I'll mark this thread as resolved. The similar names and not knowing at first Github actions use configuration over convention was throwing me off.

jkasten2 (approved):

PR #1397, opened 2025-10-14
chore: improve typings for bell elements

fadi-george:
# Description
## 1 Line Summary
- improve types for Bell elements and get rid of any usages

## Details
- replace any usages with actual element types
- split up constants used by bell message badge etc. to remove cyclical dependencies (apart from types)

# Systems Affected
   - [x] WebSDK
   - [ ] Backend
   - [ ] Dashboard

# Validation
### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [ ] No Typescript warnings

Other:

### Checklist

---

## Related Tickets

---

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Website-SDK/1397)

### DIFF
--- a/src/page/bell/Bell.ts
+++ b/src/page/bell/Bell.ts
@@ -28,14 +28,12 @@ import Message from './Message';
 import {
   BellEvent,
   BellState,
-  type BellStateType,
+  type BellStateValue,
   MESSAGE_TIMEOUT,
-  MesageType,
+  MessageType,
 } from './constants';
 
 const logoSvg = `<svg class="onesignal-bell-svg" xmlns="http://www.w3.org/2000/svg" width="99.7" height="99.7" viewBox="0 0 99.7 99.7"><circle class="background" cx="49.9" cy="49.9" r="49.9"/><path class="foreground" d="M50.1 66.2H27.7s-2-.2-2-2.1c0-1.9 1.7-2 1.7-2s6.7-3.2 6.7-5.5S33 52.7 33 43.3s6-16.6 13.2-16.6c0 0 1-2.4 3.9-2.4 2.8 0 3.8 2.4 3.8 2.4 7.2 0 13.2 7.2 13.2 16.6s-1 11-1 13.3c0 2.3 6.7 5.5 6.7 5.5s1.7.1 1.7 2c0 1.8-2.1 2.1-2.1 2.1H50.1zm-7.2 2.3h14.5s-1 6.3-7.2 6.3-7.3-6.3-7.3-6.3z"/><ellipse class="stroke" cx="49.9" cy="49.9" rx="37.4" ry="36.9"/></svg>`;
-
-type BellState = 'uninitialized' | 'subscribed' | 'unsubscribed' | 'blocked';
 
 const DEFAULT_SIZE: BellSize = 'medium';
 const DEFAULT_POSITION: BellPosition = 'bottom-right';
@@ -43,7 +41,7 @@ const DEFAULT_THEME = 'default';
 
 export default class Bell {
   public options: AppUserConfigNotifyButton;
-  public state: BellStateType = BellState._Uninitialized;
+  public state: BellStateValue = BellState._Uninitialized;
   public _ignoreSubscriptionState = false;
   public hovering = false;
   public initialized = false;
@@ -190,7 +188,7 @@ export default class Bell {
         })
         .then(() => {
           return this.message.display(
-            MesageType._Message,
+            MessageType._Message,
             this.options.text['message.action.resubscribed'],
             MESSAGE_TIMEOUT,
           );
@@ -226,7 +224,7 @@ export default class Bell {
         })
         .then(() => {
           return this.message.display(
-            MesageType._Message,
+            MessageType._Message,
             this.options.text['message.action.unsubscribed'],
             MESSAGE_TIMEOUT,
           );
@@ -248,7 +246,7 @@ export default class Bell {
 
       // If the message is a message and not a tip, don't show it (only show tips)
       // Messages will go away on their own
-      if (this.message.contentType === MesageType._Message) {
+      if (this.message.contentType === MessageType._Message) {
         this.hovering = false;
         return;
       }
@@ -258,14 +256,14 @@ export default class Bell {
         if (this.message.queued.length > 0) {
           return this.message.dequeue().then((msg: any) => {
             this.message.content = msg;
-            this.message.contentType = MesageType._Queued;
+            this.message.contentType = MessageType._Queued;
             resolve();
           });
         } else {
           this.message.content = decodeHtmlEntities(
             this.message.getTipForState(),
           );
-          this.message.contentType = MesageType._Tip;
+          this.message.contentType = MessageType._Tip;
           resolve();
         }
       })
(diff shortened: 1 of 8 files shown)


src/page/bell/Bell.ts
```diff
@@ -347,26 +334,26 @@ export default class Bell {
...
+    OneSignal._emitter.on(BellEvent._StateChanged, (state) => {
       if (!this.launcher.element) {
         // Notify button doesn't exist
         return;
       }
-      if (state.to === Bell.STATES.SUBSCRIBED) {
+      if (state.to === BellState._Subscribed) {
         this.launcher.inactivate();
-      } else if (state.to === Bell.STATES.UNSUBSCRIBED || Bell.STATES.BLOCKED) {
+      } else if (state.to === BellState._Unsubscribed || BellState._Blocked) {
```
sherwinski: Doesn't `BellState._Blocked` always evaluate to truthy? I'm not sure what this condition is trying to check for.

src/page/bell/Bell.ts
```diff
@@ -25,65 +25,40 @@ import Button from './Button';
...
+  BellEvent,
+  BellState,
+  type BellStateType,
+  MESSAGE_TIMEOUT,
+  MesageType,
+} from './constants';
 
 const logoSvg = `<svg class="onesignal-bell-svg" xmlns="http://www.w3.org/2000/svg" width="99.7" height="99.7" viewBox="0 0 99.7 99.7"><circle class="background" cx="49.9" cy="49.9" r="49.9"/><path class="foreground" d="M50.1 66.2H27.7s-2-.2-2-2.1c0-1.9 1.7-2 1.7-2s6.7-3.2 6.7-5.5S33 52.7 33 43.3s6-16.6 13.2-16.6c0 0 1-2.4 3.9-2.4 2.8 0 3.8 2.4 3.8 2.4 7.2 0 13.2 7.2 13.2 16.6s-1 11-1 13.3c0 2.3 6.7 5.5 6.7 5.5s1.7.1 1.7 2c0 1.8-2.1 2.1-2.1 2.1H50.1zm-7.2 2.3h14.5s-1 6.3-7.2 6.3-7.3-6.3-7.3-6.3z"/><ellipse class="stroke" cx="49.9" cy="49.9" rx="37.4" ry="36.9"/></svg>`;
 
 type BellState = 'uninitialized' | 'subscribed' | 'unsubscribed' | 'blocked';
```
sherwinski: This is a repeat of these [constant values](https://github.com/OneSignal/OneSignal-Website-SDK/pull/1397/files#diff-b4c0f5ddfa0fe2cdbf7f2cbd655260f2b5353bc270e0d1c8076e106f4bb230a2R1-R6)

src/page/bell/constants.ts
```diff
@@ -0,0 +1,26 @@
...
+  _StateChanged: 'notifyButtonStateChange',
+  _LauncherClick: 'notifyButtonLauncherClick',
+  _BellClick: 'notifyButtonButtonClick',
+  _SubscribeClick: 'notifyButtonSubscribeClick',
+  _UnsubscribeClick: 'notifyButtonUnsubscribeClick',
+  _Hovering: 'notifyButtonHovering',
+  _Hovered: 'notifyButtonHover',
+} as const;
+
+export const MesageType = {
```
sherwinski:
```suggestion
export const MessageType = {
```

src/page/bell/Launcher.ts
```diff
@@ -79,31 +79,23 @@ export default class Launcher extends AnimatedElement {
...
-    });
+  async inactivate() {
+    await this.bell.message.hide();
+    if (this.bell.badge.content.length > 0) {
+      await this.bell.badge.hide();
+      await Promise.all([super.inactivate(), this.resize('small')]);
+      return this.bell.badge.show();
+    } else {
+      await Promise.all([super.inactivate(), this.resize('small')]);
+      return this;
```
sherwinski: I understand this is a refactor of existing code, but why does it either return `badge.show()` or `this`?

src/page/bell/Bell.ts
```diff
@@ -311,11 +298,11 @@ export default class Bell {
         // immediately after because of the way mobile click events work. Basically only happens if HOVERING and HOVERED
         // fire within a few milliseconds of each other
         this.message
-          .waitUntilShown()
-          .then(() => delay(Message.TIMEOUT))
+          .show()
+          .then(() => delay(MESSAGE_TIMEOUT))
           .then(() => this.message.hide())
           .then(() => {
-            if (this.launcher.wasInactive && this.dialog.hidden) {
```
sherwinski, line 318: Why is this condition changed to use a negative check but [this line](https://github.com/OneSignal/OneSignal-Website-SDK/pull/1397/files#diff-10fc2614389daa8a6a9dfb4fd309dee405b309d7830af97fe26b9f8d0a71521eL302) does the opposite?

src/page/bell/Bell.ts
```diff
@@ -587,20 +574,40 @@ export default class Bell {
...
   setCustomColorsIfSpecified() {
     // Some common vars first
-    const dialogButton = this.dialog.element.querySelector('button.action');
-    const pulseRing = this.button.element.querySelector('.pulse-ring');
+    const dialogElement = this.dialog.element;
+    const dialogButton =
+      dialogElement?.querySelector<HTMLButtonElement>('button.action');
+    const buttonElement = this.button.element;
+    const pulseRing = buttonElement?.querySelector<HTMLElement>('.pulse-ring');
+    const graphic = this.graphic;
```
sherwinski: nit: add optional chaining here as well

src/page/bell/Bell.ts
```diff
@@ -347,26 +334,26 @@ export default class Bell {
...
       if (!this.launcher.element) {
         // Notify button doesn't exist
         return;
       }
-      if (state.to === Bell.STATES.SUBSCRIBED) {
+      if (state.to === BellState._Subscribed) {
         this.launcher.inactivate();
-      } else if (state.to === Bell.STATES.UNSUBSCRIBED || Bell.STATES.BLOCKED) {
+      } else if (state.to === BellState._Unsubscribed || BellState._Blocked) {
         this.launcher.activate();
```
sherwinski, line 358: I don't know if it's intentional but I would double check that other instances of `activate()` need to be awaited

src/page/bell/Bell.ts
fadi-george: Yeah its an exisiting, ill change it to state.to === BellState._Blocked

src/page/bell/Bell.ts
fadi-george: Removed in favor of BellStateValue

src/page/bell/constants.ts
fadi-george: Done

src/page/bell/Launcher.ts
fadi-george: Removed this, helpful for chaining but not really used

src/page/bell/Bell.ts
fadi-george: Done

src/page/bell/Bell.ts
fadi-george: Its in the emitter callback so awaiting wouldnt do anything from what I can tell

src/page/bell/Bell.ts
fadi-george: Line 302 is !this.dialog.hidden, there is a bang in front so it would be equivalent to shown.

sherwinski: I understand that they're equivalent. I'm asking why the choice to use `!this.dialog.shown` instead of `this.dialog.hidden` here.

fadi-george: We dont have `this.dialog.hidden`

PR #1401, opened 2025-10-17
chore: more size reduction

fadi-george:
# Description
## 1 Line Summary
- more size reduction
```
// old
  Size limit: 420 B
  Size:       420 B gzipped
  
  Size limit: 47.56 kB
  Size:       47.56 kB gzipped
  
  Size limit: 13.62 kB
  Size:       13.62 kB gzipped
  
  Size limit: 8.81 kB
  Size:       8.8 kB  gzipped


// new
  Size limit: 420 B
  Size:       420 B gzipped
  
  Size limit: 46.6 kB
  Size:       46.6 kB gzipped
  
  Size limit: 13.44 kB
  Size:       13.44 kB gzipped
  
  Size limit: 8.81 kB
  Size:       8.8 kB  gzipped
``` 

## Details
- breaks up main helper class into exported functions
- breaks up service worker class into functions
- breaks up limit store class into functions
- adds more underscore prefix for more classes
- removes unused path class methods 
- breaks up tags utils class into functions
- adds webhook sender class into functions since sender class was also broke up into a simple send function
- removed self.OneSignalWorker as it is not needed

# Systems Affected
   - [x] WebSDK
   - [ ] Backend
   - [ ] Dashboard

# Validation
### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [ ] No Typescript warnings

Other:

### Checklist

---

## Related Tickets

---

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Website-SDK/1401)

### DIFF
--- a/src/entries/worker.ts
+++ b/src/entries/worker.ts
@@ -3,5 +3,5 @@
  */
 import { run } from '../sw/serviceWorker/ServiceWorker';
 
-// The run() is already called in ServiceWorker.ts, but importing it ensures it's not tree-shaken
-void run;
+// Need to call run() to ensure the service worker is registered but also to ensure the service worker is not tree-shaken
+run();
--- a/src/shared/services/limitStore.ts
+++ b/src/shared/services/limitStore.ts
@@ -1,11 +1,11 @@
 /*
- LimitStore.put('colorado', 'rocky');
+ limitStorePut('colorado', 'rocky');
  ["rocky"]
- LimitStore.put('colorado', 'mountain');
+ limitStorePut('colorado', 'mountain');
  ["rocky", "mountain"]
- LimitStore.put('colorado', 'national');
+ limitStorePut('colorado', 'national');
  ["mountain", "national"]
- LimitStore.put('colorado', 'park');
+ limitStorePut('colorado', 'park');
  ["national", "park"]
  */
 const LIMIT = 2;
@@ -16,7 +16,7 @@ export function limitStorePut<T>(key: string, value: T) {
     store[key] = [null, null];
   }
   store[key].push(value);
-  if (store[key].length == LIMIT + 1) {
+  if (store[key].length === LIMIT + 1) {
     store[key].shift();
   }
   return store[key];
--- a/src/sw/serviceWorker/ServiceWorker.test.ts
+++ b/src/sw/serviceWorker/ServiceWorker.test.ts
@@ -31,7 +31,7 @@ import type {
   UpsertOrDeactivateSessionPayload,
 } from 'src/shared/session/types';
 import { NotificationType } from 'src/shared/subscriptions/constants';
-import { getAppId } from './ServiceWorker';
+import { getAppId, run } from './ServiceWorker';
 
 // Mock webhook notification events
 vi.mock('../webhooks/notifications/webhookNotificationEvent', () => ({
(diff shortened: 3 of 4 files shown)


src/shared/models/Path.ts
```diff
@@ -1,50 +1,30 @@
...
-
-  private readonly path: string;
+  private readonly _path: string;
 
   constructor(path: string) {
     if (!path) throw EmptyArgumentError('path');
-    this.path = path.trim();
-  }
-
-  getQueryString(): string | null {
```
fadi-george, line 19: unused

src/shared/services/limitStore.ts
```diff
@@ -0,0 +1,47 @@
+/*
+ LimitStore.put('colorado', 'rocky');
```
sherwinski: Rename to use new function name

src/shared/services/limitStore.ts
```diff
@@ -0,0 +1,47 @@
...
+ */
+const LIMIT = 2;
+const store: Record<string, Array<unknown>> = {};
+
+export function limitStorePut<T>(key: string, value: T) {
+  if (store[key] === undefined) {
+    store[key] = [null, null];
+  }
+  store[key].push(value);
+  if (store[key].length == LIMIT + 1) {
```
sherwinski: consider using strict equality
```suggestion
  if (store[key].length === LIMIT + 1) {
```

src/entries/worker.ts
```diff
@@ -1,8 +1,7 @@
...
  * New clients will only be including this entry file, which will result in a reduced service worker size.
  */
-import { OneSignalServiceWorker } from '../sw/serviceWorker/ServiceWorker';
+import { run } from '../sw/serviceWorker/ServiceWorker';
 
-// Expose this class to the global scope
-declare const self: ServiceWorkerGlobalScope;
-self.OneSignalWorker = OneSignalServiceWorker;
+// The run() is already called in ServiceWorker.ts, but importing it ensures it's not tree-shaken
+void run;
```
sherwinski: I would need confirmation that this actually works: e.g. does it actually make it to the final bundle? Even then, this approach seems confusing and potentially fragile.

I'm suspicious that it's not needed especially because `run` is already invoked at the [top-level](https://github.com/OneSignal/OneSignal-Website-SDK/pull/1401/files#diff-7855dff06ca3649799bbe3b635fe9784a94f765a208d1612f52ec2cb94815b3dR1115) here.

My suggestion would be to either move that line here

```typescript
import { run } from '../sw/serviceWorker/ServiceWorker';

run();
```

or import the entire file (although the bundle size may take a hit)

```typescript
import '../sw/serviceWorker/ServiceWorker';

// The run() is already called in ServiceWorker.ts, but importing it ensures it's not tree-shaken
void run;
```

src/shared/services/limitStore.ts
fadi-george: Done

src/shared/services/limitStore.ts
fadi-george: Done

src/entries/worker.ts
fadi-george: Both work but for now ive moved run called to worker.ts

sherwinski: Should [this call](https://github.com/OneSignal/OneSignal-Website-SDK/blob/15ff94761e9402d7c46e206c63c5058503b835a9/src/sw/serviceWorker/ServiceWorker.ts#L1187) be removed then?

fadi-george: It is removed

sherwinski (approved):

PR #1405, opened 2025-10-24
chore: more size reduction 

fadi-george:
# Description
## 1 Line Summary
- clean up types usages and add more _ prefix usages
```ts
// old
  Size limit: 420 B
  Size:       420 B gzipped
  
  Size limit: 45.63 kB
  Size:       45.63 kB gzipped
  
  Size limit: 13.44 kB
  Size:       13.44 kB gzipped
  
  Size limit: 8.81 kB
  Size:       8.8 kB  gzipped

// new
  Size limit: 420 B
  Size:       420 B gzipped
  
  Size limit: 45.32 kB
  Size:       45.32 kB gzipped
  
  Size limit: 13.3 kB
  Size:       13.3 kB gzipped
  
  Size limit: 8.81 kB
  Size:       8.8 kB  gzipped
````

## Details
- break up log class into helpers
- add unknwonoperror which is reused when validating operations for executors
- deletes `models/Subscription` in favor of an interface
- simplify getLanguage helper

# Systems Affected
   - [x] WebSDK
   - [ ] Backend
   - [ ] Dashboard

# Validation
### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:
   - [ ] No Typescript warnings

Other:

### Checklist

---

## Related Tickets

---

- - -
This change is [Reviewable](https://reviewable.io/reviews/OneSignal/OneSignal-Website-SDK/1405)

### DIFF
--- a/package.json
+++ b/package.json
@@ -84,7 +84,7 @@
     },
     {
       "path": "./build/releases/OneSignalSDK.page.es6.js",
-      "limit": "45.364 kB",
+      "limit": "45.322 kB",
       "gzip": true
     },
     {
--- a/src/core/executors/IdentityOperationExecutor.test.ts
+++ b/src/core/executors/IdentityOperationExecutor.test.ts
@@ -9,6 +9,7 @@ import {
 } from '__test__/support/helpers/requests';
 import { updateIdentityModel } from '__test__/support/helpers/setup';
 import { ExecutionResult } from 'src/core/types/operation';
+import { UnknownOpError } from 'src/shared/errors/common';
 import type { MockInstance } from 'vitest';
 import { OPERATION_NAME } from '../constants';
 import { RebuildUserService } from '../modelRepo/RebuildUserService';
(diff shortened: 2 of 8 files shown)

### CODE REVIEW

File: src/sw/serviceWorker/ServiceWorker.test.ts [hunk 1]
```diff
@@ -686,7 +686,7 @@ describe('ServiceWorker', () => {
 });
 
 // - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
-const logDebugSpy = vi.spyOn(Log, '_debug');
+const logDebugSpy = vi.spyOn(Log, 'debug');
```
sherwinski, line 689: When I run the test suite locally it fails on this line:

```
TypeError: Cannot redefine property: mock
```

src/shared/errors/common.ts
```diff
@@ -65,3 +65,6 @@ export const WrongTypeArgumentError = (argName: string) =>
 
 export const ReservedArgumentError = (argName: string) =>
   new Error(`"${argName}" is reserved`);
+
+export const UnknownOpError = (op: { _name?: string } | { _name?: string }[]) =>
+  new Error(`Unknown op(s):${op}`);
```
sherwinski, line 70:
```suggestion
  new Error(`Unknown op(s): ${JSON.stringify(op)}`);
```

src/shared/libraries/log.ts
sherwinski, line 1: why rename this from `Log.ts` to `log.ts`?

PR #1455, opened 2026-04-28
fix: [SDK-4341] populate push token from raw subscription for local-id models

sherwinski:
# Description

## 1 Line Summary

Fix subscriptions created with no push token when `login()` is called before `optIn()`.

## Details

When `login()` creates a placeholder `SubscriptionModel` with a `local-` id and empty `token: ''`, the subsequent `optIn()` flow obtains the real push token via `rawPushSubscription`. However, Branch 2 of `updatePushSubscriptionModelWithRawSubscription` was calling `createSubscribedUser(pushModel)` with the stale empty-token model, discarding the `rawPushSubscription` entirely:

```ts
// Branch 2 (before fix): rawPushSubscription is never used
if (IDManager._isLocalId(pushModel.id)) {
  return createSubscribedUser(pushModel); // pushModel.token is '' from login() placeholder
}
```

The fix extracts the model-update logic into `updateModelFromRawSubscription()` and calls it **before** `createSubscribedUser()` in the local-id branch, ensuring the token, `web_auth`, and `web_p256` fields are populated from the actual push subscription data.

# Systems Affected

- [x] WebSDK
- [ ] Backend
- [ ] Dashboard

# Validation

### Info

- Added a new unit test `should populate token from rawPushSubscription when model has local id and empty token (SDK-4341)` that reproduces the exact bug conditions: a model with a `local-` id and `token: ''` (the realistic state produced by `login()`). This test **failed** before the fix and now passes.
- Updated the existing `should create user if push subscription model has a local id` test to expect the `web_auth` and `web_p256` fields that are now correctly populated from the raw subscription.

### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:

- [x] No Typescript warnings

Other:

### Info

N/A — internal logic fix with no UI changes.

### Checklist

---

## Related Tickets

- [SDK-4341](https://linear.app/onesignal/issue/SDK-4341/safari-subscription-created-with-no-push-token-on-custom-web-setup)

---

### DIFF
--- a/package.json
+++ b/package.json
@@ -76,7 +76,7 @@
     },
     {
       "path": "./build/releases/OneSignalSDK.page.es6.js",
-      "limit": "42.28 kB",
+      "limit": "42.3 kB",
       "gzip": true
     },
     {
--- a/src/shared/managers/SubscriptionManager.test.ts
+++ b/src/shared/managers/SubscriptionManager.test.ts
@@ -112,9 +112,53 @@ describe('SubscriptionManager', () => {
             ...BASE_SUB,
             token: rawSubscription.w3cEndpoint?.toString(),
             type: 'ChromePush',
+            web_auth: rawSubscription.w3cAuth,
+            web_p256: rawSubscription.w3cP256dh,
           },
         ],
       });
+    });
+
+    test('should populate token from rawPushSubscription when model has local id and empty token (SDK-4341)', async () => {
+      const rawSubscription = getRawPushSubscription();
+      mockPushManager.getSubscription.mockResolvedValue(
+        Object.assign({}, mockPushSubscription, {
+          endpoint: rawSubscription.w3cEndpoint?.toString(),
+        }),
+      );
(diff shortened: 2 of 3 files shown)


fadi-george:
whats the original issue? is it just this "login() is called before optIn()."?

sherwinski:
> whats the original issue? is it just this "login() is called before optIn()."?

Yes, specifically `login()` has to fail/retry or there has to be some latency in the response. `login()` creates a placeholder model with `token: ''`, and if Branch 2 of `updatePushSubscriptionModelWithRawSubscription` is reached (model still has a `local-` id), the actual token from the push subscription is discarded.

It's difficult to reproduce manually which is why I'm basing the fix off of unit tests.

fadi-george:
Include ticket id in pr title.

fadi-george:
We need more details from the customer.

sherwinski:
Closing for now until we hear back

PR #1458, opened 2026-05-07
chore: update vite-plus, switch to bun, drop deepmerge

fadi-george:
# Description

## 1 Line Summary

Upgrade vite-plus to 0.1.20, switch package manager from pnpm to bun, and replace the `deepmerge` package with a small in-repo helper.

## Details

- Bumps `vite-plus` to 0.1.20 and updates related overrides + dev deps.
- Migrates the lockfile from `pnpm-lock.yaml` to `bun.lock`; `.nvmrc` is replaced by `.node-version`.
- Refreshes `AGENTS.md` to match the current vite-plus guidance.
- Adds a `deepMerge` helper in `__test__/support/helpers/general.ts` and removes the `deepmerge` dependency. The helper covers the two existing callers (`mockOSMinifiedNotificationPayload`, `TestContext.getFakeServerAppConfig`) with the same plain-object recursion + array-concat behavior we were relying on.

# Systems Affected

- [x] WebSDK
- [ ] Backend
- [ ] Dashboard

# Validation

### Info

`vp check` passes (one pre-existing lint warning in `src/shared/config/version.ts` unrelated to this change). `vp test` passes the affected suites that exercise the helper changes (`ServiceWorker.test.ts`, `app.test.ts`, 40 tests).

### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:

- [x] No Typescript warnings

Other:

- [x] Iteration: refrain from using \`elem of array\` syntax. Prefer \`forEach\` or use \`map\`
- [x] Avoid using global OneSignal accessor for \`context\` if possible. Instead, we can pass it to function/constructor so that we don't call \`OneSignal.context\`

### Info

N/A — tooling and test-helper change only.

### Checklist

---

## Related Tickets

Made with [Cursor](https://cursor.com)


.nvmrc
sherwinski, line 1: The deletion of `.nvmrc` is causing CI to fail. vp should probably be able to pick up `.node-version` instead.

package.json
```diff
@@ -48,22 +48,23 @@
...
     "vite-bundle-analyzer": "^1.3.6",
-    "vite-plugin-mkcert": "^1.17.10",
-    "vite-plus": "0.1.13",
-    "vitest": "npm:@voidzero-dev/vite-plus-test@0.1.13"
+    "vite-plugin-mkcert": "^2.0.0",
+    "vite-plus": "0.1.20"
+  },
+  "overrides": {
+    "vite": "npm:@voidzero-dev/vite-plus-core@latest",
+    "vitest": "npm:@voidzero-dev/vite-plus-test@latest"
```
sherwinski: These pinned overrides were included for a specific reason https://github.com/OneSignal/OneSignal-Website-SDK/pull/1449. If the bug no longer exits, then we can just remove the overrides altogether.

sherwinski (changes requested):

package.json
fadi-george: But even new vite + bun setups has these overrides

sherwinski: Do they not need to be deterministic then? If we pin to `@latest` then the version can change.

fadi-george: Ill look into locking the versions

sherwinski (approved):

PR #1461, opened 2026-05-13
refactor: [SDK-4535] update bell dialog logic

fadi-george:
# Description

## 1 Line Summary

Refactor the bell notification dialog to render dashboard-controlled text via safe DOM APIs, matching the pattern already used by sibling components.

## Details

[SDK-4535](https://linear.app/onesignal/issue/SDK-4535/website-sdk-stored-xss-via-unescaped-server-config-text-in-bell-dialog) — see Linear for full context.

### Background

`Dialog.ts` was building the dialog body by interpolating values from `serverConfig.config.staticPrompts.bell.dialog.*` into HTML template literals, then inserting via `addDomElement` (`insertAdjacentHTML`). Sibling bell components (`Bell.ts`, `Message.ts`, `Badge.ts`) already follow a safer convention: insert static scaffolding HTML and set dynamic values via `textContent` / `setAttribute`. This PR brings `Dialog.ts` in line.

### Change

1. `addDomElement` now inserts only static scaffolding (empty `<h1></h1>`, empty `<p></p>`, empty `` — no template-literal interpolation of dynamic values).
2. Dynamic text and attributes are populated via `textContent` / `setAttribute` against the just-inserted nodes.
3. Two private helpers `_renderSubscription` and `_renderBlocked` replace the prior `_buildSubscriptionContent` / `_buildBlockedContent` HTML-string builders.

Behavior is unchanged for valid inputs; the render is structurally identical.

### Drive-by

Fixed a type-honesty issue in `src/shared/config/version.ts`. `convertConfigToVersionTwo`'s parameter was typed `SlidedownOptionsVersion1 & SlidedownOptions`, but this function exists to **add** `prompts` to a V1 config that doesn't have one. Changed to `SlidedownOptionsVersion1 & Partial<SlidedownOptions>` so the existing `= []` default is no longer flagged as a `no-useless-default-assignment` lint warning. Tests confirm the default was load-bearing.

# Systems Affected

- [x] WebSDK
- [ ] Backend
- [ ] Dashboard

# Validation

### Info

Added 5 regression tests in `Dialog.test.ts` covering each text/attribute sink in the dialog. Each test sets the corresponding option to an HTML-like string and asserts the value renders as literal text (`textContent` round-trips the raw input) rather than parsed markup.

### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:

- [x] No Typescript warnings

Other:

### Info

UI is unchanged. Dialog rendering identical for legitimate input.

### Checklist

---

## Related Tickets

- [SDK-4535](https://linear.app/onesignal/issue/SDK-4535/website-sdk-stored-xss-via-unescaped-server-config-text-in-bell-dialog)

---

### DIFF
--- a/src/page/bell/Dialog.ts
+++ b/src/page/bell/Dialog.ts
@@ -119,10 +119,7 @@ export default class Dialog {
     if (isMobileOrTablet && browserName === Browser._Chrome) {
       instructionsHtml = `<ol><li>Access <strong>Settings</strong> by tapping the three menu dots <strong>⋮</strong></li><li>Click <strong>Site settings</strong> under Advanced.</li><li>Click <strong>Notifications</strong>.</li><li>Find and click this entry for this website.</li><li>Click <strong>Notifications</strong> and set it to <strong>Allow</strong>.</li></ol>`;
     } else {
-      const imagePath =
-        browserName === Browser._Chrome && isMobileOrTablet
-          ? undefined
-          : UNBLOCK_IMAGES[browserName];
+      const imagePath = UNBLOCK_IMAGES[browserName];
       if (imagePath) {
         const imageUrl = STATIC_RESOURCES_URL + imagePath;
         instructionsHtml = `<a href="${imageUrl}" target="_blank"><img src="${imageUrl}"></a>`;
--- a/src/page/bell/Dialog.ts
+++ b/src/page/bell/Dialog.ts
@@ -63,15 +63,18 @@ export default class Dialog {
       (state === BellState._Subscribed && isEnabled) ||
       (state === BellState._Unsubscribed && !isEnabled);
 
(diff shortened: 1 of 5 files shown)


fadi-george:
@claude review

fadi-george:
@claude review

src/page/bell/Dialog.ts
sherwinski: The `browserName === Browser._Chrome && isMobileOrTablet` already gets handled in L124 so we can just simplify this:


```suggestion
const imagePath = UNBLOCK_IMAGES[browserName];
```

sherwinski (approved):

src/page/bell/Dialog.ts
fadi-george: Thanks

PR #1462, opened 2026-05-13
chore: [SDK-4534] harden release workflow input handling

fadi-george:
# Description

## 1 Line Summary

Harden `create-release-pr.yml` by routing step outputs through `env:` and adding a least-privilege `permissions:` block.

## Details

Internal hardening pass on the release workflow. See [SDK-4534](https://linear.app/onesignal/issue/SDK-4534) for context.

### Changes

- Added a top-level `permissions:` block scoped to `contents: write` + `pull-requests: write`.
- Replaced inline `${{ steps.*.outputs.* }}` interpolation inside `run:` and `actions/github-script` `script:` bodies with per-step `env:` mappings, referenced as `$VAR` (bash) or `process.env.VAR` (JS).
- Kept the release-notes heredoc as `<< 'EOF'` and added a comment so the quoting is preserved.

# Systems Affected

- [x] WebSDK
- [ ] Backend
- [ ] Dashboard

# Validation

### Info

No automated coverage for workflow files. Verified by re-grep: every remaining `${{ steps.*.outputs.* }}` reference now lives in an `env:` key or inside the quoted heredoc. Bash logic in `Calculate new version` and JS in `Generate release notes` are functionally unchanged.

### Checklist

### Info

N/A, workflow change.

### Checklist

---

## Related Tickets

- [SDK-4534](https://linear.app/onesignal/issue/SDK-4534)

### DIFF
--- a/.github/workflows/create-release-pr.yml
+++ b/.github/workflows/create-release-pr.yml
@@ -3,6 +3,10 @@ name: Create Release PR
 on:
   workflow_dispatch:
 
+permissions:
+  contents: write
+  pull-requests: write
+
 jobs:
   bump-version:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/create-release-pr.yml
```diff
@@ -132,10 +144,11 @@ jobs:
...
 
       - name: Create release PR
+        env:
+          NEW_VERSION: ${{ steps.new_version.outputs.version }}
         run: |
-          NEW_VERSION="${{ steps.new_version.outputs.version }}"
-
-          # Write release notes to file to avoid shell interpretation
+          # Quoted heredoc ('EOF') disables shell expansion, so the interpolated
+          # release notes cannot be re-parsed as shell. Keep the quotes.
```
sherwinski: nit: is this necessary?

sherwinski (approved):

.github/workflows/create-release-pr.yml
fadi-george: shortened comment

PR #1463, opened 2026-05-15
fix: [SDK-4510] encode URL path segments and tighten alias validation

fadi-george:
# Description

## 1 Line Summary

Encode API request path segments at the network boundary and tighten alias label/id validation to harden URL construction.

## Details

Internal hardening for [SDK-4510](https://linear.app/onesignal/issue/SDK-4510).

- `OneSignalApiBase.call` now percent-encodes each segment of the request action via `encodeRFC3986URIComponent` before composing the URL, instead of trusting per-callsite sanitization. This is a single chokepoint that covers every endpoint defined in `src/core/requests/api.ts`.
- Removed the redundant inline `sanitizedAlias` block from `updateUserByAlias` since the encoding is centralized.
- `validateAliasValue` in `src/onesignal/User.ts` adds a 128-character limit (matching the documented public alias length) plus a small blocklist of characters that have no place in an alias label or id. Both `addAlias`/`addAliases` and `removeAlias`/`removeAliases` now flow through it for `label` and `id`.
- Bumped `size-limit` thresholds modestly (page.es6.js 42.32 -> 42.4 kB, sw.js 12.35 -> 12.4 kB) to fit the new code. Final sizes are 42.37 kB and 12.35 kB respectively.

# Systems Affected

- [x] WebSDK
- [ ] Backend
- [ ] Dashboard

# Validation

### Info

- New `__test__/unit/http/urlEncoding.test.ts` verifies the centralized encoding behavior at the `OneSignalApiBase.call` boundary across spaces, reserved chars, sub-delims, control chars, percent signs, and pass-through of URL-safe segments.
- Extended `src/onesignal/UserNamespace.test.ts` with cases covering the new label/id rules: length cap, traversal substring, dangerous characters, control characters, and a 128-character boundary acceptance test.

### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:

- [x] No Typescript warnings

Other:

- [x] Iteration: refrain from using \`elem of array\` syntax. Prefer \`forEach\` or use \`map\`
- [x] Avoid using global OneSignal accessor for \`context\` if possible. Instead, we can pass it to function/constructor so that we don't call \`OneSignal.context\`

### Info

Not applicable; no UI changes.

### Checklist

---

## Related Tickets

[SDK-4510](https://linear.app/onesignal/issue/SDK-4510)

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src/onesignal/User.ts
+++ b/src/onesignal/User.ts
@@ -69,7 +69,7 @@ export default class User {
     logMethodCall('addAlias', { label, id });
     if (isConsentRequiredButNotGiven()) return;
 
-    validateLabel(label, 'label');
+    validateAliasLabel(label, 'label');
     validateAliasValue(id, 'id');
 
     this.addAliases({ [label]: id });
@@ -83,7 +83,7 @@ export default class User {
 
     for (const label of Object.keys(aliases)) {
       validateAliasValue(aliases[label], `key: ${label}`);
-      validateLabel(label, `key: ${label}`);
+      validateAliasLabel(label, `key: ${label}`);
     }
 
     this._updateIdentityModel(aliases);
@@ -315,7 +315,7 @@ function validateStringLabel(label: string, labelName: string): void {
 function validateArray(
   array: string[],
   arrayName: string,
-  validateItem: (item: string, itemName: string) => void = validateLabel,
+  validateItem: (item: string, itemName: string) => void = validateAliasLabel,
 ): void {
   if (!Array.isArray(array)) throw WrongTypeArgumentError(arrayName);
 
@@ -343,7 +343,7 @@ function validateAliasValue(value: string, valueName: string): void {
   }
 }
 
-function validateLabel(label: string, labelName: string): void {
+function validateAliasLabel(label: string, labelName: string): void {
   validateAliasValue(label, labelName);
 
   if (label === 'external_id' || label === 'onesignal_id') {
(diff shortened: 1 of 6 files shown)


src/onesignal/User.ts
```diff
@@ -328,8 +328,19 @@ function validateObject(object: unknown, objectName: string): void {
...
+    throw MalformedArgumentError(valueName);
+  }
+}
+
 function validateLabel(label: string, labelName: string): void {
-  validateStringLabel(label, labelName);
+  validateAliasValue(label, labelName);
 
   if (label === 'external_id' || label === 'onesignal_id') {
     throw ReservedArgumentError(label);
```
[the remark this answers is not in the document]
fadi-george: addressed.

fadi-george:
@claude review

src/onesignal/User.ts
```diff
@@ -328,8 +332,19 @@ function validateObject(object: unknown, objectName: string): void {
...
+
+function validateAliasValue(value: string, valueName: string): void {
+  validateStringLabel(value, valueName);
+
+  if (value.length > 128 || INVALID_ALIAS_PATTERN.test(value)) {
+    throw MalformedArgumentError(valueName);
+  }
+}
+
 function validateLabel(label: string, labelName: string): void {
```
sherwinski: nit: since `validateLabel` is alias-specific, it would be more clear to have the function name reflect that

sherwinski (approved):

src/onesignal/User.ts
fadi-george: Renamed to `validateAliasLabel` in 140a635.

PR #1468, opened 2026-05-28
fix: [SDK-4336] guard IndexedDB Options writes from iOS Safari PWA wedge

### LINKED ISSUE #1452: [Bug]: OneSignal.init() hangs on subsequent page navigations in iOS Safari PWA after push subscription is registered
### What happened?

OneSignal.init() hangs for approximately 30 minutes when navigating to a page that calls init() in an iOS Safari PWA (added to home screen), after push notification permission has been granted and a subscription has been registered.

The issue does **not** occur on the first visit or before push registration. It only manifests after a push subscription exists.
This issue does not occur on macOS Chrome.

### What browsers are you seeing the problem on?

Safari PWA — added to home screen

### What operating system are you running?

iOS 18.3.2 (iPhone) / iOS 26.3.1 (iPad)

### Steps to reproduce?

1. Add the site as a PWA (Add to Home Screen) and open Page A.
    - init() completes successfully. ("!!!! OneSignal initialized" is logged.)
2. Tap the link to navigate to Page B.
3. Tap the link to navigate to Page A.
    - init() completes successfully.
4. Tap the "Register" button to subscribe to push notifications.
    - Push subscription is created; test notifications are received.
5. Tap the link to navigate to Page B.
6. Tap the link to navigate to Page A.
    - init() hangs for 30 minutes. ("!!!! OneSignal initialized" is not logged until 30 minutes later.)

### Minimal Reproduction Code

**Page A** (/A):

```html
<!DOCTYPE html>
<html>
<head>
  <title>Test Page A</title>
  <script src="<https://cdn.onesignal.com/sdks/web/v16/OneSignalSDK.page.js>" defer></script>
  <script>
    console.log("!!!! OneSignal initialize");
    window.OneSignalDeferred = window.OneSignalDeferred || [];
    OneSignalDeferred.push(async function(OneSignal) {
      await OneSignal.init({
        appId: "YOUR_APP_ID"
      });
      console.log("!!!! OneSignal initialized");
    });
  </script>
</head>
<body>
  <h1>Test Page A</h1>
  <a href='/B'>Page B</a>
  <button onclick="requestNotificationPermission()">Register</button>
  <script>
    async function requestNotificationPermission() {
      window.OneSignalDeferred = window.OneSignalDeferred || [];
      OneSignalDeferred.push(async function(OneSignal) {
        await OneSignal.Notifications.requestPermission();
      });
    }
  </script>
</body>
</html>
```

**Page B** (/B): A simple page with a link back to Page A.

### What did you expect to happen?

OneSignal.init() should resolve every time the page is loaded, regardless of whether a push subscription already exists.

### Actual Behavior

When a push subscription already exists, OneSignal.init() hangs for approximately 30 minutes on subsequent navigations. The initialization stalls after [WM] Page listening and does not proceed to internalInit() / sessionInit() / establishSWChannel.

After 30 minutes, the SDK re-triggers the initialization flow from [WM] Page listening and eventually completes successfully (see "Recovery" log below).

#### Successful initialization (before push registration)

```
[Log] !!!! OneSignal initialize (ja, line 9)
[Warning] PushSubscriptionNamespace: skipping initialization. One or more required params are falsy: initialize: false, subscription: undefined (OneSignalSDK.page.es6.js, line 1)
[Info] Web SDK loaded (version 160603, Browser environment). (OneSignalSDK.page.es6.js, line 1)
[Debug] Current Page URL: <https://xxxxx> (OneSignalSDK.page.es6.js, line 1)
[Debug] init() (OneSignalSDK.page.es6.js, line 1)
[Debug] Browser Environment: safari 18.3 (OneSignalSDK.page.es6.js, line 1)
[Debug] Final web app config: – {appId: "xxxxx", hasUnsupportedSubdomain: false, siteName: "xxxxx", …} (OneSignalSDK.page.es6.js, line 1)
{appId: "xxxxx", hasUnsupportedSubdomain: false, siteName: "xxxxx", origin: "<https://xxxxx>", restrictedOriginEnabled: false, …}Object
[Debug] CoreModule.init() (OneSignalSDK.page.es6.js, line 1)
[Warning] PushSubscriptionNamespace: skipping initialization. One or more required params are falsy: initialize: false, subscription: undefined (OneSignalSDK.page.es6.js, line 1)
[Debug] CoreModuleDirector.getIdentityModel() (OneSignalSDK.page.es6.js, line 1)
[Debug] CoreModuleDirector.getPropertiesModel() (OneSignalSDK.page.es6.js, line 1)
[Debug] CoreModuleDirector.getPushSubscriptionModel() (OneSignalSDK.page.es6.js, line 1)
[Debug] CoreModuleDirector.getPushSubscriptionModelByLastKnownToken() (OneSignalSDK.page.es6.js, line 1)
[Debug] CoreModuleDirector.getSubscriptionOfTypeWithToken({"type":"Push","token":""}) (OneSignalSDK.page.es6.js, line 1)
[Debug] CoreModuleDirector.getAllPushSubscriptionModels() (OneSignalSDK.page.es6.js, line 1)
[Debug] CoreModuleDirector.getPushSubscriptionModelByCurrentToken() (OneSignalSDK.page.es6.js, line 1)
[Debug] [WM] Page listening (<https://xxxxxx>) (OneSignalSDK.page.es6.js, line 1)
[Info] Set pageTitle to be 'xxxxx'. (OneSignalSDK.page.es6.js, line 1)
[Debug] internalInit() (OneSignalSDK.page.es6.js, line 1)
[Debug] shouldInstallWorker state: – "None" (OneSignalSDK.page.es6.js, line 1)
[Debug] Page views: tab=2, total=3 (OneSignalSDK.page.es6.js, line 1)
[Debug] sessionInit() (OneSignalSDK.page.es6.js, line 1)
[Debug] establishSWChannel (OneSignalSDK.page.es6.js, line 1)
[Debug] CoreModuleDirector.getPushSubscriptionModel() (OneSignalSDK.page.es6.js, line 1)
[Debug] CoreModuleDirector.getPushSubscriptionModelByLastKnownToken() (OneSignalSDK.page.es6.js, line 1)
[Debug] CoreModuleDirector.getSubscriptionOfTypeWithToken({"type":"Push","token":""}) (OneSignalSDK.page.es6.js, line 1)
[Debug] CoreModuleDirector.getAllPushSubscriptionModels() (OneSignalSDK.page.es6.js, line 1)
[Debug] CoreModuleDirector.getPushSubscriptionModelByCurrentToken() (OneSignalSDK.page.es6.js, line 1)
[Debug] isOptedOut() (OneSignalSDK.page.es6.js, line 1)
[Debug] isOptedOut() (OneSignalSDK.page.es6.js, line 1)
[Info] handleAutoResubscribe – {autoResubscribe: true, isOptedOut: false} (OneSignalSDK.page.es6.js, line 1)
[Debug] CoreModuleDirector.getPushSubscriptionModel() (OneSignalSDK.page.es6.js, line 1)
[Debug] CoreModuleDirector.getPushSubscriptionModelByLastKnownToken() (OneSignalSDK.page.es6.js, line 1)
[Debug] CoreModuleDirector.getSubscriptionOfTypeWithToken({"type":"Push","token":""}) (OneSignalSDK.page.es6.js, line 1)
[Debug] CoreModuleDirector.getAllPushSubscriptionModels() (OneSignalSDK.page.es6.js, line 1)
[Debug] CoreModuleDirector.getPushSubscriptionModelByCurrentToken() (OneSignalSDK.page.es6.js, line 1)
[Debug] (Browser) » initializeInternal (OneSignalSDK.page.es6.js, line 1)
[Debug] Checking sub expiration (OneSignalSDK.page.es6.js, line 1)
[Log] !!!! OneSignal initialized (ja, line 15)
[Debug] Sub not expired (OneSignalSDK.page.es6.js, line 1)
[Debug] (Browser) » initialize (OneSignalSDK.page.es6.js, line 1)
```

#### Failed initialization (after push registration — hangs)

```
[Log] !!!! OneSignal initialize (ja, line 9)
[Warning] PushSubscriptionNamespace: skipping initialization. One or more required params are falsy: initialize: false, subscription: undefined (OneSignalSDK.page.es6.js, line 1)
[Info] Web SDK loaded (version 160603, Browser environment). (OneSignalSDK.page.es6.js, line 1)
[Debug] Current Page URL: <https://xxxxx> (OneSignalSDK.page.es6.js, line 1)
[Debug] init() (OneSignalSDK.page.es6.js, line 1)
[Debug] Browser Environment: safari 18.3 (OneSignalSDK.page.es6.js, line 1)
[Debug] Final web app config: – {appId: "xxxxx", hasUnsupportedSubdomain: false, siteName: "xxxxx", …} (OneSignalSDK.page.es6.js, line 1)
{appId: "xxxxx", hasUnsupportedSubdomain: false, siteName: "xxxxx", origin: "<https://xxxxx>", restrictedOriginEnabled: false, …}Object
[Debug] CoreModule.init() (OneSignalSDK.page.es6.js, line 1)
[Warning] PushSubscriptionNamespace: skipping initialization. One or more required params are falsy: initialize: false, subscription: undefined (OneSignalSDK.page.es6.js, line 1)
[Debug] CoreModuleDirector.getIdentityModel() (OneSignalSDK.page.es6.js, line 1)
[Debug] CoreModuleDirector.getPropertiesModel() (OneSignalSDK.page.es6.js, line 1)
[Debug] CoreModuleDirector.getPushSubscriptionModel() (OneSignalSDK.page.es6.js, line 1)
[Debug] CoreModuleDirector.getPushSubscriptionModelByLastKnownToken() (OneSignalSDK.page.es6.js, line 1)
[Debug] CoreModuleDirector.getSubscriptionOfTypeWithToken({"type":"Push","token":"<https://web.push.apple.com/xxxxx>"}) (OneSignalSDK.page.es6.js, line 1)
[Debug] CoreModuleDirector.getAllPushSubscriptionModels() (OneSignalSDK.page.es6.js, line 1)
[Debug] CoreModuleDirector.getPushSubscriptionModelByCurrentToken() (OneSignalSDK.page.es6.js, line 1)
[Debug] [WM] Page listening (<https://xxxxx>) (OneSignalSDK.page.es6.js, line 1)
[Debug] CoreModuleDirector.getSubscriptionOfTypeWithToken({"type":"Push","token":"<https://web.push.apple.com/xxxxx>"}) (OneSignalSDK.page.es6.js, line 1)
[Debug] CoreModuleDirector.getAllPushSubscriptionModels() (OneSignalSDK.page.es6.js, line 1)
```

#### Recovery after ~30 minutes idle (initialization eventually resumes and completes)

```
[Debug] CoreModuleDirector.getPushSubscriptionModelByCurrentToken() (OneSignalSDK.page.es6.js, line 1)
[Debug] [WM] Page listening (<https://xxxxx>) (OneSignalSDK.page.es6.js, line 1)
[Debug] CoreModuleDirector.getSubscriptionOfTypeWithToken({"type":"Push","token":"<https://web.push.apple.com/xxxxx>"}) (OneSignalSDK.page.es6.js, line 1)
[Debug] CoreModuleDirector.getAllPushSubscriptionModels() (OneSignalSDK.page.es6.js, line 1)
[Info] Set pageTitle to be 'XXXXX'. (OneSignalSDK.page.es6.js, line 1)
[Debug] internalInit() (OneSignalSDK.page.es6.js, line 1)
[Debug] shouldInstallWorker state: – "OneSignal Worker" (OneSignalSDK.page.es6.js, line 1)
[Info] Checking SW version... (OneSignalSDK.page.es6.js, line 1)
[Debug] [WM] Page->SW unicast 'GetWorkerVersion' (OneSignalSDK.page.es6.js, line 1)
[Debug] [WM] Page->SW direct 'GetWorkerVersion' (OneSignalSDK.page.es6.js, line 1)
[Debug] [WM] Page received: – {command: "GetWorkerVersion", payload: "160603"} (OneSignalSDK.page.es6.js, line 1)
[Info] SW version current: 160603 (OneSignalSDK.page.es6.js, line 1)
[Debug] Page views: tab=4, total=5 (OneSignalSDK.page.es6.js, line 1)
[Debug] sessionInit() (OneSignalSDK.page.es6.js, line 1)
[Debug] establishSWChannel (OneSignalSDK.page.es6.js, line 1)
[Debug] CoreModuleDirector.getPushSubscriptionModel() (OneSignalSDK.page.es6.js, line 1)
[Debug] CoreModuleDirector.getPushSubscriptionModelByLastKnownToken() (OneSignalSDK.page.es6.js, line 1)
[Debug] CoreModuleDirector.getSubscriptionOfTypeWithToken({"type":"Push","token":"<https://web.push.apple.com/xxxxx>"}) (OneSignalSDK.page.es6.js, line 1)
[Debug] CoreModuleDirector.getAllPushSubscriptionModels() (OneSignalSDK.page.es6.js, line 1)
[Debug] CoreModuleDirector.getPushSubscriptionModelByCurrentToken() (OneSignalSDK.page.es6.js, line 1)
[Debug] CoreModuleDirector.getSubscriptionOfTypeWithToken({"type":"Push","token":"<https://web.push.apple.com/xxxxx>"}) (OneSignalSDK.page.es6.js, line 1)
[Debug] CoreModuleDirector.getAllPushSubscriptionModels() (OneSignalSDK.page.es6.js, line 1)
[Debug] isOptedOut() (OneSignalSDK.page.es6.js, line 1)
[Debug] isOptedOut() (OneSignalSDK.page.es6.js, line 1)
[Info] handleAutoResubscribe – {autoResubscribe: true, isOptedOut: false} (OneSignalSDK.page.es6.js, line 1)
[Debug] shouldInstallWorker state: – "OneSignal Worker" (OneSignalSDK.page.es6.js, line 1)
[Info] Checking SW version... (OneSignalSDK.page.es6.js, line 1)
[Debug] [WM] Page->SW unicast 'GetWorkerVersion' (OneSignalSDK.page.es6.js, line 1)
[Debug] [WM] Page->SW direct 'GetWorkerVersion' (OneSignalSDK.page.es6.js, line 1)
[Debug] [WM] Page received: – {command: "GetWorkerVersion", payload: "160603"} (OneSignalSDK.page.es6.js, line 1)
[Info] SW version current: 160603 (OneSignalSDK.page.es6.js, line 1)
[Debug] SW ready, continuing subscription (OneSignalSDK.page.es6.js, line 1)
[Debug] Existing push sub options present (OneSignalSDK.page.es6.js, line 1)
[Debug] Subscribing with options: – {userVisibleOnly: true, applicationServerKey: ArrayBuffer} (OneSignalSDK.page.es6.js, line 1)
{userVisibleOnly: true, applicationServerKey: ArrayBuffer}Object
[Debug] CoreModuleDirector.getPushSubscriptionModel() (OneSignalSDK.page.es6.js, line 1)
[Debug] CoreModuleDirector.getPushSubscriptionModelByLastKnownToken() (OneSignalSDK.page.es6.js, line 1)
[Debug] CoreModuleDirector.getSubscriptionOfTypeWithToken({"type":"Push","token":"<https://web.push.apple.com/xxxxx>"}) (OneSignalSDK.page.es6.js, line 1)
[Debug] CoreModuleDirector.getAllPushSubscriptionModels() (OneSignalSDK.page.es6.js, line 1)
[Debug] CoreModuleDirector.getPushSubscriptionModelByCurrentToken() (OneSignalSDK.page.es6.js, line 1)
[Debug] CoreModuleDirector.getSubscriptionOfTypeWithToken({"type":"Push","token":"<https://web.push.apple.com/xxxxx>"}) (OneSignalSDK.page.es6.js, line 1)
[Debug] CoreModuleDirector.getAllPushSubscriptionModels() (OneSignalSDK.page.es6.js, line 1)
[Debug] CoreModuleDirector.getIdentityModel() (OneSignalSDK.page.es6.js, line 1)
[Debug] (Browser) » register (OneSignalSDK.page.es6.js, line 1)
[Debug] checkAndTriggerSubscriptionChanged() (OneSignalSDK.page.es6.js, line 1)
[Debug] CoreModuleDirector.getPushSubscriptionModel() (OneSignalSDK.page.es6.js, line 1)
[Debug] CoreModuleDirector.getPushSubscriptionModelByLastKnownToken() (OneSignalSDK.page.es6.js, line 1)
[Debug] CoreModuleDirector.getSubscriptionOfTypeWithToken({"type":"Push","token":"<https://web.push.apple.com/xxxxx>"}) (OneSignalSDK.page.es6.js, line 1)
[Debug] CoreModuleDirector.getAllPushSubscriptionModels() (OneSignalSDK.page.es6.js, line 1)
[Debug] CoreModuleDirector.getPushSubscriptionModelByCurrentToken() (OneSignalSDK.page.es6.js, line 1)
[Debug] CoreModuleDirector.getSubscriptionOfTypeWithToken({"type":"Push","token":"<https://web.push.apple.com/xxxxx>"}) (OneSignalSDK.page.es6.js, line 1)
[Debug] CoreModuleDirector.getAllPushSubscriptionModels() (OneSignalSDK.page.es6.js, line 1)
[Debug] CoreModuleDirector.getPushSubscriptionModel() (OneSignalSDK.page.es6.js, line 1)
[Debug] CoreModuleDirector.getPushSubscriptionModelByLastKnownToken() (OneSignalSDK.page.es6.js, line 1)
[Debug] CoreModuleDirector.getSubscriptionOfTypeWithToken({"type":"Push","token":"<https://web.push.apple.com/xxxxx>"}) (OneSignalSDK.page.es6.js, line 1)
[Debug] CoreModuleDirector.getAllPushSubscriptionModels() (OneSignalSDK.page.es6.js, line 1)
[Debug] CoreModuleDirector.getPushSubscriptionModelByCurrentToken() (OneSignalSDK.page.es6.js, line 1)
[Debug] CoreModuleDirector.getSubscriptionOfTypeWithToken({"type":"Push","token":"<https://web.push.apple.com/xxxxx>"}) (OneSignalSDK.page.es6.js, line 1)
[Debug] CoreModuleDirector.getAllPushSubscriptionModels() (OneSignalSDK.page.es6.js, line 1)
[Debug] CoreModuleDirector.getPushSubscriptionModel() (OneSignalSDK.page.es6.js, line 1)
[Debug] CoreModuleDirector.getPushSubscriptionModelByLastKnownToken() (OneSignalSDK.page.es6.js, line 1)
[Debug] CoreModuleDirector.getSubscriptionOfTypeWithToken({"type":"Push","token":"<https://web.push.apple.com/xxxxx>"}) (OneSignalSDK.page.es6.js, line 1)
[Debug] CoreModuleDirector.getAllPushSubscriptionModels() (OneSignalSDK.page.es6.js, line 1)
[Debug] CoreModuleDirector.getPushSubscriptionModelByCurrentToken() (OneSignalSDK.page.es6.js, line 1)
[Debug] CoreModuleDirector.getSubscriptionOfTypeWithToken({"type":"Push","token":"<https://web.push.apple.com/xxxxx>"}) (OneSignalSDK.page.es6.js, line 1)
[Debug] CoreModuleDirector.getAllPushSubscriptionModels() (OneSignalSDK.page.es6.js, line 1)
[Debug] CoreModuleDirector.getPushSubscriptionModel() (OneSignalSDK.page.es6.js, line 1)
[Debug] CoreModuleDirector.getPushSubscriptionModelByLastKnownToken() (OneSignalSDK.page.es6.js, line 1)
[Debug] CoreModuleDirector.getSubscriptionOfTypeWithToken({"type":"Push","token":"<https://web.push.apple.com/xxxxx>"}) (OneSignalSDK.page.es6.js, line 1)
[Debug] CoreModuleDirector.getAllPushSubscriptionModels() (OneSignalSDK.page.es6.js, line 1)
[Debug] CoreModuleDirector.getPushSubscriptionModelByCurrentToken() (OneSignalSDK.page.es6.js, line 1)
[Debug] CoreModuleDirector.getSubscriptionOfTypeWithToken({"type":"Push","token":"<https://web.push.apple.com/xxxxx>"}) (OneSignalSDK.page.es6.js, line 1)
[Debug] CoreModuleDirector.getAllPushSubscriptionModels() (OneSignalSDK.page.es6.js, line 1)
[Debug] (Browser) » initializeInternal (OneSignalSDK.page.es6.js, line 1)
[Debug] Checking sub expiration (OneSignalSDK.page.es6.js, line 1)
[Log] !!!! OneSignal initialized (ja, line 15)
[Debug] Sub not expired (OneSignalSDK.page.es6.js, line 1)
[Debug] setupOnFocusAndOnBlurForSession (OneSignalSDK.page.es6.js, line 1)
[Debug] (Browser) » initialize (OneSignalSDK.page.es6.js, line 1)
```

### Analysis

Comparing the two logs, the key differences are:

1. Successful init: getSubscriptionOfTypeWithToken is called with an empty token (""), and the flow continues to Set pageTitle → internalInit() → sessionInit() → establishSWChannel → completion.
2. Failed init: getSubscriptionOfTypeWithToken is called with a real Apple push token (https://web.push.apple.com/...), and the flow stalls after [WM] Page listening. It does not proceed to internalInit() for ~30 minutes.

This suggests the SDK may be waiting on a Service Worker response or some asynchronous operation related to the existing push subscription that does not resolve promptly in the iOS Safari PWA environment on subsequent navigations.

sherwinski:
# Description

## 1 Line Summary

Stops `OneSignal.init()` from hanging for ~30 minutes on iOS 26 Safari PWAs by capping `Options`-store readwrite operations with a 1.5s hard timeout and tripping a page-scoped circuit breaker once the wedge is detected.

Note: A bug report ([315804](https://bugs.webkit.org/show_bug.cgi?id=315804)) was filed in Webkit upstream to investigate this.

### Root cause

On iOS 26 Safari running as a Home-Screen PWA (`display: standalone`), the navigation back into the app *after a successful push subscription* leaves the `ONE_SIGNAL_SDK_DB` IndexedDB in a poisoned state where every `readwrite` request on the `Options` object store stalls indefinitely. The `IDBRequest` is created and goes to `pending`, but **no event ever fires** — no `success`, no `error`, no transaction `complete` / `abort`, no `IDBDatabase.onclose`. WebKit's internal transaction watchdog only forcibly aborts the wedged transaction after roughly 30 minutes; until then `await db.put('Options', ...)` never settles.

`OneSignal.init()` always writes to `Options` very early — `initSaveState` writes `pageTitle`, `saveInitOptions` writes 7+ entries (webhook URLs, persistNotification, click-handler config, `lastPushToken`, `isPushEnabled`, etc.). Without this guard, **the very first of those writes blocks init forever** and the support team observes \"init hanging for 30 minutes, then eventually recovers\" — exactly the watchdog timer.

What we ruled out and how:

- **NetworkProcess crash family** (WebKit bugs [273827](https://bugs.webkit.org/show_bug.cgi?id=273827) / [277615](https://bugs.webkit.org/show_bug.cgi?id=277615) / [309386](https://bugs.webkit.org/show_bug.cgi?id=309386)). Those throw `UnknownError: Connection to Indexed Database server lost` and fire `IDBDatabase.onclose`. We never see either.
- **Process suspension** ([202705](https://bugs.webkit.org/show_bug.cgi?id=202705)). That throws `TransactionInactiveError` synchronously. We throw nothing.
- **PWA total freeze** ([211018](https://bugs.webkit.org/show_bug.cgi?id=211018)). All other JS keeps running — timers fire, fetch works, the SW heartbeat keeps ticking — only this one `IDBRequest` is dead.
- **Connection-level wedge.** Reads (`get`, `getAll`) on the same DB connection still work. `readwrite` on other stores still works. Closing and re-opening the database returns a fresh `IDBDatabase` whose first `readwrite` on `Options` wedges identically.
- **Origin-wide IDB wedge.** A diagnostic probe that opened a *different* IndexedDB name at the same origin and issued a `readwrite put` completed in 11 ms while the real DB was hung. The wedge is per-database, not origin-wide.

Filed upstream as WebKit bug [315804](https://bugs.webkit.org/show_bug.cgi?id=315804) ("A readwrite IDBTransaction never fires oncomplete, onerror, or onabort after the user subscribes to web push in an installed PWA") with a link back to this PR for steps on how to reproduce. The source comment in `client.ts` also references the bug so the workaround can be removed once WebKit ships a fix.

### Fix

Three commits, each independently revertable:

1. **`81415fdd` — fix: fail-fast Options writes.** Wrap `db.put('Options', ...)` and `db.delete('Options', ...)` with a 1500ms hard timeout. On timeout, log a `[SDK-4336]` warning and resolve the promise as a no-op so init proceeds. Other stores keep their existing behavior. The values that don't get persisted are session metadata that the SW reads with sensible fallbacks if missing or stale, so push delivery is unaffected.

2. **`1ae8abdf` — perf: short-circuit after first wedge.** Once a single `Options` `readwrite` times out we know the store is poisoned for the rest of this page's lifetime. Add a module-scoped `optionsWriteWedged` flag so the remaining 7 Options writes in `initSaveState` + `saveInitOptions` short-circuit immediately instead of each independently paying the 1500ms timeout. Cuts init latency on a wedged page from ~12s to ~1.5s. The flag is page-scoped (resets on navigation), so a subsequent navigation will probe the wedge fresh.

3. **`88e4cd59` — chore(preview): repro sandbox** (will be reverted before merge). Two-page demo + dev-server hardening that lets a reviewer reproduce the original 30-minute hang on `main` and confirm this branch resolves it. Detailed testing instructions in a separate PR comment.

A 4th commit (`3c41181b`) generalizing the timeout to *all* readwrite stores was tried and reverted (`83cbff87`) because we couldn't validate it on device — every subsequent on-device repro happened with an empty operation queue. Parked in branch history with the validation steps captured in a Linear comment on SDK-4336.

# Systems Affected

- [x] WebSDK
- [ ] Backend
- [ ] Dashboard

# Validation

### Info

- Full suite: **512/512 pass**, lint clean, formatter clean.
- The fix path is exercised by the existing `client.test.ts` round-trip tests (every `Options` write goes through the new wrapper, so all 12 client tests are effectively also covering the timeout path's happy case).
- On-device verification on iPhone running iOS 26.4 (logs11.txt → logs14.txt in the chat): pre-fix init hung indefinitely; post-fix init completes within ~1.5s on the wedged navigation, with a `[SDK-4336] db.put(Options) timed out … Tripping circuit breaker` warning and 7 follow-up `db.put(Options) skipped` warnings, then `internalInit` → SW handshake → `sessionInit` proceed normally.
- I haven't added a unit test specifically for the timeout path because the existing `client.test.ts` uses `fake-indexeddb` which doesn't reproduce the WebKit-specific wedge — a synthetic timeout test would only verify the timer plumbing, which is straightforward enough that the on-device evidence is more meaningful.

### Checklist

**Programming Checklist**

Interfaces:

Functions:

Typescript:

- [x] No Typescript warnings

Other:

- [x] Avoid using global OneSignal accessor for `context` if possible.

### Info

N/A — runtime correctness fix, no UI changes. See on-device console logs in the SDK-4336 Linear ticket and chat history.

### Checklist

---

## Related Tickets

[SDK-4336](https://linear.app/onesignal/issue/SDK-4336/bug-onesignalinit-hangs-on-subsequent-page-navigations-in-ios-safari)

### DIFF
--- a/preview/vite.config.ts
+++ b/preview/vite.config.ts
@@ -43,8 +43,25 @@ export default defineConfig({
     // SDK fetch URL (e.g. `http://localhost:4000/...` in HTTP mode) lands here.
     port: useHttps ? 4001 : 4000,
     strictPort: true,
+    // HMR's WebSocket targets `wss://localhost:<port>`, which iOS devices on a
+    // tunnel (ngrok, etc.) can't reach. Disabling stops the runaway
+    // reconnect loop that floods the console with `ws.send` rejections.
+    hmr: false,
   },
   plugins: [
+    {
+      // Vite's dev server injects `<script src="/@vite/client">` into every
+      // HTML response, even when `server.hmr` is false. That client then
+      // retries a WebSocket forever, which on a tunneled iOS device floods
+      // the console with `ws.send` rejections and is heavy enough to
+      // skew our SDK timing. Strip it out for the preview server, where
+      // we don't need HMR.
+      name: 'strip-vite-client',
+      enforce: 'post',
+      transformIndexHtml(html) {
+        return html.replace(/<script[^>]*src="\/@vite\/client"[^>]*><\/script>\s*/g, '');
+      },
+    },
     ...(useHttps
       ? [
           mkcert({
@@ -87,7 +104,29 @@ export default defineConfig({
           if (contentType) {
             res.setHeader('Content-Type', contentType);
           }
+          // Avoid stale SDK bundles being served from the browser HTTP cache,
+          // which is especially aggressive on iOS Safari/PWA. Without this,
+          // rebuilding the SDK during a debug session won't take effect on the
+          // device until you fully wipe website data.
+          res.setHeader('Cache-Control', 'no-store, no-cache, must-revalidate');
+          res.setHeader('Pragma', 'no-cache');
           fs.createReadStream(filePath).pipe(res);
+        });
+      },
+    },
+    {
+      // Same hygiene for the sandbox HTML (pageA/pageB/index) and root SW
+      // file. Vite's default dev-server cache policy is fine for source files
+      // but our HTML directly references the SDK and embeds repro state.
+      name: 'no-store-html',
+      configureServer(server) {
+        server.middlewares.use((req, res, next) => {
+          const url = req.url ?? '';
+          if (/\.(html|json|js)(\?|$)/.test(url) && !url.startsWith('/sdks/')) {
+            res.setHeader('Cache-Control', 'no-store, no-cache, must-revalidate');
+            res.setHeader('Pragma', 'no-cache');
+          }
+          next();
         });
       },
     },
--- a/src/shared/database/client.ts
+++ b/src/shared/database/client.ts
@@ -105,6 +105,12 @@ export const getDb = (version = VERSION) => {
 const OPTIONS_WRITE_TIMEOUT_MS = 1500;
 let optionsWriteWedged = false;
 
+export const isOptionsWriteWedged = () => optionsWriteWedged;
+
+// `op` is invoked synchronously so the timeout scopes only to the readwrite
+// request itself — DB open/upgrade time is awaited by callers before they
+// hand us a sync closure, so a slow `open()` (e.g. cross-tab `blocked` event,
+// schema migration on a slow device) won't false-trip the breaker.
 function guardOptionsWrite<T>(
   storeName: IDBStoreName,
   label: string,
@@ -152,13 +158,13 @@ export const db = {
     return (await dbPromise).getAll(storeName);
   },
   async put<K extends IDBStoreName>(storeName: K, value: IndexedDBSchema[K]['value']) {
-    return guardOptionsWrite(storeName, `put(${storeName})`, async () =>
-      (await dbPromise).put(storeName, value),
-    );
+    const _db = await dbPromise;
+    return guardOptionsWrite(storeName, `put(${storeName})`, () => _db.put(storeName, value));
   },
   async delete<K extends IDBStoreName>(storeName: K, key: IndexedDBSchema[K]['key']) {
-    return guardOptionsWrite(storeName, `delete(${storeName}/${String(key)})`, async () =>
-      (await dbPromise).delete(storeName, key),
+    const _db = await dbPromise;
+    return guardOptionsWrite(storeName, `delete(${storeName}/${String(key)})`, () =>
+      _db.delete(storeName, key),
     );
   },
 };
--- a/src/shared/database/client.ts
+++ b/src/shared/database/client.ts
@@ -93,35 +93,26 @@ export const getDb = (version = VERSION) => {
   return dbPromise;
 };
 
-// SDK-4336: iOS 26 Safari PWA, on the navigation back to a page after the
-// user has subscribed to push, leaves the `ONE_SIGNAL_SDK_DB` database in
-// a state where every `readwrite` request on the `Options` object store
-// stalls indefinitely (no `success`, no `error`, no transaction lifecycle
-// event). Fresh connections inherit the same per-store WebKit lock state,
-// so reopening doesn't help. Reads work, and `readwrite` on other stores
-// works — only `readwrite` on `Options` is poisoned.
-//
-// Without this guard, `OneSignal.init()` blocks on the very first Options
-// `put` in `initSaveState`/`saveInitOptions` and never returns, which the
-// support team has observed as init "hanging for ~30 minutes" (the time
-// it eventually takes WebKit's internal watchdog to abort the wedged
-// transaction).
-//
-// Workaround: cap `readwrite` operations on the `Options` store with a
-// short timeout. On timeout, we log a warning, trip a circuit breaker
-// for the rest of the page's lifetime so subsequent Options writes
-// short-circuit instead of each paying the timeout, and resolve the
-// promise as a no-op so init proceeds. The values that didn't get
-// persisted are session metadata (`pageTitle`, `persistNotification`,
-// webhook URLs, click-handler config, `lastPushToken`, `isPushEnabled`,
-// etc.) that the service worker reads with sensible fallbacks if
-// missing or stale.
+// On iOS Safari PWA after a push subscription, `readwrite` requests on the
+// `Options` object store can stall indefinitely (no success/error/abort).
+// Other stores and reads are unaffected, and reopening the DB doesn't help.
+// Without this guard, `OneSignal.init()` hangs until WebKit's watchdog
+// eventually aborts the transaction (~30 minutes). Workaround: cap Options
+// writes with a short timeout, then trip a page-scoped circuit breaker so
+// subsequent writes short-circuit. The values that fail to persist are
+// session metadata the SW reads with sensible fallbacks. Remove if WebKit
+// ever fixes the underlying bug.
 const OPTIONS_WRITE_TIMEOUT_MS = 1500;
 let optionsWriteWedged = false;
 
-function withOptionsWriteTimeout<T>(label: string, op: () => Promise<T>): Promise<T | undefined> {
+function guardOptionsWrite<T>(
+  storeName: IDBStoreName,
+  label: string,
+  op: () => Promise<T>,
+): Promise<T | undefined> {
+  if (storeName !== 'Options') return op();
   if (optionsWriteWedged) {
-    Log._warn(`[SDK-4336] db.${label} skipped (wedged)`);
+    Log._warn(`db.${label} skipped (Options store wedged)`);
     return Promise.resolve(undefined);
   }
   return new Promise<T | undefined>((resolve, reject) => {
@@ -130,7 +121,7 @@ function withOptionsWriteTimeout<T>(label: string, op: () => Promise<T>): Promis
       if (settled) return;
       settled = true;
       optionsWriteWedged = true;
-      Log._warn(`[SDK-4336] db.${label} timed out; tripping circuit breaker`);
+      Log._warn(`db.${label} timed out; tripping Options-write circuit breaker`);
       resolve(undefined);
     }, OPTIONS_WRITE_TIMEOUT_MS);
     op().then(
@@ -161,18 +152,14 @@ export const db = {
     return (await dbPromise).getAll(storeName);
   },
   async put<K extends IDBStoreName>(storeName: K, value: IndexedDBSchema[K]['value']) {
-    const op = async () => (await dbPromise).put(storeName, value);
-    if (storeName === 'Options') {
-      return withOptionsWriteTimeout(`put(${storeName})`, op);
-    }
-    return op();
+    return guardOptionsWrite(storeName, `put(${storeName})`, async () =>
+      (await dbPromise).put(storeName, value),
+    );
   },
   async delete<K extends IDBStoreName>(storeName: K, key: IndexedDBSchema[K]['key']) {
-    const op = async () => (await dbPromise).delete(storeName, key);
-    if (storeName === 'Options') {
-      return withOptionsWriteTimeout(`delete(${storeName}/${String(key)})`, op);
-    }
-    return op();
+    return guardOptionsWrite(storeName, `delete(${storeName}/${String(key)})`, async () =>
+      (await dbPromise).delete(storeName, key),
+    );
   },
 };
 
--- a/src/shared/database/client.ts
+++ b/src/shared/database/client.ts
@@ -121,7 +121,7 @@ let optionsWriteWedged = false;
 
 function withOptionsWriteTimeout<T>(label: string, op: () => Promise<T>): Promise<T | undefined> {
   if (optionsWriteWedged) {
-    Log._warn(`[SDK-4336] db.${label} skipped; Options store is known wedged for this page.`);
+    Log._warn(`[SDK-4336] db.${label} skipped (wedged)`);
     return Promise.resolve(undefined);
   }
   return new Promise<T | undefined>((resolve, reject) => {
@@ -130,12 +130,7 @@ function withOptionsWriteTimeout<T>(label: string, op: () => Promise<T>): Promis
       if (settled) return;
       settled = true;
       optionsWriteWedged = true;
-      Log._warn(
-        `[SDK-4336] db.${label} timed out after ${OPTIONS_WRITE_TIMEOUT_MS}ms; ` +
-          `IndexedDB Options store is wedged (likely iOS Safari PWA after push ` +
-          `subscription). Tripping circuit breaker; subsequent Options writes ` +
-          `on this page will be skipped immediately.`,
-      );
+      Log._warn(`[SDK-4336] db.${label} timed out; tripping circuit breaker`);
       resolve(undefined);
     }, OPTIONS_WRITE_TIMEOUT_MS);
     op().then(
--- a/src/shared/database/client.ts
+++ b/src/shared/database/client.ts
@@ -93,7 +93,58 @@ export const getDb = (version = VERSION) => {
   return dbPromise;
 };
 
-// Export db object with the same API as before
+// SDK-4336: iOS 26 Safari PWA, on the navigation back to a page after the
+// user has subscribed to push, leaves the `ONE_SIGNAL_SDK_DB` database in
+// a state where every `readwrite` request on the `Options` object store
+// stalls indefinitely (no `success`, no `error`, no transaction lifecycle
+// event). Fresh connections inherit the same per-store WebKit lock state,
+// so reopening doesn't help. Reads work, and `readwrite` on other stores
+// works — only `readwrite` on `Options` is poisoned.
+//
+// Without this guard, `OneSignal.init()` blocks on the very first Options
+// `put` in `initSaveState`/`saveInitOptions` and never returns, which the
+// support team has observed as init "hanging for ~30 minutes" (the time
+// it eventually takes WebKit's internal watchdog to abort the wedged
+// transaction).
+//
+// Workaround: cap `readwrite` operations on the `Options` store with a
+// short timeout. On timeout, we log a warning and resolve the promise as
+// a no-op so init proceeds. The values that didn't get persisted are
+// session metadata (`pageTitle`, `persistNotification`, webhook URLs,
+// click-handler config, `lastPushToken`, `isPushEnabled`, etc.) that the
+// service worker reads with sensible fallbacks if missing or stale.
+const OPTIONS_WRITE_TIMEOUT_MS = 1500;
+
+function withOptionsWriteTimeout<T>(label: string, op: () => Promise<T>): Promise<T | undefined> {
+  return new Promise<T | undefined>((resolve, reject) => {
+    let settled = false;
+    const timer = setTimeout(() => {
+      if (settled) return;
+      settled = true;
+      Log._warn(
+        `[SDK-4336] db.${label} timed out after ${OPTIONS_WRITE_TIMEOUT_MS}ms; ` +
+          `IndexedDB Options store is wedged (likely iOS Safari PWA after push ` +
+          `subscription). Skipping write to avoid blocking init.`,
+      );
+      resolve(undefined);
+    }, OPTIONS_WRITE_TIMEOUT_MS);
+    op().then(
+      (v) => {
+        if (settled) return;
+        settled = true;
+        clearTimeout(timer);
+        resolve(v);
+      },
+      (e) => {
+        if (settled) return;
+        settled = true;
+        clearTimeout(timer);
+        reject(e);
+      },
+    );
+  });
+}
+
 export const db = {
   async get<K extends IDBStoreName>(
     storeName: K,
@@ -105,10 +156,18 @@ export const db = {
     return (await dbPromise).getAll(storeName);
   },
   async put<K extends IDBStoreName>(storeName: K, value: IndexedDBSchema[K]['value']) {
-    return (await dbPromise).put(storeName, value);
+    const op = async () => (await dbPromise).put(storeName, value);
+    if (storeName === 'Options') {
+      return withOptionsWriteTimeout(`put(${storeName})`, op);
+    }
+    return op();
   },
   async delete<K extends IDBStoreName>(storeName: K, key: IndexedDBSchema[K]['key']) {
-    return (await dbPromise).delete(storeName, key);
+    const op = async () => (await dbPromise).delete(storeName, key);
+    if (storeName === 'Options') {
+      return withOptionsWriteTimeout(`delete(${storeName}/${String(key)})`, op);
+    }
+    return op();
   },
 };
 
--- a/src/shared/database/client.ts
+++ b/src/shared/database/client.ts
@@ -94,34 +94,36 @@ export const getDb = (version = VERSION) => {
 };
 
 // SDK-4336: iOS 26 Safari PWA, on the navigation back to a page after the
-// user has subscribed to push, leaves the `ONE_SIGNAL_SDK_DB` database in
-// a state where every `readwrite` request on the `Options` object store
-// stalls indefinitely (no `success`, no `error`, no transaction lifecycle
-// event). Fresh connections inherit the same per-store WebKit lock state,
-// so reopening doesn't help. Reads work, and `readwrite` on other stores
-// works — only `readwrite` on `Options` is poisoned.
+// user has subscribed to push, leaves the `ONE_SIGNAL_SDK_DB` database
+// in a state where `readwrite` requests stall indefinitely (no
+// `success`, no `error`, no transaction lifecycle event). Fresh
+// connections inherit the same WebKit lock state, so reopening doesn't
+// help. Reads still work; only `readwrite` is poisoned.
 //
-// Without this guard, `OneSignal.init()` blocks on the very first Options
-// `put` in `initSaveState`/`saveInitOptions` and never returns, which the
-// support team has observed as init "hanging for ~30 minutes" (the time
-// it eventually takes WebKit's internal watchdog to abort the wedged
-// transaction).
+// Without this guard, `OneSignal.init()` blocks on the first `Options`
+// `put` in `initSaveState`/`saveInitOptions` and never returns, and
+// even after init the operation queue (`OperationRepo`) gets stuck
+// because `_executeOperations` awaits writes to the `operations` /
+// `identity` / `properties` / `pushSubscriptions` stores that never
+// settle. The support team has observed this as init "hanging for ~30
+// minutes" (the time it eventually takes WebKit's internal watchdog to
+// abort the wedged transaction) plus persistent failure to flush
+// queued user/subscription operations.
 //
-// Workaround: cap `readwrite` operations on the `Options` store with a
-// short timeout. On timeout, we log a warning, trip a circuit breaker
-// for the rest of the page's lifetime so subsequent Options writes
-// short-circuit instead of each paying the timeout, and resolve the
-// promise as a no-op so init proceeds. The values that didn't get
-// persisted are session metadata (`pageTitle`, `persistNotification`,
-// webhook URLs, click-handler config, `lastPushToken`, `isPushEnabled`,
-// etc.) that the service worker reads with sensible fallbacks if
-// missing or stale.
-const OPTIONS_WRITE_TIMEOUT_MS = 1500;
-let optionsWriteWedged = false;
+// Workaround: cap every `readwrite` op (`put` / `delete` / `clear`)
+// with a short timeout. On timeout, log a `[SDK-4336]` warning, trip a
+// page-scoped circuit breaker so subsequent readwrites short-circuit
+// instead of each paying the timeout, and resolve the promise as a
+// no-op so callers proceed. The values that don't get persisted are
+// either session metadata the SW re-derives from network state on the
+// next visit, or queued operations whose effects are idempotent and
+// will be re-attempted next session.
+const READWRITE_TIMEOUT_MS = 1500;
+let readwriteWedged = false;
 
-function withOptionsWriteTimeout<T>(label: string, op: () => Promise<T>): Promise<T | undefined> {
-  if (optionsWriteWedged) {
-    Log._warn(`[SDK-4336] db.${label} skipped; Options store is known wedged for this page.`);
+function withReadwriteTimeout<T>(label: string, op: () => Promise<T>): Promise<T | undefined> {
+  if (readwriteWedged) {
+    Log._warn(`[SDK-4336] db.${label} skipped; IndexedDB readwrite is wedged for this page.`);
     return Promise.resolve(undefined);
   }
   return new Promise<T | undefined>((resolve, reject) => {
@@ -129,15 +131,15 @@ function withOptionsWriteTimeout<T>(label: string, op: () => Promise<T>): Promis
     const timer = setTimeout(() => {
       if (settled) return;
       settled = true;
-      optionsWriteWedged = true;
+      readwriteWedged = true;
       Log._warn(
-        `[SDK-4336] db.${label} timed out after ${OPTIONS_WRITE_TIMEOUT_MS}ms; ` +
-          `IndexedDB Options store is wedged (likely iOS Safari PWA after push ` +
-          `subscription). Tripping circuit breaker; subsequent Options writes ` +
-          `on this page will be skipped immediately.`,
+        `[SDK-4336] db.${label} timed out after ${READWRITE_TIMEOUT_MS}ms; ` +
+          `IndexedDB is wedged (likely iOS Safari PWA after push subscription). ` +
+          `Tripping circuit breaker; subsequent readwrites on this page will be ` +
+          `skipped immediately.`,
       );
       resolve(undefined);
-    }, OPTIONS_WRITE_TIMEOUT_MS);
+    }, READWRITE_TIMEOUT_MS);
     op().then(
       (v) => {
         if (settled) return;
@@ -166,23 +168,21 @@ export const db = {
     return (await dbPromise).getAll(storeName);
   },
   async put<K extends IDBStoreName>(storeName: K, value: IndexedDBSchema[K]['value']) {
-    const op = async () => (await dbPromise).put(storeName, value);
-    if (storeName === 'Options') {
-      return withOptionsWriteTimeout(`put(${storeName})`, op);
-    }
-    return op();
+    return withReadwriteTimeout(`put(${storeName})`, async () =>
+      (await dbPromise).put(storeName, value),
+    );
   },
   async delete<K extends IDBStoreName>(storeName: K, key: IndexedDBSchema[K]['key']) {
-    const op = async () => (await dbPromise).delete(storeName, key);
-    if (storeName === 'Options') {
-      return withOptionsWriteTimeout(`delete(${storeName}/${String(key)})`, op);
-    }
-    return op();
+    return withReadwriteTimeout(`delete(${storeName}/${String(key)})`, async () =>
+      (await dbPromise).delete(storeName, key),
+    );
   },
 };
 
 export const clearStore = async <K extends IDBStoreName>(storeName: K) => {
-  return (await dbPromise).clear(storeName);
+  return withReadwriteTimeout(`clear(${storeName})`, async () =>
+    (await dbPromise).clear(storeName),
+  );
 };
 
 export const getObjectStoreNames = async () => {
--- a/src/shared/database/client.ts
+++ b/src/shared/database/client.ts
@@ -101,7 +101,7 @@ export const getDb = (version = VERSION) => {
 // writes with a short timeout, then trip a page-scoped circuit breaker so
 // subsequent writes short-circuit. The values that fail to persist are
 // session metadata the SW reads with sensible fallbacks. Remove if WebKit
-// ever fixes the underlying bug.
+// ever fixes the underlying bug: https://bugs.webkit.org/show_bug.cgi?id=315804
 const OPTIONS_WRITE_TIMEOUT_MS = 1500;
 let optionsWriteWedged = false;
 
--- a/src/shared/database/client.ts
+++ b/src/shared/database/client.ts
@@ -107,44 +107,26 @@ let optionsWriteWedged = false;
 
 export const isOptionsWriteWedged = () => optionsWriteWedged;
 
-// `op` is invoked synchronously so the timeout scopes only to the readwrite
-// request itself — DB open/upgrade time is awaited by callers before they
-// hand us a sync closure, so a slow `open()` (e.g. cross-tab `blocked` event,
-// schema migration on a slow device) won't false-trip the breaker.
+// `op` is invoked synchronously (callers await `dbPromise` first), so the
+// timeout scopes only to the readwrite request, not DB open/upgrade. Once a
+// write times out we trip a page-scoped circuit breaker so the rest of init's
+// Options writes short-circuit instead of each paying the full timeout.
 function guardOptionsWrite<T>(
   storeName: IDBStoreName,
   label: string,
   op: () => Promise<T>,
 ): Promise<T | undefined> {
   if (storeName !== 'Options') return op();
-  if (optionsWriteWedged) {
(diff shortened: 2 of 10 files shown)


sherwinski:
## Testing instructions for reviewers

The sandbox commit (`chore(preview): add iOS PWA repro sandbox`) is included only so you can reproduce the bug end-to-end on a real iOS device. **It will be reverted before merge** — please don't ship it.

You'll need:

- A real iOS device running iOS 26.x (the bug doesn't surface in the iOS Simulator's WebKit). Mine: iPhone running iOS 26.4.
- An ngrok / Cloudflare tunnel account so the iOS device can reach your dev machine's HTTPS server.
- A OneSignal app whose **Site URL exactly matches the tunnel host** you'll be using, with a custom service-worker integration pointing to `push/onesignal/OneSignalSDKWorker.js`. iOS Web Push enforces an exact origin match, so a pre-existing app pointing at `https://localhost:4001` fails immediately with `Error: Can only be used on: https://localhost:4001`. Easiest path: create a fresh disposable test app in the dashboard.

### One-time setup

```bash
git checkout sherwin/sdk-4336

# 1. Start the tunnel FIRST so you know your host. Free ngrok now issues
#    *.ngrok-free.dev, and the host is RANDOM per restart, so every new host
#    needs a fresh build (step 3) AND a dashboard Site URL update.
ngrok http https://localhost:4001

# 2. Capture the BARE host — no scheme, no port. Pasting the full https:// URL
#    bakes a broken `https://https://...` origin; a stray `:4001` also fails
#    because ngrok serves 443.
HOST=$(curl -s localhost:4040/api/tunnels | jq -r '.tunnels[0].public_url' | sed 's#https://##')
echo "$HOST"   # e.g. abc1-2-3.ngrok-free.dev

# 3. Build the dev SDK against your tunnel host. NO_DEV_PORT=true drops the port
#    so the device fetches assets from https://$HOST/... (443) through the tunnel.
BUILD_ORIGIN=$HOST NO_DEV_PORT=true vp run build:dev-prod

# 4. Serve the prebuilt SDK. This MUST run from the preview/ directory — that is
#    what binds https://localhost:4001. Do NOT use `vp dev --filter @onesignal/preview`
#    (the package is named `preview`, so the filter matches nothing and vp serves
#    the root config on :4000), and do NOT use `vp run start` (it re-runs build:dev
#    and re-bakes BUILD_ORIGIN=localhost, clobbering step 3).
cd preview && SDK_ENV=dev vp dev
```

Sanity-check the baked origin before testing (exactly one `https://`, bare host, no port):

```bash
rg "BUILD_ORIGIN = " build/releases/Dev-OneSignalSDK.page.js
# var BUILD_ORIGIN = "abc1-2-3.ngrok-free.dev";
```

Then set the OneSignal app's **Site URL** to `https://$HOST/`.

### Reproduce the bug (control)

The control is the repro sandbox **without** the fix. `main` does not contain the sandbox pages (`pageA.html` / `pageB.html` live only in the sandbox commit), so don't `git checkout main` and open them — overlay just the `preview/` dir onto main instead:

```bash
git checkout main
git checkout sherwin/sdk-4336 -- preview          # sandbox pages, no fix
BUILD_ORIGIN=$HOST NO_DEV_PORT=true vp run build:dev-prod
cd preview && SDK_ENV=dev vp dev
```

1. On the iPhone, open `https://$HOST/pageA.html?app_id=<APP_ID>` in Safari.
2. Tap the Share icon → **Add to Home Screen** → Add. Keep **Open as a Web App** toggled **on** (the bug only reproduces in a standalone PWA, not a Safari tab/bookmark).
3. Open the installed PWA from the Home Screen. Confirm there's **no Safari URL bar** — that's how you know it's standalone.
4. Tap **Go to Page B** → tap **Go to Page A** → tap **Register** → accept the system prompt.
5. **Verify you actually subscribed** — this is the precondition for the wedge, which only happens *after a push subscription exists*. Attach your Mac (Safari → Develop → [device] → [the PWA's web view]) and run:
   ```js
   await OneSignal.User.PushSubscription.id   // must be a non-null GUID
   OneSignal.Notifications.permission         // true
   ```
   If `id` is null, `init()` will keep completing normally and the bug will *not* reproduce — fix the subscription first (service worker reachable at `/push/onesignal/OneSignalSDKWorker.js`, dashboard app config correct).
6. Now trigger a **fresh `init()`** while subscribed. The wedge isn't tied to one specific gesture — any path that re-runs `OneSignal.init()` in the standalone PWA after the subscription exists will hit it. The likely ways you'll encounter it in the wild, any one of which reproduces:
   - **In-app navigation:** tap **Go to Page B** → tap **Go to Page A** (each page calls `init()`).
   - **Relaunch the app:** force-quit the PWA (swipe it out of the app switcher) and reopen it from the Home Screen. This is the most common real-world trigger — users closing and reopening the installed web app.
   - **Cold start after backgrounding:** leave the PWA backgrounded long enough that iOS tears down the web view, then reopen.
7. On that fresh load, with the inspector attached *before* you trigger it: the body renders and you'll see the `init()` / `CoreModule.init()` debug lines, then the console **freezes with no `OneSignal initialized (Nms)` log line**. Leave it ~30 minutes and it eventually resumes by itself — that's the WebKit watchdog firing.
   > If you relaunched the PWA (rather than navigating), the Web Inspector won't follow automatically — re-select the PWA's web view under Safari → Develop → [device] after reopening to see the frozen console.

> Note: `init()` completing on the *earlier* loads (before you're subscribed) is expected and correct. Only a load that runs `init()` **after a push subscription exists** wedges.

### Verify the fix

```bash
git checkout sherwin/sdk-4336
BUILD_ORIGIN=$HOST NO_DEV_PORT=true vp run build:dev-prod
cd preview && SDK_ENV=dev vp dev
```

Reset state on the device (delete the PWA from the Home Screen, then Settings → Apps → Safari → Advanced → Website Data → remove the tunnel host's data), reinstall, and run the same sequence (including a force-quit/reopen, since that's the most common trigger). On the wedged load you should see roughly:

```
[Log]     !!!! [SDK-4336 PAGE A] OneSignal initialize
[Debug]   init()
...
[Warning] db.put(Options) timed out
...
[Log]     !!!! [SDK-4336 PAGE A] OneSignal initialized (~1500ms)
```

One Options write times out at 1500ms and trips the circuit breaker; the remaining Options writes short-circuit silently (no per-write log). Total init on the wedged load is ~1.5s instead of indefinite. No `Ops in progress` runaway loop, no exceptions, push delivery still works.

### What I confirmed on my device

- **Pre-fix:** init never completed in 30+ minutes of waiting (reproduced via both in-app navigation and force-quit/reopen after registering).
- **With the fast-fail timeout only** (before the short-circuit optimization): init completes after ~12s (8 separate Options writes each timing out individually at 1.5s).
- **With the timeout + circuit breaker** (shipped behavior): init completes after ~1.5s (one timeout, the rest short-circuit).

Captured logs are in the SDK-4336 Linear ticket if you want to compare to your own run.

### Pre-merge

Before merging, drop the sandbox commit (`chore(preview): add iOS PWA repro sandbox`) — e.g. `git rebase -i origin/main` and drop that commit — then `git push --force-with-lease`. Only the fix commits need to ship.

preview/vite.config.ts
```diff
@@ -43,8 +43,25 @@
...
+      // Vite's dev server injects `<script src="/@vite/client">` into every
+      // HTML response, even when `server.hmr` is false. That client then
+      // retries a WebSocket forever, which on a tunneled iOS device floods
+      // the console with `ws.send` rejections and is heavy enough to
+      // skew our SDK timing. Strip it out for the preview server, where
+      // we don't need HMR.
+      name: 'strip-vite-client',
+      enforce: 'post',
+      transformIndexHtml(html) {
+        return html.replace(/<script[^>]*src="\/@vite\/client"[^>]*><\/script>\s*/g, '');
```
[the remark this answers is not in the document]
sherwinski: Safe to ignore

sherwinski:
Note: we'll also roll back the `[SDK-4336]...` logs before merging.

sherwinski:
@claude review

src/shared/database/client.ts
```diff
@@ -93,7 +93,63 @@ export const getDb = (version = VERSION) => {
...
+      },
+      (e) => {
+        if (settled) return;
+        settled = true;
+        clearTimeout(timer);
+        reject(e);
+      },
+    );
+  });
+}
```
fadi-george:
```suggestion
// SDK-4336: iOS 26 Safari PWA leaves the DB in a state where every `readwrite` on the `Options` store stalls forever
// (reopening doesn't help; only `Options` writes are poisoned), hanging `OneSignal.init()`. Workaround: cap Options
// writes with a short timeout; on timeout trip a circuit breaker so later writes short-circuit, and resolve as a no-op.
// The unpersisted values are session metadata the SW reads with fallbacks.
const OPTIONS_WRITE_TIMEOUT_MS = 1500;
let optionsWriteWedged = false;
function guardOptionsWrite<T>(
  storeName: IDBStoreName,
  label: string,
  op: () => Promise<T>,
): Promise<T | undefined> {
  if (storeName !== 'Options') return op();
  if (optionsWriteWedged) return Promise.resolve(undefined);
  let timer: ReturnType<typeof setTimeout>;
  const timeout = new Promise<undefined>((resolve) => {
    timer = setTimeout(() => {
      optionsWriteWedged = true;
      Log._warn(`${label} timed out`);
      resolve(undefined);
    }, OPTIONS_WRITE_TIMEOUT_MS);
  });
  return Promise.race([op(), timeout]).finally(() => clearTimeout(timer));
}
```

src/shared/database/client.ts
```diff
@@ -105,10 +161,18 @@ export const db = {
     return (await dbPromise).getAll(storeName);
   },
   async put<K extends IDBStoreName>(storeName: K, value: IndexedDBSchema[K]['value']) {
-    return (await dbPromise).put(storeName, value);
+    const op = async () => (await dbPromise).put(storeName, value);
+    if (storeName === 'Options') {
+      return withOptionsWriteTimeout(`put(${storeName})`, op);
+    }
+    return op();
```
fadi-george: could make it dry with helper function

src/shared/database/client.ts
```diff
@@ -93,7 +93,63 @@ export const getDb = (version = VERSION) => {
...
+// Workaround: cap `readwrite` operations on the `Options` store with a
+// short timeout. On timeout, we log a warning, trip a circuit breaker
+// for the rest of the page's lifetime so subsequent Options writes
+// short-circuit instead of each paying the timeout, and resolve the
+// promise as a no-op so init proceeds. The values that didn't get
+// persisted are session metadata (`pageTitle`, `persistNotification`,
+// webhook URLs, click-handler config, `lastPushToken`, `isPushEnabled`,
+// etc.) that the service worker reads with sensible fallbacks if
+// missing or stale.
+const OPTIONS_WRITE_TIMEOUT_MS = 1500;
```
fadi-george, line 105: How'd you settle on this value?

sherwinski: Just by tinkering with it and trying to find a balance between letting it hang too long and letting a non-poisoned state operate normally. We can increase it, since either way it will be a lot better than the current ~30 min it takes to resolve itself.

src/shared/database/client.ts
sherwinski: Took the spirit of this in `c3f8e5b2` — collapsed the two branches and the helper into one `guardOptionsWrite(storeName, label, op)`. Kept the explicit `settled` flag instead of `Promise.race`, since `race` causes the wrapped op's later rejection to become an unhandled promise rejection (timeout fires first → race resolves `undefined`, then op's eventual reject has no handler). Also kept the short-circuit `Log._warn` because it's useful for support diagnostics. Comment block trimmed at the same time.

src/shared/database/client.ts
sherwinski: Done in c3f8e5b2 — both `put` and `delete` now go through a single `guardOptionsWrite` helper that handles the `storeName === 'Options'` gate internally.

src/shared/database/client.ts
fadi-george: So what happens if it does fail, do operations just linger?

sherwinski: On timeout we resolve the wrapper promise with `undefined` and trip the breaker, so the JS caller proceeds. The underlying `IDBRequest` itself isn't cancellable — it stays queued in WebKit's transaction queue and eventually settles (~30 min later, when the OS watchdog aborts the wedged transaction). By then the page is usually unloaded. Our `op().then(...)` handler has an `if (settled) return;` guard, so the eventual settlement is a no-op for us — no unhandled rejection, no late state mutation. Net effect on the page: zero; one zombie IDB request lingers on the WebKit-internal side until the OS reaps it.

src/shared/database/client.ts
```diff
@@ -93,7 +93,54 @@ export const getDb = (version = VERSION) => {
...
+function guardOptionsWrite<T>(
+  storeName: IDBStoreName,
+  label: string,
+  op: () => Promise<T>,
+): Promise<T | undefined> {
+  if (storeName !== 'Options') return op();
+  if (optionsWriteWedged) {
+    Log._warn(`db.${label} skipped (Options store wedged)`);
+    return Promise.resolve(undefined);
+  }
```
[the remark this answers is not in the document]
sherwinski: Good catch — confirmed by reading `initSaveState`. The `Ids` store isn't guarded, so on a wedged Options reset the `Ids.appId` commit would succeed and pin the new appId while the previous app's metadata stayed put, with no path back into the reset branch on later loads.

Fixed in the next commit (incoming): exported `isOptionsWriteWedged()` from `client.ts` and added an early `return` in `initSaveState` between the four `Options` resets and the rest of the reset block. If the breaker tripped, we log a warning and skip both the `Ids` clears and the `Ids.appId` commit, letting a future non-wedged load complete the reset cleanly. The `_subscriptionModelStore._clear()` is also gated, so we don't hydrate against a partially-reset store.

src/shared/database/client.ts
```diff
@@ -93,7 +93,54 @@ export const getDb = (version = VERSION) => {
...
+  return new Promise<T | undefined>((resolve, reject) => {
+    let settled = false;
+    const timer = setTimeout(() => {
+      if (settled) return;
+      settled = true;
+      optionsWriteWedged = true;
+      Log._warn(`db.${label} timed out; tripping Options-write circuit breaker`);
+      resolve(undefined);
+    }, OPTIONS_WRITE_TIMEOUT_MS);
+    op().then(
```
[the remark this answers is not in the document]
sherwinski: Valid — confirmed by re-reading `db.put`/`db.delete`: the timer was started in the `Promise` executor, but `op` was an `async` lambda whose first action was `await dbPromise`, so the budget did cover both DB open and the put.

Fixed in the next commit (incoming): `db.put` and `db.delete` now `await dbPromise` *before* invoking `guardOptionsWrite`, and pass a sync closure that goes straight to the readwrite request. The timeout now scopes strictly to the IDB request, not to DB open/upgrade. Updated the helper's leading comment to call this out explicitly.

src/shared/helpers/init.ts
```diff
@@ -352,6 +352,15 @@ export async function initSaveState(overridingPageTitle?: string) {
     await db.put('Options', { key: 'lastPushId', value: null });
     await db.put('Options', { key: 'lastPushToken', value: null });
     await db.put('Options', { key: 'lastOptedIn', value: null });
+    // Bail out if the Options reset got circuit-broken. Committing the new
+    // appId now would strand the previous app's metadata under it, and the
+    // `previousAppId !== appId` gate above would keep us out of this branch
+    // on later loads — leaving the stale values permanent. Skipping the
+    // appId commit instead lets a future non-wedged load complete the reset.
+    if (isOptionsWriteWedged()) {
+      Log._warn('App ID change reset deferred; will retry on next non-wedged load');
```
fadi-george: avoid overusing logs, they bloat sdk size

src/shared/database/client.ts
```diff
@@ -105,10 +158,14 @@ export const db = {
...
   },
   async put<K extends IDBStoreName>(storeName: K, value: IndexedDBSchema[K]['value']) {
-    return (await dbPromise).put(storeName, value);
+    const _db = await dbPromise;
+    return guardOptionsWrite(storeName, `put(${storeName})`, () => _db.put(storeName, value));
   },
   async delete<K extends IDBStoreName>(storeName: K, key: IndexedDBSchema[K]['key']) {
-    return (await dbPromise).delete(storeName, key);
+    const _db = await dbPromise;
+    return guardOptionsWrite(storeName, `delete(${storeName}/${String(key)})`, () =>
```
fadi-george: isnt key always string

src/shared/database/client.ts
```diff
@@ -93,7 +93,60 @@ export const getDb = (version = VERSION) => {
...
+// hand us a sync closure, so a slow `open()` (e.g. cross-tab `blocked` event,
+// schema migration on a slow device) won't false-trip the breaker.
+function guardOptionsWrite<T>(
+  storeName: IDBStoreName,
+  label: string,
+  op: () => Promise<T>,
+): Promise<T | undefined> {
+  if (storeName !== 'Options') return op();
+  if (optionsWriteWedged) {
+    Log._warn(`db.${label} skipped (Options store wedged)`);
```
fadi-george: maybe dont need this log

src/shared/database/client.ts
```diff
@@ -93,7 +93,60 @@ export const getDb = (version = VERSION) => {
...
+const OPTIONS_WRITE_TIMEOUT_MS = 1500;
+let optionsWriteWedged = false;
+
+export const isOptionsWriteWedged = () => optionsWriteWedged;
+
+// `op` is invoked synchronously so the timeout scopes only to the readwrite
+// request itself — DB open/upgrade time is awaited by callers before they
+// hand us a sync closure, so a slow `open()` (e.g. cross-tab `blocked` event,
+// schema migration on a slow device) won't false-trip the breaker.
+function guardOptionsWrite<T>(
```
fadi-george, line 114: could do to optimize for bundle size:
```
function guardOptionsWrite<T>(
  storeName: IDBStoreName,
  label: string,
  op: () => Promise<T>,
): Promise<T | undefined> {
  if (storeName !== 'Options') return op();
  if (optionsWriteWedged) {
    return Promise.resolve(undefined);
  }
  let timer: ReturnType<typeof setTimeout>;
  const timeout = new Promise<undefined>((resolve) => {
    timer = setTimeout(() => {
      optionsWriteWedged = true;
      Log._warn(`db.${label} timed out`);
      resolve(undefined);
    }, OPTIONS_WRITE_TIMEOUT_MS);
  });
  return Promise.race([op(), timeout]).finally(() => clearTimeout(timer));
}
```

fadi-george:
would fadi/sdk-4336-options-write-timeout be enough?

sherwinski:
~~Superseded by #1469 (and the underlying core fix [7f5df216](https://github.com/OneSignal/OneSignal-Website-SDK/commit/7f5df21652bd90ad266c0c8b28520004c6eba481), which landed directly on main while this PR was in review). The new PR keeps just the additive hardening (circuit breaker + appId-deferral + clearTimeout + timeout warn log) on top of the main commit, dropping the demo sandbox and other churn.~~

src/shared/database/client.ts
sherwinski: Adopted in 647e220f — switched to the `Promise.race([op(), timeout]).finally(clearTimeout)` form. Kept the `label` param so the timeout log still names the store/key, but everything else matches your snippet.

src/shared/database/client.ts
sherwinski: Dropped in 647e220f. The single `timed out` log on first wedge is the actionable signal; the per-skip logs were just noise.

src/shared/database/client.ts
sherwinski: Right — Options keys are always strings. Dropped the `String(key)` cast in 647e220f.

src/shared/helpers/init.ts
sherwinski: Dropped in 647e220f — the comment right above the guard already documents the deferral, so the log was redundant.

src/shared/database/client.ts
```diff
@@ -93,7 +93,42 @@ export const getDb = (version = VERSION) => {
...
+  let timer: ReturnType<typeof setTimeout>;
+  const timeout = new Promise<undefined>((resolve) => {
+    timer = setTimeout(() => {
+      optionsWriteWedged = true;
+      Log._warn(`db.${label} timed out`);
+      resolve(undefined);
+    }, OPTIONS_WRITE_TIMEOUT_MS);
+  });
+  return Promise.race([op(), timeout]).finally(() => clearTimeout(timer));
+}
```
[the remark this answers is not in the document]
sherwinski: Will address in a follow-up

fadi-george:
Verified 30min timeout (pre-fix).

fadi-george (approved):

PR #1469, opened 2026-06-01
fix: [SDK-4336] add circuit breaker and defer appId commit on wedge

sherwinski:
## Summary

Two follow-up fixes layered on top of [7f5df216](https://github.com/OneSignal/OneSignal-Website-SDK/commit/7f5df21652bd90ad266c0c8b28520004c6eba481), surfaced during review of the (now-closed) #1468:

- **Circuit breaker** — once any Options write times out, trip a page-scoped flag so the remaining ~7 Options writes during a wedged init short-circuit immediately. Cuts wedged-init time from ~14s (7 × 2s timeout) to ~2s (single timeout, rest skip).
- **Defer the new-appId commit on wedge** — `Ids.appId` is unguarded, so on a wedged page during an app-ID switch the `Ids.appId` write would succeed while the previous app's `Options.{isPushEnabled,lastPushId,lastPushToken,lastOptedIn}` stayed put. The `previousAppId !== appId` gate would then keep the reset branch from re-entering on later loads — making the cross-app contamination permanent. With the new `isOptionsWriteWedged()` accessor, `initSaveState` now bails early on a tripped breaker so a future non-wedged load can finish the reset cleanly.

Also fixes a minor leak — `Promise.race` now uses `.finally(() => clearTimeout(timer))` so healthy pages don't accumulate dangling 2s timers across the dozens of Options writes init issues.

Adds a single `Log._warn` on timeout so support can correlate the breaker tripping with init slowness in console logs.

## Test plan

- [x] `vp test --run` — 512/512 passing.
- [x] `vp check` — clean (5 unrelated pre-existing warnings).
- [x] `vp run build:prod` — passes; bundle deltas covered by a +80 B / +120 B size-limit bump.
- [ ] On-device sanity check on iOS 26 PWA (the original repro path), confirming wedged init now resolves in ~2s and the `db.Options timed out` warning fires once.

## Why a follow-up rather than amending #1468

#1468 had been chasing the same ticket and was effectively superseded by 7f5df216 landing on `main`. Closing it as superseded and re-uploading just the additive hardening on top of `main` keeps this diff minimal and reviewable (3 files, +26/-6).

### DIFF
--- a/package.json
+++ b/package.json
@@ -77,12 +77,12 @@
     },
     {
       "path": "./build/releases/OneSignalSDK.page.es6.js",
-      "limit": "42.4 kB",
+      "limit": "42.5 kB",
       "gzip": true
     },
     {
       "path": "./build/releases/OneSignalSDK.sw.js",
-      "limit": "12.354 kB",
+      "limit": "12.43 kB",
       "gzip": true
     },
     {
--- a/package.json
+++ b/package.json
@@ -77,12 +77,12 @@
     },
     {
       "path": "./build/releases/OneSignalSDK.page.es6.js",
-      "limit": "42.5 kB",
+      "limit": "42.6 kB",
       "gzip": true
     },
     {
       "path": "./build/releases/OneSignalSDK.sw.js",
-      "limit": "12.43 kB",
+      "limit": "12.55 kB",
       "gzip": true
     },
     {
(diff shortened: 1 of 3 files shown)


fadi-george:
just remake your branch

PR #1478, opened 2026-07-22
fix: [SDK-4897] add accessible name to subscription bell launcher button

sherwinski:
# Description

## 1 Line Summary

Add a default, configurable `aria-label` to the Subscription Bell launcher button so it has an accessible name and no longer fails accessibility audits.

## Details

The bell launcher is an icon-only `<button>` with no text content and no `aria-label`, so automated accessibility audits (Lighthouse/PageSpeed, axe) report **"Buttons must have discernible text"** on every customer site that enables the bell. Screen-reader users and agentic browsers relying on the accessibility tree cannot identify the control. Reported by an EU public-sector customer subject to the EU Web Accessibility Directive (WCAG 2.1 AA).

Changes:

- The launcher button now gets `aria-label="Manage notifications"` by default when the bell is created.
- The label is configurable and localizable via the existing bell text config: `notifyButton: { text: { 'launcher.label': '...' } }` (new optional `BellText` key).
- The decorative bell SVG inside the button is marked `aria-hidden="true"` so it doesn't leak into the accessibility tree.
- `Message._getTipForState` gained a `?? ''` fallback because the new optional key widened the text lookup type.
- The gzipped `page.es6.js` size limit was bumped 42.65 kB → 42.7 kB; the limit only had ~10 B of headroom on `main` and this change adds ~50 B.

Other icon-only controls were audited per the ticket: the dialog's subscribe/unsubscribe buttons already have visible text and the dialog is a light-dismiss popover with no close button, so the launcher was the only control needing a name.

# Systems Affected

- [x] WebSDK
- [ ] Backend
- [ ] Dashboard

# Validation

### Info

- New unit tests in `Bell.test.ts` cover the default label, a custom localized label via text config, and the SVG being hidden from the accessibility tree.
- `vp check`, full test suite (526 tests), and `build:prod` (size-limit + bundle API validation) all pass.

### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:

- [x] No Typescript warnings

Other:

### Info

Not needed — no visual change; the only DOM difference is the `aria-label` attribute and `aria-hidden` on the SVG.

### Checklist

---

## Related Tickets

[SDK-4897](https://linear.app/onesignal/issue/SDK-4897/web-sdk-v16-subscription-bell-launcher-button-has-no-accessible-name)

---

### DIFF
--- a/src/page/bell/Bell.ts
+++ b/src/page/bell/Bell.ts
@@ -34,7 +34,7 @@ const VALID_THEMES: readonly string[] = ['default', 'inverse'];
 const DEFAULT_LAUNCHER_LABEL = 'Manage notifications';
 
 const DEFAULT_TEXT: BellText = {
-  'launcher.label': DEFAULT_LAUNCHER_LABEL,
+  'launcher.button.aria-label': DEFAULT_LAUNCHER_LABEL,
   'tip.state.unsubscribed': 'Subscribe to notifications',
   'tip.state.subscribed': "You're subscribed to notifications",
   'tip.state.blocked': "You've blocked notifications",
@@ -218,7 +218,7 @@ export default class Bell {
     );
     this._button._element?.setAttribute(
       'aria-label',
-      this._options.text['launcher.label'] || DEFAULT_LAUNCHER_LABEL,
+      this._options.text['launcher.button.aria-label'] || DEFAULT_LAUNCHER_LABEL,
     );
     addDomElement(
       this._launcher._selector,
--- a/src/page/bell/Bell.ts
+++ b/src/page/bell/Bell.ts
@@ -21,7 +21,7 @@ import Dialog from './Dialog';
 import Launcher from './Launcher';
 import Message from './Message';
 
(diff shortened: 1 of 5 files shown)


sherwinski:
@claude review once

src/page/bell/Bell.ts
```diff
@@ -213,6 +216,10 @@ export default class Bell {
       'beforeend',
       '<button type="button" class="onesignal-bell-launcher-button" popovertarget="onesignal-bell-dialog"></button>',
```
fadi-george, line 217: is this the butter, why not just all the aria-label here and reduce complexity

src/page/bell/Bell.ts
```diff
@@ -31,7 +31,10 @@ const VALID_SIZES: readonly string[] = ['small', 'medium', 'large'];
 const VALID_POSITIONS: readonly string[] = ['bottom-left', 'bottom-right'];
 const VALID_THEMES: readonly string[] = ['default', 'inverse'];
 
+const DEFAULT_LAUNCHER_LABEL = 'Manage notifications';
+
 const DEFAULT_TEXT: BellText = {
+  'launcher.label': DEFAULT_LAUNCHER_LABEL,
```
fadi-george: dont need, just inline? see other comment

src/shared/prompts/types.ts
```diff
@@ -116,6 +116,7 @@ export type BellSize = 'small' | 'medium' | 'large';
 export type BellPosition = 'bottom-left' | 'bottom-right';
 
 export interface BellText {
+  'launcher.label'?: string;
```
fadi-george: ideally we wouldnt need this

fadi-george: name is kind of vague should it be one of these:
- launcher.button.aria-label
- button.aria-label
- bell.aria-label
- bell.button.aria-label

fadi-george:
Ill guess well also need mintlify docs changes later for new label

src/shared/prompts/types.ts
sherwinski: Good point, let's go with `launcher.button.aria-label`

sherwinski: Renamed to `launcher.button.aria-label` in bb13fd4f. Went with that one since it follows the existing `dialog.main.button.*` hierarchy, and unlike the other text keys this one never renders visible text, so naming the attribute it populates seemed worth it. Skipped the `bell.` prefix since every key in this config is already bell-scoped.

fadi-george (approved):

PR #1479, opened 2026-07-22
fix: [SDK-4897] add alt text to bell dialog images

sherwinski:
# Description

## 1 Line Summary

Add alt text to the two images in the bell dialog so they pass "images must have alternate text" and "links must have discernible text" accessibility audits.

## Details

Follow-up to #1478 (merged).

Follow-up to the launcher `aria-label` fix, covering the remaining audit failures inside the dialog (per the ticket: "apply the same treatment to any other icon-only controls in the bell/dialog"):

- The notification icon `` in the subscription view now has `alt=""` — it is decorative next to the dialog text, so it is correctly hidden from the accessibility tree instead of being announced by filename.
- The unblock-instructions image in the blocked view now has `alt="Unblock notifications"`. This also gives its wrapping icon-only `<a>` its accessible name, fixing the "links must have discernible text" failure.
- The gzipped `page.es6.js` size limit was bumped 42.7 kB → 42.75 kB (was 8 B over; the limit is routinely bumped in small increments).

These only surface while the dialog popover is open, so resting-page scans (e.g. PageSpeed) do not catch them, but manual audits of the open dialog would.

# Systems Affected

- [x] WebSDK
- [ ] Backend
- [ ] Dashboard

# Validation

### Info

- Extended existing `Dialog.test.ts` cases to assert the icon's empty alt and the instructions image's non-empty alt.
- `vp check`, full test suite (526 tests), and `build:prod` (size-limit + bundle API validation) all pass.

### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:

- [x] No Typescript warnings

Other:

### Info

Not needed — no visual change; only `alt` attributes were added.

### Checklist

---

## Related Tickets

[SDK-4897](https://linear.app/onesignal/issue/SDK-4897/web-sdk-v16-subscription-bell-launcher-button-has-no-accessible-name)

---

### DIFF
--- a/src/page/bell/Dialog.ts
+++ b/src/page/bell/Dialog.ts
@@ -122,7 +122,7 @@ export default class Dialog {
       const imagePath = UNBLOCK_IMAGES[browserName];
       if (imagePath) {
         const imageUrl = STATIC_RESOURCES_URL + imagePath;
-        instructionsHtml = `<a href="${imageUrl}" target="_blank"><img src="${imageUrl}" alt="How to unblock notifications"></a>`;
+        instructionsHtml = `<a href="${imageUrl}" target="_blank"><img src="${imageUrl}" alt="Unblock notifications"></a>`;
       }
     }
 
--- a/src/page/bell/Dialog.ts
+++ b/src/page/bell/Dialog.ts
@@ -89,7 +89,7 @@ export default class Dialog {
 
     const iconHtml =
       imageUrl !== 'default-icon'
-        ? `<div class="push-notification-icon"><img></div>`
+        ? `<div class="push-notification-icon"><img alt=""></div>`
         : `<div class="push-notification-icon push-notification-icon-default"></div>`;
 
     const isSubscribed = this._bell._state === BellState._Subscribed;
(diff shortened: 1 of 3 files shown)


src/page/bell/Dialog.ts
```diff
@@ -122,7 +122,7 @@ export default class Dialog {
       const imagePath = UNBLOCK_IMAGES[browserName];
       if (imagePath) {
         const imageUrl = STATIC_RESOURCES_URL + imagePath;
-        instructionsHtml = `<a href="${imageUrl}" target="_blank"><img src="${imageUrl}"></a>`;
+        instructionsHtml = `<a href="${imageUrl}" target="_blank"><img src="${imageUrl}" alt="How to unblock notifications"></a>`;
```
fadi-george: maybe can shorten to "Unblock notifications" to less bundle size slightly

fadi-george (approved):

src/page/bell/Dialog.ts
sherwinski: Done — shortened to "Unblock notifications" (also matches the dialog's blocked title).

PR #1482, opened 2026-07-30
fix: [SDK-4968] send the first on_session increment for newly created users

sherwinski:
# Description

## 1 Line Summary

Send the `session_count` delta on the first session for newly created users, so new web subscribers land on 1 session instead of 0.

## Details

The SDK deliberately skipped the first `on_session` increment when a session originated from a user create. That was correct while the backend seeded `session_count = 1` on create — the skip existed to avoid double counting.

That assumption stopped holding when [perseus#2023](https://github.com/OneSignal/perseus/pull/2023) shipped for UD-4311 and made `CreateSubscription` default `session_count` to `0`. With the skip still in place, new web subscribers got `0` and stayed there until their next visit.

The skip lived in two places, both keyed off `SessionOrigin._UserCreate`:

- `src/shared/helpers/service-worker.ts` — `sendOnSessionCallIfNotPlayerCreate()` returned early for the create origin. This is the main path (Chrome/Firefox/Safari with a service worker).
- `src/shared/managers/sessionManager/SessionManager.ts` — pre-set `_onSessionSent = true` on create, which made the page-side fallback (used when service workers are unavailable) bail out.

Both are removed, so the follow-up `PATCH /users/by/onesignal_id/{id}` now sends `{"deltas": {"session_count": 1}}` for new and returning users alike. `0 + 1 = 1`, matching what the Android and iOS SDKs already do.

The now-unused `sessionOrigin` parameter was dropped from `upsertSession()`. The page-to-service-worker message payload (`UpsertOrDeactivateSessionPayload`) is **unchanged**, so there is no wire-format break between an old service worker and a new page script.

# Systems Affected

- [x] WebSDK
- [ ] Backend
- [ ] Dashboard

# Validation

### Info

- Full suite: 528 tests across 60 files pass.
- `vp check`: 0 errors (5 pre-existing warnings in files untouched by this PR).
- `vp run build:prod`: passes, including size-limit checks. The SW bundle got marginally smaller since this removes code.
- Confirmed the added tests are genuine regression tests by stashing the source fix and re-running: both fail without it. The parameterized case passes for `_UserNewSession` even without the fix, which confirms the returning-user path was already correct and user-create was the only broken origin.

### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:

- [x] No Typescript warnings

Other:

### Info

No UI surface is affected — this changes the body of a network request only.

### Checklist

---

## Related Tickets

- SDK-4968
- UD-4311 (backend fix: perseus#2023, turbine#1762)

---

### DIFF
--- a/package.json
+++ b/package.json
@@ -77,12 +77,12 @@
     },
     {
       "path": "./build/releases/OneSignalSDK.page.es6.js",
-      "limit": "42.75 kB",
+      "limit": "42.7 kB",
       "gzip": true
     },
     {
       "path": "./build/releases/OneSignalSDK.sw.js",
-      "limit": "12.51 kB",
+      "limit": "12.5 kB",
       "gzip": true
     },
     {
--- a/src/shared/helpers/service-worker.ts
+++ b/src/shared/helpers/service-worker.ts
@@ -10,9 +10,9 @@ import Log from '../libraries/Log';
 import type { OutcomesNotificationClicked } from '../models/OutcomesNotificationEvents';
 import Path from '../models/Path';
 import type { OutcomesConfig } from '../outcomes/types';
-import { SessionOrigin, SessionStatus } from '../session/constants';
+import { SessionStatus } from '../session/constants';
 import { initializeNewSession } from '../session/helpers';
-import type { Session, SessionOriginValue } from '../session/types';
+import type { Session } from '../session/types';
 import { getBaseUrl } from './general';
 import { getConfigAttribution } from './OutcomesHelper';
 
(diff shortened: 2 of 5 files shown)


sherwinski:
@claude review once

fadi-george (approved):

fadi-george:
Id adjust size threshholds

sherwinski:
Good call — tightened in bcaa4ac6.

Measured the actual gzipped sizes on this branch against the base commit:

| Bundle | main | this PR | Δ | new limit |
| --- | --- | --- | --- | --- |
| `OneSignalSDK.sw.js` | 12,504 B | 12,475 B | −29 B | 12.51 kB → **12.5 kB** |
| `OneSignalSDK.page.es6.js` | 42,706 B | 42,697 B | −9 B | 42.75 kB → **42.7 kB** |
| `OneSignalSDK.page.js` | 405 B | 405 B | — | unchanged |
| `OneSignalSDK.page.styles.css` | 6,816 B | 6,816 B | — | unchanged |

Rounded each down to the next 0.05 kB boundary to match the convention in the existing history, and left the two bundles this PR doesn't touch alone. `build:prod` passes at the new limits.

PR #1484, opened 2026-08-26
fix: [SDK-5014] omit local subscription IDs from the create user request

sherwinski:
# Description

## 1 Line Summary

Stop the SDK from sending `local-*` placeholder subscription IDs in the Create User request, which the backend rejects as invalid UUIDs.

## Details

The `login()` and `logout()` flows enqueue a `TransferSubscriptionOperation` with the current push subscription model ID. That ID can still be a `local-*` placeholder when the create-subscription operation did not complete first (for example, after a network failure dropped the pending operations). `_canStartExecute` blocks such a transfer op from starting on its own, but `OperationRepo._getGroupableOperations` does not check `_canStartExecute` for grouped operations. A `LoginUserOperation` with the same comparison key groups the transfer op, and the transfer branch in `LoginUserOperationExecutor` copied the local ID into the payload as `{"id": "local-..."}`. The backend rejects the request with `uuid: incorrect UUID length 42`.

The failure then persists: the 400 maps to `FailPauseOpRepo`, so the operations re-queue, persist in IndexedDB, and fail again on every page load. Affected users stay unsubscribed until they clear site data.

The fix changes the transfer branch for a local subscription ID. A transfer grouped with a create op keeps the create data and adds no `id`. A lone transfer op rebuilds the subscription data from the model in `SubscriptionModelStore`, again with no `id`, so the backend creates the subscription and ID translation still runs. A model with no token is skipped, because the store listener never sends a token-less subscription as enabled and the model self-heals on the next subscribe. A create user request that ends with no identity and no subscriptions is dropped (`FailNoretry`) instead of sent, which matches the Android SDK. Users who are stuck today recover on their next page load because the persisted login operation now sends a valid payload.

The size limit for `OneSignalSDK.page.es6.js` moves from 42.7 kB to 42.78 kB; the new branch pushed the gzip size 67 bytes over the old limit.

The manual repro harness used during review is not part of the final diff. Commit `9a0f42e` on this branch has the last version of it, and the PR comment "How to test this manually" has the steps.

# Systems Affected

- [x] WebSDK
- [ ] Backend
- [ ] Dashboard

# Validation

### Info

5 new unit tests in `LoginUserOperationExecutor.test.ts`. Before the fix, the first test reproduced the exact invalid payload from the customer report (`subscriptions: [{"id": "local-..."}]`). The other tests cover the create-plus-transfer group, the rebuild from the model on logout, the token-less model, and the dropped empty payload. All pass after the fix, along with the full suite, `vp check`, and `vp run build:prod`.

### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:

- [x] No Typescript warnings

Other:

### Info

Not needed: the change affects a network payload with no UI impact. The unit tests assert the payload contents.

### Checklist

---

## Related Tickets

[SDK-5014](https://linear.app/onesignal/issue/SDK-5014/web-sdk-v16-sends-local-prefixed-placeholder-subscription-id-in-create)

---

### DIFF
--- a/src/core/executors/LoginUserOperationExecutor.ts
+++ b/src/core/executors/LoginUserOperationExecutor.ts
@@ -3,6 +3,7 @@ import { ExecutionResult, type IOperationExecutor } from 'src/core/types/operati
 import { getResponseStatusType, ResponseStatusType } from 'src/shared/helpers/network';
 import Log from 'src/shared/libraries/Log';
 import { checkAndTriggerUserChanged } from 'src/shared/listeners';
+import { IDManager } from 'src/shared/managers/IDManager';
 
 import { IdentityConstants, OPERATION_NAME } from '../constants';
 import { type IPropertiesModelKeys } from '../models/PropertiesModel';
@@ -289,6 +290,11 @@ export class LoginUserOperationExecutor implements IOperationExecutor {
 
       case operation instanceof TransferSubscriptionOperation: {
         const subscriptionId = operation._subscriptionId;
+
+        // A local ID means the subscription does not exist on the backend yet,
+        // so there is nothing to transfer and the backend rejects local IDs as
+        // invalid UUIDs. A grouped create operation still creates it.
+        if (IDManager._isLocalId(subscriptionId)) return currentSubs;
 
         return {
           ...currentSubs,
--- a/src/core/executors/LoginUserOperationExecutor.ts
+++ b/src/core/executors/LoginUserOperationExecutor.ts
@@ -308,7 +308,7 @@ export class LoginUserOperationExecutor implements IOperationExecutor {
           if (currentSubs[subscriptionId]) return currentSubs;
 
           const model = this._subscriptionsModelStore._getBySubscriptionId(subscriptionId);
-          if (!model) return currentSubs;
+          if (!model?.token) return currentSubs;
 
           return {
             ...currentSubs,
--- a/src/core/executors/LoginUserOperationExecutor.ts
+++ b/src/core/executors/LoginUserOperationExecutor.ts
@@ -149,6 +149,15 @@ export class LoginUserOperationExecutor implements IOperationExecutor {
     }
 
     const subscriptionList = Object.entries(subscriptions);
+
+    // The backend rejects a create user request that has no identity and no
+    // subscriptions, so drop the operation instead of a request that always
+    // fails. The Android SDK has the same guard.
+    if (Object.keys(identity).length === 0 && subscriptionList.length === 0) {
+      Log._debug('CreateUser: skipped, no identity and no subscriptions');
+      return { _result: ExecutionResult._FailNoretry };
+    }
+
     const response = await createNewUser(
       { appId: createUserOperation._appId },
       {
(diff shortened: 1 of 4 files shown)


sherwinski:
@claude review once

sherwinski:
## How to test this manually (updated for the logout path)

The branch includes a repro harness in `index.html` (dev server page). It has 4 ordered repro buttons, a state card that shows whether the push subscription ID is a `local-*` placeholder, a reset button, and a log that prints every Create User request and response.

**Background:** the bug needs a push subscription whose ID is still a `local-*` placeholder, with no pending create operation left to translate it. In production this happens when the SDK drops the queued operations (see the PR description). The "Seed stuck local ID" button creates the same state instantly: it rewrites the subscription model ID with the `NoPropogate` tag, so no operation is enqueued. The logout step then exercises Create User deterministically — a login after the seed can take the Set Alias path instead, because the OneSignal ID is already server-assigned.

### Setup

1. Run `vp install`, then `vp run dev`.
2. Open the printed localhost URL with `?app_id=<your test app id>` (or set `VITE_APP_ID` in `.env`). Use a test app that allows the localhost origin.

### See the bug (without the fix)

1. From `main`, copy only the harness: `git checkout main && git checkout sherwin/sdk-5014 -- index.html`.
2. Click **Reset SDK state**.
3. Click **1. Subscribe** and grant permission. Wait for the green "server-assigned" tag.
4. Click **2. Login**.
5. Click **3. Seed stuck local ID**. The tag turns red.
6. Click **4. Logout**. The log shows the Create User request with `"subscriptions": [{"id": "local-..."}]` and the 400 response: `uuid: incorrect UUID length 42`.
7. Reload the page. The same 400 repeats with no clicks, because the failed operations persist in IndexedDB.

### Confirm the fix

1. Check out this branch (discard the copied `index.html` first: `git checkout -- index.html`).
2. Repeat steps 2–6. At step 6 the request carries the full subscription data (token, type, and so on) with no `id`, and the response is 200. The state card then shows a new server-assigned subscription ID for the anonymous user.
3. Recovery check: reproduce the stuck loop on `main`, then switch to this branch and reload without a reset. The persisted operations now send a valid payload and succeed.

Note: an anonymous create user request with no identity and no subscriptions (the `user-3` case from the earlier review) no longer goes to the network; the SDK drops the operation instead.

fadi-george:
maybe explore a change like this in LoginManger:
```
if (pushOp) {
  const operation = IDManager._isLocalId(pushOp.id)
    ? new CreateSubscriptionOperation({
        appId,
        onesignalId: newOneSignalId,
        subscriptionId: pushOp.id,
        type: pushOp.type,
        enabled: pushOp.enabled,
        token: pushOp.token,
        notification_types: pushOp._notification_types,
      })
    : new TransferSubscriptionOperation(appId, newOneSignalId, pushOp.id);
  OneSignal._coreDirector._operationRepo._enqueue(operation);
}
```

dombartenope:
@fadi-george @sherwinski is this something going live in an upcoming update?

fadi-george:
I did this check for logout:
1. Reset SDK state and subscribe; wait for a real subscription ID.
2. Log in first:
```js
await OneSignal.login('logout-repro-user');
```
3. Seed a local ID without creating an operation:
```js
const model = await OneSignal._coreDirector._getPushSubscriptionModel();
model._setProperty('id', `local-${crypto.randomUUID()}`, 1);
```
4. Log out:
```js
await OneSignal.logout();
```
5. Inspect `POST /apps/{appId}/users`. It sends:
```json
{ "identity": {}, "subscriptions": [] }
```
6. The backend returns `400 user-3`. 

But at least reloading recovers from it.

src/core/executors/LoginUserOperationExecutor.ts
```diff
@@ -290,6 +300,32 @@ export class LoginUserOperationExecutor implements IOperationExecutor {
...
 
+        // A local ID means the subscription does not exist on the backend yet,
+        // so there is nothing to transfer and the backend rejects local IDs as
+        // invalid UUIDs. Send the subscription data from its model instead so
+        // the backend creates it. A grouped create operation already has it.
+        if (IDManager._isLocalId(subscriptionId)) {
+          if (currentSubs[subscriptionId]) return currentSubs;
+
+          const model = this._subscriptionsModelStore._getBySubscriptionId(subscriptionId);
+          if (!model) return currentSubs;
```
fadi-george: for extra safety we could 
`if (!model?.token) return currentSubs;`

fadi-george (approved):

src/core/executors/LoginUserOperationExecutor.ts
sherwinski: Applied in bb471b98 with a test. A model without a token would go out as `enabled: true, token: ''`, which the store listener never sends (it normalizes a token-less model to `enabled: false`). The model self-heals on the next subscribe through `createSubscribedUser`, so there is no loss.

sherwinski:
Hey @dombartenope, this is now live in [v160610](https://github.com/OneSignal/OneSignal-Website-SDK/releases/tag/160610) 🚀

PR #1488, opened 2026-09-15
feat: [SDK-5223] add the identity verification two-gate model

sherwinski:
# Description

## 1 Line Summary

Add the Identity Verification two-gate model as 2 functions.

## Details

Third M1 pull request for [Identity Verification for Web SDK](https://linear.app/onesignal/project/identity-verification-for-web-sdk-800ed255b785). Stacked on #1487 (SDK-5224), which is stacked on #1486 (SDK-5222). Targets `sherwin/sdk-5224` now; retarget to `feat/identity-verification` after the 2 earlier PRs merge.

Mirrors the behavior of Android `core/internal/config/impl/IdentityVerificationService.kt`. The shape follows web conventions: 2 exported functions, no class, no interface, no DI. This is the pattern for the rest of the IV project: behavioral parity with Android, TypeScript idioms for the implementation.

- New `src/core/identityVerification.ts`:
  - `isIvBehaviorActive() = requirement === required`. Gates behavior: bearer, alias switch, unauthorized handling. Android: `ivBehaviorActive`.
  - `isIvCodePathEnabled() = flag(sdk_identity_verification) || isIvBehaviorActive()`. Gates whether the new code runs at all. Android: `newCodePathsRun`.
  - Both gates are derived on read from the persisted requirement (SDK-5222) and the cached flag (SDK-5224). Nothing is cached. `UNKNOWN` reads as false for both gates when the flag is off.
- No class, interface, or `CoreModule` wiring. Call sites import the 2 functions. Tests drive the gates through `setJwtRequirement` and `setFeatureFlags`, which only produces states production can reach.
- The hydrated handler was removed. On web the requirement is settled before `CoreModule` exists, so SDK-5230 will read the gate directly in `OperationRepo._start` after saved operations load.
- The test environment (`initOSGlobals`) persists the requirement the way `getAppConfig` does, so `overrideServerConfig: { config: { jwt_required: true } }` reaches the gates.

Design rule for later call sites: gate on `isIvCodePathEnabled()` at the call site, and gate on `isIvBehaviorActive()` inside the IV helper. Phase 3 (flag on, requirement off) must produce requests identical to the legacy path.

No behavior change for customers. Nothing reads the gates yet.

**Bundle size.** No change. Nothing in production imports the module yet, so the bundler drops it. The first consumer (SDK-5230) will pull in the module plus the requirement and flag readers from the 2 earlier PRs, about 190 B. SDK-5252 (M6) tracks the budget for the whole IV project.

# Systems Affected

- [x] WebSDK
- [ ] Backend
- [ ] Dashboard

# Validation

### Info

New `src/core/identityVerification.test.ts`, 9 cases:

- The 6-row truth table from `IdentityVerificationServiceTests.kt:50-84`, plus derived-on-read and the local override.
- The test environment persists the requirement from the fake server config, the same way `getAppConfig` does, so the gates follow `overrideServerConfig`.

`vp check`, `vp test` (571 tests), and `vp run build:prod` pass.

### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:

- [x] No Typescript warnings

Other:

### Info

Not needed. No UI change.

### Checklist

---

## Related Tickets

- [SDK-5223](https://linear.app/onesignal/issue/SDK-5223)
- Design: [Web SDK Identity Verification design](https://linear.app/onesignal/document/web-sdk-identity-verification-design-android-rules-to-web-components-b6aa0087fabf), section 2

---


src/core/IdentityVerificationService.ts
```diff
@@ -0,0 +1,29 @@
...
+/**
+ * Single source of truth for Identity Verification gates. Both gates are
+ * derived on read from the persisted requirement and the feature flag, so a
+ * change in either source shows up on the next read. UNKNOWN reads as false.
+ *
+ * Call sites gate on `_newCodePathsRun`; the IV helpers gate on
+ * `_ivBehaviorActive` inside. With the flag on and the requirement off, the
+ * new code runs but must produce requests identical to the legacy path.
+ */
+export class IdentityVerificationService implements IIdentityVerificationService {
```
fadi-george: i think we should avoid unnecessary classes, these could just be export functions

src/core/IdentityVerificationService.ts
```diff
@@ -0,0 +1,29 @@
...
+  get _ivBehaviorActive(): boolean {
```
fadi-george: i feel boolean functions should start something like
isX or hasY or shouldZ
so maybe rename to shouldUseJWT or isJWTActive

src/core/IdentityVerificationService.ts
```diff
@@ -0,0 +1,29 @@
...
+    return getJwtRequirement() === JwtRequirement._Required;
+  }
+
+  get _newCodePathsRun(): boolean {
```
fadi-george: not a big fan of this name, doesnt seem explain what it does.
maybe should be renamed "shouldUseIdentityVerification"

src/core/IdentityVerificationService.ts
sherwinski: Done in 0a33a5c0. The class, the `IIdentityVerificationService` interface, the `CoreModule` field, the director getter, and the test mock are gone. `src/core/identityVerification.ts` exports 2 functions. Tests drive the gates through `setJwtRequirement` and `setFeatureFlags`, which only produces states production can reach. This is the pattern for the rest of the IV project: behavioral parity with Android, TypeScript conventions for the shape.

src/core/IdentityVerificationService.ts
sherwinski: Renamed to `isIvBehaviorActive()`. I kept the "IV behavior" noun because the design doc and the M2 to M5 tickets use that term, so it stays greppable across both SDKs.

src/core/IdentityVerificationService.ts
sherwinski: Renamed to `isIvCodePathEnabled()`. I did not take `shouldUseIdentityVerification` because it reads like the other gate. In Phase 3 (flag on, requirement off) this gate returns `true` while the SDK must send requests identical to legacy, so a name that says "use IV" would invite gating the bearer on it. "Code path" keeps the distinction: call sites gate on the code path, IV helpers gate on `isIvBehaviorActive()` inside.

fadi-george (approved):

PR #1489, opened 2026-09-15
feat: [SDK-5225] add a multi-user JWT token store and remove the dead `jwtToken` row

sherwinski:
# Description

## 1 Line Summary

Add a multi-user `JwtTokenStore` in localStorage and remove the dead single-slot `jwtToken` IndexedDB row.

## Details

Fourth and last M1 pull request for [Identity Verification for Web SDK](https://linear.app/onesignal/project/identity-verification-for-web-sdk-800ed255b785). Covers SDK-5225 and SDK-5226 in one change, as SDK-5226 asks. Stacked on #1488 (SDK-5223). Targets `sherwin/sdk-5223` now; retarget to `feat/identity-verification` after the earlier PRs merge.

Follows decision D3 in the design doc: localStorage, one JSON map under one key. Mirrors Android `user/internal/jwt/JwtTokenStore.kt`.

**`src/core/JwtTokenStore.ts`**

- In-memory map `externalId -> jwt`, lazy loaded from localStorage key `os_jwt_tokens`. Persisted as one JSON object. Malformed JSON logs a warning, deletes the value, and starts empty. The in-memory map is a `Map`, so externalIds such as `constructor` cannot collide with `Object.prototype`. A failed localStorage write logs a warning and keeps the in-memory change, so `login()` does not reject and the token still serves this session. Non-object values and non-string entries are dropped on load.
- `_getJwt(externalId)` returns `undefined` when absent. Android returns `null`; `undefined` matches the map lookup on web and avoids a conversion.
- `_putJwt(externalId, jwt)`: a missing token is a no-op. Persists and fires the internal update listener only on change.
- `_invalidateJwt(externalId)`: removes, persists, then fires the developer-facing invalidated listener with `{ externalId }`, only if an entry existed. Does not fire the internal update listener. Each listener runs in its own try/catch, so one throwing listener cannot break the others or reach the operation queue. The store updates its state before it fires, so a listener that calls `_putJwt` during the event sees the removal first.
- `_pruneToExternalIds(activeIds)`: drops tokens for users not in the set and fires the internal listener for each removed id. No production caller yet; the prune-on-init caller needs operations to carry `externalId` (M2, SDK-5227).
- Two listener audiences, like Android: `_addUpdateListener` and `_addUserJwtInvalidatedListener`, each with a remove. No replay for late subscribers.
- Storage is unconditional. Use of the tokens is gated by `IdentityVerificationService` in M2 and M3.
- Listeners fire synchronously. Android dispatches the public event on a background thread because its operation queue runs on a dedicated thread; web has one thread and the per-listener try/catch gives the same isolation.

**Wiring**

- `CoreModule` constructs the store. `CoreModuleDirector._jwtTokenStore` exposes it.
- `LoginManager._login` writes through `_putJwt(externalId, token)` before the "externalId already set" check, so `login(sameId, newToken)` stores the new token. The "wake the queue" half of that behavior is SDK-5240 (M4).

**SDK-5226: dead row removal**

- Removed the fire-and-forget `db.put('Ids', { id: token, type: 'jwtToken' })`. Nothing read it.
- Removed `'jwtToken'` from the `IdKey` union.
- No one-time cleanup of a stale `jwtToken` row. The ticket rates it low value, and the row is harmless.

Not in this PR: the cross-tab `storage` listener (D6) is SDK-5248 in M5. The public `userJwtInvalidated` event on `OneSignal.User` is SDK-5241 in M4; it will subscribe to this store.

**Bundle size.** Adds about 390 B to `page.es6.js`. The page limit moves to the measured 43.44 kB, per the repo convention. Across the 4 M1 PRs the page bundle grows from 42.78 kB to 43.44 kB (about 660 B); the two-gate module in #1488 adds nothing until a production caller imports it. SDK-5252 (M6) tracks the budget for the whole IV project.

# Systems Affected

- [x] WebSDK
- [ ] Backend
- [ ] Dashboard

# Validation

### Info

New `src/core/JwtTokenStore.test.ts`, 32 cases, mirrors `JwtTokenStoreTests.kt:18-293`:

- get missing, put and replace, put null or undefined no-op, invalidate, invalidate absent.
- Persistence: round trip through a fresh instance for put and invalidate, exact persisted shape, malformed JSON recovery with a warning and deletion, non-object and non-string entries dropped, a failed write keeps the token in memory.
- Prototype keys: `constructor`, `toString`, `__proto__`, and `hasOwnProperty` behave like any other externalId.
- Prune: removes inactive, fires per removed id, no-op when nothing to remove.
- Internal listener: fires on change only, not on invalidate, not after unsubscribe.
- Public listener: fires with `externalId`, not for an absent token, throwing subscriber isolated, late subscriber gets nothing, removed listener stops, a listener that removes itself does not skip the next one, re-entrant put sees the removal first.
- The director exposes the store.

`LoginManager.test.ts`: the token test now asserts the store call and the stored value. New case: `login(sameId, newToken)` stores the token.

`vp check`, `vp test` (607 tests), and `vp run build:prod` pass.

### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:

- [x] No Typescript warnings

Other:

### Info

Not needed. No UI change.

### Checklist

---

## Related Tickets

- [SDK-5225](https://linear.app/onesignal/issue/SDK-5225/add-a-multi-user-jwt-token-store)
- [SDK-5226](https://linear.app/onesignal/issue/SDK-5226/remove-the-dead-single-slot-jwttoken-row-and-idkey-entry)
- Decision: [SDK-5214](https://linear.app/onesignal/issue/SDK-5214/decide-jwt-token-storage-localstorage-or-indexeddb)
- Design: [Web SDK Identity Verification design](https://linear.app/onesignal/document/web-sdk-identity-verification-design-android-rules-to-web-components-b6aa0087fabf), D3

---


src/shared/helpers/localStorage.ts
```diff
@@ -57,6 +59,27 @@ export function getJwtRequirement(): JwtRequirementValue {
...
+  const json = localStorage.getItem(JWT_TOKENS);
+  if (json === null) return {};
+  try {
+    const parsed: unknown = JSON.parse(json);
+    if (typeof parsed !== 'object' || parsed === null || Array.isArray(parsed)) return {};
+    return Object.fromEntries(
+      Object.entries(parsed).filter(([, v]) => typeof v === 'string'),
+    ) as Record<string, string>;
+  } catch {
+    Log._warn('JwtTokenStore: failed to parse persisted tokens, starting fresh');
```
fadi-george, line 78: maybe delete localStorage JWT_TOKENS if it gets corrupted and fails to read?

src/page/managers/LoginManager.ts
```diff
@@ -21,9 +20,7 @@ export default class LoginManager {
   }
```
fadi-george, line 21: maybe clear the jwt on logout or is that a follow up?

src/core/JwtTokenStore.ts
```diff
@@ -0,0 +1,87 @@
...
+
+  _addInternalUpdateListener(listener: JwtUpdatedListener): void {
+    this._updateListeners._subscribe(listener);
+  }
+
+  _removeInternalUpdateListener(listener: JwtUpdatedListener): void {
+    this._updateListeners._unsubscribe(listener);
+  }
+
+  _addUserJwtInvalidatedListener(listener: UserJwtInvalidatedListener): void {
```
fadi-george, line 34: do we really want to allow multiple subscribe listenrs?

fadi-george: i guess its fine

src/core/JwtTokenStore.ts
```diff
@@ -0,0 +1,87 @@
...
+ * prune. The public invalidated listener fires on invalidate only, and only
+ * for subscribers present at that time. Logout and user switch must not call
+ * invalidate; the developer would read that as "refresh your token".
+ */
+export class JwtTokenStore {
+  private _tokens?: Record<string, string>;
+  private _updateListeners = new EventProducer<JwtUpdatedListener>();
+  private _invalidatedListeners = new EventProducer<UserJwtInvalidatedListener>();
+
+  _addInternalUpdateListener(listener: JwtUpdatedListener): void {
```
fadi-george, line 26: dont need to say "internal"? anything with `_` would be obfuscated anyway

src/shared/helpers/localStorage.ts
sherwinski: Done in 536dfe02. `getJwtTokens` now removes the value when the parse fails, so the warning fires once instead of on every load.

src/page/managers/LoginManager.ts
sherwinski: Deferred on purpose. The store is multi-user so operations still queued under the previous user can find their token after a switch (design D3). Logout also must not fire `userJwtInvalidated`; a developer would read that as "refresh your token". Cleanup happens on cold start through `_pruneToExternalIds`, once operations carry `externalId` (SDK-5227, M2).

src/core/JwtTokenStore.ts
sherwinski: Yes. In M4 the public `userJwtInvalidated` event on `OneSignal.User` forwards from this store, and developers can attach more than one handler there. On the SDK side the operation repo subscribes to the update listener in M2, and the cross-tab sync in M5 may add a second subscriber.

src/core/JwtTokenStore.ts
sherwinski: Renamed to `_addUpdateListener` / `_removeUpdateListener` in 536dfe02.

src/core/JwtTokenStore.ts
```diff
@@ -0,0 +1,87 @@
...
+  _addUserJwtInvalidatedListener(listener: UserJwtInvalidatedListener): void {
+    this._invalidatedListeners._subscribe(listener);
+  }
+
+  _removeUserJwtInvalidatedListener(listener: UserJwtInvalidatedListener): void {
+    this._invalidatedListeners._unsubscribe(listener);
+  }
+
+  _getJwt(externalId: string): string | undefined {
+    return this._load()[externalId];
```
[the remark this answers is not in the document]
sherwinski: Fixed in 4b8c22f: the in-memory map is a `Map`, and the persisted object is read with `Object.entries`, which returns own keys only. Tests cover `constructor`, `toString`, `__proto__`, and `hasOwnProperty`.

src/core/JwtTokenStore.ts
```diff
@@ -0,0 +1,87 @@
...
+  }
+
+  // A missing token is a no-op. Use _invalidateJwt to remove one.
+  _putJwt(externalId: string, jwt: string | null | undefined): void {
+    if (!jwt) return;
+    const tokens = this._load();
+    if (tokens[externalId] === jwt) return;
+    tokens[externalId] = jwt;
+    setJwtTokens(tokens);
```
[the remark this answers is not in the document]
sherwinski: Addressed in 536dfe02 with a different choice. `_persist` catches the write failure and logs it, and the in-memory map keeps the change. A rollback would drop a token the session can still use; keeping it means M3 can still attach the bearer for this page load, and the next load reads whatever the disk has. `login()` no longer rejects on a storage failure.

src/page/managers/LoginManager.ts
```diff
@@ -21,9 +20,7 @@ export default class LoginManager {
...
 
   private static async _login(externalId: string, token?: string): Promise<void> {
-    if (token) {
-      void db.put('Ids', { id: token, type: 'jwtToken' });
-    }
+    OneSignal._coreDirector._jwtTokenStore._putJwt(externalId, token);
```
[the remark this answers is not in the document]
sherwinski: The `setItem` half is covered by the `_persist` try/catch in 536dfe02. I am not wrapping the update listeners: they are SDK-owned (the operation repo in M2), so a throw there is an SDK bug that tests should surface, not a condition to hide. The public invalidated listener keeps its per-listener try/catch because that code is developer-owned. If M2 shows a reason to isolate the internal path, that PR can add it.

sherwinski:
Follow-up on the multi-model summary items without a thread:

- `EventProducer._fire` now iterates a copy (4b8c22f), with a test in `EventProducer.test.ts` for a listener that removes itself.
- App-id change in `initSaveState` and `os_jwt_tokens`: deferred. A JWT is signed for one app, so a token reused across apps is rejected by the server and takes the normal invalidate path in M3. The scenario needs 2 apps on one origin with the same `externalId`, which is a development-time case. If we want a purge there, it belongs with the cold-start prune in SDK-5227 rather than in this PR.

fadi-george (approved):

PR #1492, opened 2026-09-17
feat: [SDK-5227] add `externalId` and `requiresJwt` to every operation

sherwinski:
# Description

## 1 Line Summary

Add `externalId` and `_requiresJwt` to `Operation`, and pass the owner's `externalId` through every operation constructor and enqueue site.

## Details

First of 7 M2 pull requests for [Identity Verification for Web SDK](https://linear.app/onesignal/project/identity-verification-for-web-sdk-800ed255b785). M2 is the operation queue runtime. Targets `feat/identity-verification`, which now contains the 4 merged M1 PRs.

Recommended review order for the M2 stack: SDK-5227 → SDK-5228 → SDK-5232 → SDK-5233 → SDK-5231 → SDK-5229 → SDK-5230. Each PR is stacked on the one before it.

Mirrors Android `core/internal/operations/Operation.kt`, where every operation carries the `externalId` of the user it was enqueued for.

**`src/core/operations/Operation.ts`**

- `externalId?: string` joins the base model. The base constructor takes it as an optional trailing parameter.
- `_externalId` getter and protected setter. The value is captured at construction, so the operation stays bound to its original user if the current user changes before it runs.
- `_requiresJwt` getter returns `true`. Subclasses can override to `false` for endpoints that do not check auth. Android exposes the same computed property and does not persist it; web does the same.
- An anonymous operation has no `externalId` key, so its persisted JSON is unchanged. A row persisted before this change rehydrates as anonymous.

**Operation constructors**

Every concrete operation accepts `externalId` and passes it to its base. Positional constructors take it as a trailing parameter; object-style constructors (`CreateSubscriptionOperation`, `UpdateSubscriptionOperation`, `TrackCustomEventOperation`) read it from the argument object. `LoginUserOperation` and `TrackCustomEventOperation` already stored `externalId` in their own type; they now use the base field.

**Enqueue sites**

- `IdentityModelStoreListener` passes `model._externalId` to alias operations.
- `PropertiesModelStoreListener` now takes the `IdentityModelStore` to read the owner, like Android.
- `SubscriptionModelStoreListener` passes the owner to create, update, and delete subscription operations.
- `LoginManager._switchUser`, the push grant path in `subscription/page.ts`, `RebuildUserService`, and the follow-up operations in `LoginUserOperationExecutor` and `SubscriptionOperationExecutor` pass the owner through.
- `CoreModule` wires the new listener dependency.

Nothing reads either field yet. The dispatch gate (SDK-5232), the suppression (SDK-5229), and the purge (SDK-5230) consume `_externalId`; the gate consumes `_requiresJwt`.

**Bundle size.** Adds 73 B gzipped to `page.es6.js`. The page limit moves from 43.44 kB to the measured 43.52 kB, per the repo convention.

# Systems Affected

- [x] WebSDK
- [ ] Backend
- [ ] Dashboard

# Validation

### Info

New `src/core/operations/Operation.test.ts`:

- an identified operation carries `externalId` and serializes it.
- an anonymous operation has no `externalId` key in its JSON.
- `_requiresJwt` defaults to `true`.
- `externalId` round-trips through `OperationModelStore._create`.
- a legacy row with no `externalId` rehydrates as anonymous.

New `src/core/listeners/listeners.test.ts`:

- with an identified user, alias, property, and subscription operations carry the user's `externalId`.
- with an anonymous user, the same operations carry `undefined`.

`LoginUserOperationExecutor.test.ts` updated for the new constructor shape.

### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:

- [x] No Typescript warnings

Other:

### Info

Not needed. No UI change.

### Checklist

---

## Related Tickets

- [SDK-5227](https://linear.app/onesignal/issue/SDK-5227/add-externalid-and-requiresjwt-to-operation-and-every-operation-type)


src/core/executors/LoginUserOperationExecutor.ts
```diff
@@ -76,6 +76,7 @@ export class LoginUserOperationExecutor implements IOperationExecutor {
         loginUserOp._existingOnesignalId,
         IdentityConstants._ExternalID,
         loginUserOp._externalId,
+        loginUserOp._externalId,
```
fadi-george, line 79: this looks really strange

src/core/listeners/IdentityModelStoreListener.ts
```diff
@@ -29,9 +29,15 @@ export class IdentityModelStoreListener extends SingletonModelStoreListener<Iden
   ): Operation {
     const appId = getAppId();
     if (newValue != null && typeof newValue === 'string') {
-      return new SetAliasOperation(appId, model._onesignalId, property, newValue);
+      return new SetAliasOperation(
```
fadi-george, line 32: are we mimicking android sdk?

fadi-george, line 32: with more than 4 params I would say just pass an object at that point

src/core/operations/BaseAliasOperation.ts
```diff
@@ -11,8 +11,14 @@ export abstract class BaseAliasOperation<
   U extends object = AliasOperation,
   T extends U & AliasOperation = U & AliasOperation,
 > extends Operation<T> {
-  constructor(operationName: string, appId?: string, onesignalId?: string, label?: string) {
-    super(operationName, appId, onesignalId);
+  constructor(
```
fadi-george, line 14: with more than 4 params I would say just pass an object at that point

src/core/operations/BaseFullSubscriptionOperation.ts
fadi-george, line 22: with more than 4 params I would say just pass an object at that point

src/core/operations/BaseTagOperation.ts
```diff
@@ -11,8 +11,14 @@ export abstract class BaseTagOperation<
   U extends object = TagOperation,
   T extends U & TagOperation = U & TagOperation,
 > extends Operation<T> {
-  constructor(operationName: string, appId?: string, onesignalId?: string, key?: string) {
-    super(operationName, appId, onesignalId);
+  constructor(
```
fadi-george, line 14: with more than 4 params I would say just pass an object at that point

src/core/operations/BaseSubscriptionOperation.ts
fadi-george, line 11: with more than 4 params I would say just pass an object at that point

src/core/operations/SetAliasOperation.ts
```diff
@@ -9,9 +9,21 @@ type AliasOp = {
 // Reference: https://github.com/OneSignal/OneSignal-Android-SDK/blob/5.1.31/OneSignalSDK/onesignal/core/src/main/java/com/onesignal/user/internal/operations/SetAliasOperation.kt
 export class SetAliasOperation extends BaseAliasOperation<AliasOp> {
```
fadi-george, line 10: with more than 4 params I would say just pass an object at that point

src/core/executors/LoginUserOperationExecutor.ts
sherwinski: Agreed. If we refactor to pass an object, we can specifically label the first as `value` and the second as something like `owner`

src/core/listeners/IdentityModelStoreListener.ts
sherwinski: yes, Android also passes `externalId` on all operations for JWT

sherwinski: Agreed, I'll handle that in a separate PR.

src/core/operations/Operation.ts
```diff
@@ -14,17 +14,19 @@ type BaseOperation = {
...
   B extends object = BaseOperation,
   T extends B & BaseOperation = B & BaseOperation,
 > extends Model<T> {
-  constructor(name: string, appId?: string, onesignalId?: string) {
+  constructor(name: string, appId?: string, onesignalId?: string, externalId?: string) {
     super();
     this._name = name;
     if (appId) this._appId = appId;
     if (onesignalId) this._onesignalId = onesignalId;
+    if (externalId) this._externalId = externalId;
```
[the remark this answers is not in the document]
sherwinski: Address in [b0e1f85](https://github.com/OneSignal/OneSignal-Website-SDK/pull/1492/commits/b0e1f85e07a284e0a21d68757176d22afd79e8bc) to reject on empty string.

src/core/executors/LoginUserOperationExecutor.ts
fadi-george, line 74: make SetAliasOperation take in an object?

src/onesignal/OneSignal.ts
```diff
@@ -108,6 +108,10 @@ export default class OneSignal {
       throw WrongTypeArgumentError('externalId');
     }
 
+    if (externalId === '') {
```
fadi-george: could combine with line 103 i.e. if (!externalId)

src/core/operations/SetAliasOperation.ts
fadi-george: whyd you resolve this

sherwinski: See [this comment](https://github.com/OneSignal/OneSignal-Website-SDK/pull/1492#discussion_r4065870762)

src/core/executors/LoginUserOperationExecutor.ts
sherwinski: Will handle this in the next PR, same as [this comment](https://github.com/OneSignal/OneSignal-Website-SDK/pull/1492#discussion_r4065870762).

src/core/operations/Operation.ts
```diff
@@ -51,6 +53,26 @@ export abstract class Operation<
...
+  protected set _externalId(value: string | undefined) {
+    this._setProperty('externalId', value);
+  }
+
+  /**
+   * Whether this operation needs a valid JWT when Identity Verification is active.
+   * Subclasses may override to false for endpoints that do not check auth.
+   */
+  get _requiresJwt(): boolean {
+    return true;
```
fadi-george, line 73: why is this set to true

fadi-george (approved):

src/core/operations/Operation.ts
sherwinski: This was a direct port of the Android implementation. The base class always returns true so that subclasses can  override it for specific endpoints that don't support JWT -- which does not apply to any endpoint at this time.

PR #1495, opened 2026-09-17
feat: [SDK-5233] make `_enqueueAndWait` reject with an error value

sherwinski:
# Description

## 1 Line Summary

Make `OperationRepo._enqueueAndWait` reject with an `OperationFailedError` that carries a typed reason.

## Details

Fourth of 7 M2 pull requests for [Identity Verification for Web SDK](https://linear.app/onesignal/project/identity-verification-for-web-sdk-800ed255b785). Stacked on #1494 (SDK-5232). Targets `sherwin/sdk-5232` now; retarget to `feat/identity-verification` after the earlier PRs merge.

`_enqueueAndWait` rejected with `undefined`, so a caller could not tell why the operation did not complete. The 401 handler (SDK-5231) and the enqueue suppression (SDK-5229) need to tell the caller which case applied, so this PR adds the error value first.

**`src/core/types/operation.ts`**

- New `OperationFailureReason` constant with 5 values: `unauthorized`, `suppressed`, `dropped`, `conflict`, `paused`. `OperationFailureReasonValue` is the union type.

**`src/shared/errors/common.ts`**

- New `OperationFailedError extends Error` with a `reason: OperationFailureReasonValue` field.

**`src/core/operationRepo/OperationRepo.ts`**

- `OperationQueueItem.resolver` now has the signature `(value: boolean, reason?: OperationFailureReasonValue) => void`. A `false` value must carry the reason.
- `_enqueueAndWait` wraps the resolver: `true` resolves; `false` rejects with `new OperationFailedError(reason)`. A missing reason defaults to `dropped`.
- Every failure site in the queue passes a reason: `FailNoretry` and the executor-throws path pass `dropped`, `FailConflict` passes `conflict`, `FailPauseOpRepo` passes `paused`, and `FailUnauthorized` passes `unauthorized`. A `failureReason(result)` helper maps the execution result to the reason.

No production caller reads the rejection value in this PR, so behavior is unchanged for callers that only `await` or `catch`. `unauthorized` and `suppressed` get their IV-specific paths in SDK-5231 and SDK-5229.

**Bundle size.** Adds about 80 B gzipped to `page.es6.js`. The page limit moves to the measured 43.87 kB.

# Systems Affected

- [x] WebSDK
- [ ] Backend
- [ ] Dashboard

# Validation

### Info

`src/core/operationRepo/OperationRepo.test.ts`, new `enqueueAndWait` block:

- resolves on success.
- rejects with reason `dropped` on `FailNoretry`.
- rejects with reason `conflict` on `FailConflict`.
- rejects with reason `unauthorized` on `FailUnauthorized`.
- rejects with reason `paused` on `FailPauseOpRepo`.
- rejects with reason `dropped` when the executor throws.

Each case asserts the rejection is an `OperationFailedError`.

### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:

- [x] No Typescript warnings

Other:

### Info

Not needed. No UI change.

### Checklist

---

## Related Tickets

- [SDK-5233](https://linear.app/onesignal/issue/SDK-5233/make-enqueueandwait-reject-with-an-error-value)


src/shared/errors/common.ts
```diff
@@ -34,6 +36,15 @@ export class ChannelCaptureError extends Error {
   }
 }
 
+export class OperationFailedError extends Error {
```
fadi-george, line 39: doesnt seem useful to have this extra class

src/shared/errors/common.ts
```diff
@@ -34,6 +36,15 @@ export class ChannelCaptureError extends Error {
...
+  reason: OperationFailureReasonValue;
```
fadi-george, line 40: can this be _reason so it can be mangled ?

src/core/types/operation.ts
```diff
@@ -52,6 +52,23 @@ export const ExecutionResult = {
 
 export type ExecutionResultValue = (typeof ExecutionResult)[keyof typeof ExecutionResult];
 
+/** Why an operation passed to _enqueueAndWait did not complete. */
+export const OperationFailureReason = {
+  /** The server rejected the request as unauthorized. */
+  _Unauthorized: 'unauthorized',
```
fadi-george, line 58: can these just be 0, 1, 2 etc.?

src/core/operationRepo/OperationRepo.ts
```diff
@@ -27,13 +31,27 @@ const removeOpFromDB = (op: Operation) => {
   void db.delete('operations', op._modelId);
 };
 
+const failureReason = (result: ExecutionResultValue): OperationFailureReasonValue => {
```
fadi-george, line 34: do you do anything with these? do you really need these enums?

src/core/types/operation.ts
sherwinski: Only downside is that the debug logs will be less useful: `Operation failed: ${reason}` would read `Operation failed: 1`

src/core/operationRepo/OperationRepo.ts
sherwinski: We can cut down on the extra ones.

src/core/types/operation.ts
sherwinski: But I guess that's fine since it matches the similar pattern in `ExecutionResult`

src/shared/errors/common.ts
sherwinski: Yep

sherwinski: Renamed in 98133efe

fadi-george (approved):

PR #1496, opened 2026-09-17
feat: [SDK-5231] handle an unauthorized result under Identity Verification

sherwinski:
# Description

## 1 Line Summary

On `FailUnauthorized` under IV, invalidate the user's JWT, wake waiters with `unauthorized`, and re-queue the operations at the head.

## Details

Fifth of 7 M2 pull requests for [Identity Verification for Web SDK](https://linear.app/onesignal/project/identity-verification-for-web-sdk-800ed255b785). Stacked on #1495 (SDK-5233). Targets `sherwin/sdk-5233` now; retarget to `feat/identity-verification` after the earlier PRs merge.

Mirrors the `FAIL_UNAUTHORIZED` branch in Android `OperationRepo.kt`.

**`src/core/operationRepo/OperationRepo.ts`**

`FailUnauthorized` now has its own branch in the execution result switch. When `isIvCodePathEnabled()` is on, `_handleFailUnauthorized(ops, ivBehaviorActive)` runs first. It returns `true` and takes over only when IV behavior is active and the starting operation has an `externalId`. Then it:

1. Calls `JwtTokenStore._invalidateJwt(externalId)` if the stored token is still the one the request went out with. `_executeOperations` snapshots the token before it calls the executor. If `login` or `updateUserJwt` stored a newer token while the request was in flight, the newer token is kept, no event fires, and the re-queued operations retry with it on the next pass. Otherwise the store removes the token and fires its invalidated listener, so the app can supply a fresh token. The public `userJwtInvalidated` event that subscribes to this listener is SDK-5241 (M4). Review follow-up: without the snapshot a late 401 could wipe a good token.
2. Wakes every `_enqueueAndWait` caller in the batch with `false` and the reason `unauthorized`.
3. Re-queues the operations at the head of the queue with no resolver. Bucket and retry count are kept. The operations stay in the store.

The dispatch gate from SDK-5232 then holds the re-queued operations until a new token is stored for that user, so there is no retry loop. A new `login` or `updateUserJwt` releases them on the next tick.

When IV behavior is inactive, or the starting operation is anonymous, the handler returns `false` and the result drops the operations the same as `FailNoretry`. The drop path is now a shared `_dropAndWake` helper that `FailNoretry` and `FailConflict` also use.

**Bundle size.** Adds 196 B gzipped to `page.es6.js`. The page limit moves to the measured 44.07 kB.

# Systems Affected

- [x] WebSDK
- [ ] Backend
- [ ] Dashboard

# Validation

### Info

`src/core/operationRepo/OperationRepo.test.ts`, new `FailUnauthorized` block under the dispatch gate:

- IV active, identified operation: the token is removed, the invalidated listener fires with the `externalId`, the waiter rejects with `unauthorized`, and the operation is back at the head of the queue and still in the store.
- IV active, identified operation: a token stored while the request was in flight is kept, no event fires, and the re-queued operation is eligible right away.
- IV active, identified operation: the re-queued operation does not spin; several ticks pass with one execution, and a new token releases it.
- IV active, anonymous operation: the operation is dropped, no token is touched, and no event fires.
- IV inactive: `FailUnauthorized` drops the operation the same as `FailNoretry`.

### Checklist

**Programming Checklist**
Interfaces:

Functions:

Typescript:

- [x] No Typescript warnings

Other:

### Info

Not needed. No UI change.

### Checklist

---

## Related Tickets

- [SDK-5231](https://linear.app/onesignal/issue/SDK-5231/unauthorized-result-under-iv-invalidate-token-fire-event-wake-waiters)


src/core/operationRepo/OperationRepo.ts
```diff
@@ -294,6 +305,54 @@ export class OperationRepo implements IOperationRepo, IStartableService {
...
+  ): boolean {
+    if (!ivBehaviorActive) return false;
+    const externalId = ops[0].operation._externalId;
+    if (!externalId) return false;
+
+    if (this._jwtTokenStore._getJwt(externalId) === jwtAtDispatch) {
+      this._jwtTokenStore._invalidateJwt(externalId);
+      Log._info(
+        `OpRepo: 401 Unauthorized, JWT invalidated for externalId=${externalId}. Ops re-queued.`,
+      );
```
fadi-george, line 342: i would determine if we really need extra logs as they would add to bundle size
if you really need them, then keep message as short as possible e.g. JWT invalidated

sherwinski: I'll shorten them

src/core/operationRepo/OperationRepo.ts
```diff
@@ -206,13 +210,20 @@ export class OperationRepo implements IOperationRepo, IStartableService {
           break;
 
         case ExecutionResult._FailUnauthorized:
+          // Outer gate: the IV handler runs only on the new code path.
+          if (
+            isIvCodePathEnabled() &&
```
fadi-george, line 215: i think it was renamed to something else, I guess fix that when/after merging

fadi-george (approved):

